PR #665, opened 2026-03-01
feat: add mcp-servers chart with translate sidecar and self-registration

jomcgi:
## Summary

- Adds `charts/mcp-servers/` — a single range-based Helm chart that deploys any number of MCP servers from `.Values.servers`
- Each server gets a Deployment, Service, ServiceAccount, and optional RBAC, 1Password secret, SigNoz HTTPCheck alert, and Context Forge gateway registration Job
- Supports two deployment modes per server: native HTTP (single container) or stdio with translate sidecar (upstream IBM `mcp-context-forge` image exposing streamable-http)
- Migrates `signoz-mcp` as the first server entry, removing the dedicated `charts/signoz-mcp/` chart and `overlays/prod/signoz-mcp/` overlay
- Replaces the monolithic registration job in `context-forge` with per-server self-registration Jobs

## Design

See `docs/plans/2026-03-01-mcp-servers-chart-design.md` for the full design document.

Adding a new MCP server is now a single values entry:

```yaml
servers:
  - name: my-server
    image: { repository: ghcr.io/example/server, tag: v1.0.0 }
    port: 8000
    translate: { enabled: false }
    registration: { enabled: true }
    alert: { enabled: true, url: "http://my-server.mcp-servers:8000/health" }
```

## Test plan

- [x] `bazel test //charts/mcp-servers:lint_test` — chart lints cleanly
- [x] `bazel test //overlays/prod/mcp-servers:template_test` — renders valid YAML with signoz-mcp values
- [x] `bazel test //overlays/prod/...:template_test` — all 12 prod overlays pass (including context-forge after registration job removal)
- [x] `bazel test //charts/...:lint_test` — all 14 chart lint tests pass
- [ ] Verify ArgoCD syncs mcp-servers app after merge
- [ ] Verify signoz-mcp pod is running in mcp-servers namespace
- [ ] Verify registration Job completes and signoz-mcp appears in Context Forge gateway
- [ ] Verify HTTPCheck alert ConfigMap is picked up by signoz-dashboard-sidecar

### DIFF
--- a/charts/mcp-servers/templates/alert.yaml
+++ b/charts/mcp-servers/templates/alert.yaml
@@ -61,6 +61,10 @@ data:
           "queryType": "builder"
         },
         "selectedQueryName": "A",
+        "op": "2",
+        "target": 1,
+        "matchType": {{ $.Values.alertDefaults.matchType | quote }},
+        "targetUnit": "",
         "thresholds": {
           "kind": "basic",
           "spec": [


charts/mcp-servers/templates/registration-job.yaml
```diff
@@ -60,30 +69,31 @@ spec:
...
-                  "url": "{{ .Values.registration.signoz.url }}",
-                  "transport": "{{ .Values.registration.signoz.transport }}"
+                  "name": "{{ .name }}",
+                  "url": "http://{{ .name }}.{{ $.Release.Namespace }}.svc.cluster.local:{{ include "mcp-servers.port" . }}/mcp",
+                  "transport": "{{ .registration.transport | default "STREAMABLEHTTP" }}"
                 }'
 
               echo ""
               echo "Registration complete."
           envFrom:
```
jomcgi: **Cross-namespace secret dependency: registration Job will fail at runtime.**

The registration Job does `envFrom.secretRef.name: context-forge-jwt`, but this secret doesn't exist in the `mcp-servers` namespace. The context-forge chart creates a `OnePasswordItem` named `context-forge` (not `context-forge-jwt`) in the `mcp-gateway` namespace. Two problems:

1. **Name mismatch** — the 1Password secret is `context-forge`, not `context-forge-jwt`
2. **Namespace isolation** — even if the name matched, Kubernetes secrets are namespace-scoped. The `mcp-gateway` namespace secret isn't visible to Jobs in `mcp-servers`

The chart either needs to:
- Add a `OnePasswordItem` for the gateway JWT secret in the `mcp-servers` namespace (with a dedicated 1Password item containing just `JWT_SECRET_KEY`)
- Or use an `ExternalSecret`/cross-namespace mechanism

Without this fix, every registration Job will fail with `secret "context-forge-jwt" not found`.

charts/mcp-servers/templates/deployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        {{- if and .translate .translate.enabled }}
+        - name: server
+          image: "{{ .image.repository }}:{{ .image.tag }}"
+          imagePullPolicy: {{ .image.pullPolicy | default "IfNotPresent" }}
+          {{- with .env }}
+          env:
+            {{- toYaml . | nindent 12 }}
```
jomcgi: **Translate mode: `server` container appears non-functional in the current architecture.**

When `translate.enabled` is true, this template creates two containers:
1. **`server`** — runs the upstream MCP server image with `stdin: true`, no ports, no probes
2. **`translate`** — runs `mcpgateway.translate --stdio <command> --expose-streamable-http`

The `--stdio` flag takes a command argument that the translate process spawns as an *internal subprocess*. It doesn't connect to another container's stdin. In Kubernetes, containers in a pod do NOT share stdin/stdout — they're isolated processes.

This means:
- The `server` container will start, find nothing on stdin, and either idle or exit
- The `translate` container runs `<command>` internally, but that command must exist *inside the translate image* (the IBM mcp-context-forge image), not in the server image

**Suggested fix:** In translate mode, only run the `translate` container. The `server` container should be omitted. If the MCP server binary needs to be available inside the translate container, consider either:
- An init container that copies the binary to a shared `emptyDir` volume
- Building a custom image that includes both
- If `--stdio <command>` expects the command from the translate image itself (e.g. `python -m some_module`), the `server` image field becomes unused in translate mode

This isn't blocking since `signoz-mcp` uses `translate.enabled: false`, but it will break when the first stdio server is onboarded.

PR #1203, opened 2026-03-16
fix(cluster-agents): remove hardcoded .svc.cluster.local URL defaults

jomcgi:
## Summary

- `main.go` had hardcoded `.svc.cluster.local` fallback defaults for `SIGNOZ_URL` and `ORCHESTRATOR_URL`, violating the `no-hardcoded-k8s-service-url` semgrep rule
- This caused `bazel test //...` to fail in CI, preventing the container image from being built and pushed
- Without a fresh image, the cluster-agents pod was stuck in `ImagePullBackOff` or running a stale image — causing the "cluster-agents Unreachable" alert to fire
- Both URLs are already correctly set in `deploy/values.yaml` and injected as env vars via Helm, so no hardcoded defaults are needed

## Test plan

- [ ] CI semgrep test `no-hardcoded-k8s-service-url` passes for `main.go`
- [ ] `bazel test //...` passes and image is pushed on merge to main
- [ ] cluster-agents pod comes up healthy and `/health` endpoint returns 200
- [ ] "cluster-agents Unreachable" alert clears

### DIFF
--- a/projects/agent_platform/cluster_agents/deploy/Chart.yaml
+++ b/projects/agent_platform/cluster_agents/deploy/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: cluster-agents
 description: Autonomous cluster monitoring agents
 type: application
-version: 0.3.8
+version: 0.3.9
 appVersion: "0.2.0"
 annotations:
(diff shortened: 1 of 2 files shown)


jomcgi:
Closing — all changes were already landed on main via other PRs.

PR #1271, opened 2026-03-17
test(trips): add unit tests for wildlife, delete-points, and backfill-elevation tools

jomcgi:
## Summary

- **`wildlife_test.py`** (45 tests): `CaptureQueue` (init, add, state transitions, `get_pending`, `reset_downloading`, `get_stats`), `GracefulShutdown` (signal handling, double-signal force-quit), `PerfStats` (capture/download tracking, summary output incl. throughput calculation)
- **`delete_trip_points_test.py`** (22 tests): `publish_delete` tombstone format (subject, payload, JSON encoding), date/source filtering logic, gap grouping by date, batch publish behaviour, dry-run guard
- **`backfill_elevation_test.py`** (35 tests): `TripPoint.from_dict`/`to_dict` (full fields, defaults, elevation omission when `None`, zero-elevation preservation), `publish_point`, `replay_stream` (deduplication, tombstones, malformed messages, `StreamNotFoundError`), and `run_backfill` force/dry-run filtering

## Test plan

- [ ] CI runs `bazel test //projects/trips/tools/detect-wildlife:wildlife_test`
- [ ] CI runs `bazel test //projects/trips/tools/delete-trip-points:delete_trip_points_test`
- [ ] CI runs `bazel test //projects/trips/tools/backfill-elevation:backfill_elevation_test`
- [ ] All 102 new tests pass

### DIFF
--- a/projects/trips/tools/backfill-elevation/backfill_elevation_test.py
+++ b/projects/trips/tools/backfill-elevation/backfill_elevation_test.py
@@ -441,9 +441,7 @@ class TestReplayStream:
         import nats.js.errors
 
         mock_js = AsyncMock()
-        mock_js.pull_subscribe = AsyncMock(
-            side_effect=nats.js.errors.NotFoundError()
-        )
+        mock_js.pull_subscribe = AsyncMock(side_effect=nats.js.errors.NotFoundError())
         result = await replay_stream(mock_js)
         assert result == []
 


jomcgi (commented):
## Review Summary

102 tests across three files. The **wildlife tests are excellent** — they use a real SQLite database via `tmp_path`, cover all state machine transitions, retry-count exhaustion, and signal-handling edge cases with the right level of granularity. However, the delete-trip-points and backfill-elevation files have structural problems that provide **false coverage confidence**.

---

### ✅ `wildlife_test.py` — Approve

- `CaptureQueue` tests hit real SQLite via `tmp_path`; no mocking of the persistence layer (correct).
- State machine tests cover all transitions: `PENDING → DOWNLOADING → COMPLETED/FAILED`, retry exhaustion at `MAX_RETRIES`, and the restart-recovery path (`reset_downloading`).
- `get_pending` boundary: failed records at exactly `MAX_RETRIES` are excluded — this is the critical off-by-one edge case and it's covered.
- `GracefulShutdown`: double-signal `SystemExit` path tested by calling `_handler` directly; context manager signal restoration verified.
- `PerfStats.summary()` string-matching is reasonable since the output format is what users see.
- Naming follows existing project conventions (`TestFooBar` class + `test_verb_noun` methods).

---

### ❌ `delete_trip_points_test.py` — Blocking Issues

**1. `TestDateFiltering` and `TestGapListingLogic` test inline re-implementations, not source code.**

The filtering logic in `main.py` lives inside the `_delete` inner function of the `by_date` command and the `_list` inner function of `list_gaps`. Because inner functions cannot be imported, the tests copy-paste the identical filtering expression and assert against it:

```python
# From the test — this is NOT calling main.py
result = [
    p
    for p in points
    if p["timestamp"].startswith(date) and p["source"] == source
]
assert len(result) == 2
```

This will pass even if `main.py` contains a bug in that same filter expression. These tests verify the algorithm in a vacuum, giving false coverage confidence. The fix is either to refactor `main.py` to expose testable helper functions (e.g. `filter_points_by_date_source(points, date, source)`), or to use `typer`'s CLI test runner (`typer.testing.CliRunner`) with a mocked `httpx` client to exercise the actual command.

**2. `test_dry_run_does_not_publish` is trivially vacuous.**

```python
async def test_dry_run_does_not_publish(self):
    mock_js = AsyncMock()
    dry_run = True
    point_ids = ["id1", "id2"]
    if not dry_run:          # Always False — publish_delete is never called
        for pid in point_ids:
            await publish_delete(mock_js, pid)
    mock_js.publish.assert_not_called()
```

The test sets `dry_run = True` itself and wraps the call in `if not dry_run:`, so `publish_delete` is never reached. The assertion is trivially satisfied without exercising any production code. This test should instead call the `by_date` CLI command with `--dry-run` via `CliRunner` and verify no NATS publishes occur.

---

### ⚠️ `backfill_elevation_test.py` — Issues

**3. `test_consumer_unsubscribed_after_replay` uses incorrect mock access.**

```python
async def test_consumer_unsubscribed_after_replay(self):
    mock_js = self._make_js([[]])
    await replay_stream(mock_js)
    consumer = await mock_js.pull_subscribe.return_value   # ← bug
    consumer.unsubscribe.assert_called_once()
```

`mock_js.pull_subscribe.return_value` is the `consumer` AsyncMock. Awaiting it (`await consumer_async_mock`) invokes its coroutine protocol and returns `consumer_async_mock.return_value` — a *different* child mock whose `unsubscribe` attribute has never been called. The correct access pattern is:

```python
consumer = mock_js.pull_subscribe.return_value   # direct ref, no await
consumer.unsubscribe.assert_called_once()
```

If this test is currently passing it is because the assertion is accidentally succeeding against a fresh mock's call count, not because `unsubscribe` is being verified.

**4. `TestRunBackfillFiltering` re-implements `run_backfill` filtering inline.**

Same pattern as the delete-trip-points issue: the force/non-force filter logic is copy-pasted from `run_backfill` and evaluated directly, not via the imported function. These tests don't exercise `run_backfill` at all.

**5. `run_backfill` end-to-end path is untested.**

The bulk of `run_backfill` (NATS connection, `ElevationClient`, publishing loop) has no test coverage. A test mocking `nats.connect` and `ElevationClient` would verify that elevations are correctly attached to points before re-publishing, and that dry-run exits before publishing.

---

### Suggestions (non-blocking)

- **Refactor inner functions** in `delete_trip_points/main.py` and `backfill_elevation/main.py` to be module-level helpers so they can be tested directly.
- **Use `typer.testing.CliRunner`** for CLI-level tests where inner-function extraction is impractical.
- `test_elevation_zero_preserved` is correctly present — zero is a valid elevation and the `if self.elevation is not None` guard handles it properly. Good catch.

PR #1342, opened 2026-03-19
test(orchestrator): add coverage for summarization integration

jomcgi:
## Summary

- Adds 9 new tests in `consumer_test.go` covering the summarization integration added in commit range `20d24db6..219bf1d4`
- `summarizeAndStore` error paths: `store.Get` failure after LLM success, `store.Put` failure, and empty title+summary early-exit guard
- `summarizeAndStoreOnJob`: direct test verifying title+summary are written to the in-memory job record; LLM error leaves existing fields unchanged
- Terminal state summarization: `processJob` with `JobSucceeded` and `JobFailed` both trigger `summarizeAndStoreOnJob` when plan steps are present, and the results are persisted
- Nil-summarizer safety: `processJob` with plan steps and a nil summarizer completes without panic

## Test plan

- [ ] `bazel test //projects/agent_platform/orchestrator:orchestrator_unit_test` passes
- [ ] All 9 new tests (`TestSummarizeAndStore_*`, `TestSummarizeAndStoreOnJob_*`, `TestProcessJob_*Plan*`, `TestProcessJob_NilSummarizer_WithPlanSteps_NoPanic`) are present and green

### DIFF
--- a/projects/agent_platform/chart/Chart.yaml
+++ b/projects/agent_platform/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: agent-platform
 description: Agent platform — umbrella chart bundling goose sandboxes, MCP servers, and supporting components
 type: application
-version: 0.27.9
+version: 0.27.10
 appVersion: "0.2.0"
 annotations:
   org.opencontainers.image.source: "https://github.com/jomcgi/homelab"
--- a/projects/agent_platform/deploy/application.yaml
+++ b/projects/agent_platform/deploy/application.yaml
@@ -9,7 +9,7 @@ spec:
     # Chart from OCI registry (pushed by CI via Bazel helm_push)
     - repoURL: ghcr.io/jomcgi/homelab/charts
       chart: agent-platform
-      targetRevision: 0.27.9
+      targetRevision: 0.27.10
       helm:
         releaseName: agent-platform
         valueFiles:


jomcgi (commented):
## Code Review: test(orchestrator): add coverage for summarization integration

### Overall Verdict: APPROVE (posted as COMMENT — GitHub blocks self-approval)

The 9 new tests are well-structured, well-targeted, and correct against the production code. They exercise the code paths described in the PR body with no flaky patterns, no goroutine leaks, and no reliance on real timing. A few minor observations are noted below, none of which warrant blocking the merge.

---

### 1. Code path coverage

**summarizeAndStore**

| Scenario | Test | Verdict |
|---|---|---|
| `store.Get` failure after LLM success | `TestSummarizeAndStore_StoreGetFailure` | Correct — `failGetStore` always returns an error from `Get`; production code logs and returns |
| `store.Put` failure after Get succeeds | `TestSummarizeAndStore_StorePutFailure` | Correct — `errOnPutStore{failAfter:0}` fails the first Put into the wrapper |
| Empty title + summary early-exit guard | `TestSummarizeAndStore_EmptyTitleAndSummary` | Correct — verifies the existing field is not clobbered |

One **gap** worth noting: `TestSummarizeAndStore_StoreGetFailure` uses `failGetStore{inner: newMemStore()}` and never seeds the job into the inner store. That is intentional and fine (the test is only checking that Get failure is handled gracefully), but the `inner` store is effectively unused. A comment noting this intent would help future readers, but it is not a correctness issue.

**summarizeAndStoreOnJob**

| Scenario | Test | Verdict |
|---|---|---|
| LLM success — fields mutated on the job struct | `TestSummarizeAndStoreOnJob_SetsFields` | Correct — tests both `Title` and `Summary` fields on the job pointer |
| LLM error — fields left unchanged | `TestSummarizeAndStoreOnJob_LLMError` | Correct — `Title` is pre-set, verified it is not zeroed out on error |

**Terminal state summarization via processJob**

| Scenario | Test | Verdict |
|---|---|---|
| `JobSucceeded` with plan → summary written to store | `TestProcessJob_SuccessWithPlan_SetsSummary` | Correct |
| `JobFailed` (retries=0) with plan → summary written to store | `TestProcessJob_FailedWithPlan_SetsSummary` | Correct |
| Nil summarizer with plan steps → no panic | `TestProcessJob_NilSummarizer_WithPlanSteps_NoPanic` | Correct — `SummarizePlan` has a `s == nil` guard |

**Missing coverage (low severity, not blocking)**

- Trigger 2 / Trigger 3 in the ticker loop (`summarizeAndStore` called during a mid-execution flush) is not directly exercised. Those paths require a real tick, which is why they are omitted — this is a deliberate and acceptable trade-off since exercising them would require either replacing the ticker or inserting sleep-based timing.
- `TestSummarizeAndStore_EmptyTitleAndSummary` only asserts that `Title` is preserved and `Summary == ""`. The assertion on `Summary` is vacuous since the initial job has `Summary: ""` (zero value). A future test with a pre-set summary would strengthen the guard against clobbering a non-empty existing summary.

---

### 2. Error path coverage

All three error sub-paths in `summarizeAndStore` are covered:
- LLM failure: covered by `TestSummarizeAndStoreOnJob_LLMError` (exercised via `summarizeAndStoreOnJob`, which delegates to the same `SummarizePlan` call)
- `store.Get` failure: `TestSummarizeAndStore_StoreGetFailure`
- `store.Put` failure: `TestSummarizeAndStore_StorePutFailure`

Both return early (log + return) without panicking. Tests confirm this correctly.

---

### 3. Flakiness and concurrency patterns

No sleeps, no `time.After`, no goroutine leaks. All `httptest.Server` instances are correctly deferred with `Close()`. The fake sandbox `runFn` is synchronous. `processJob` itself spawns an internal goroutine for the sandbox and a ticker loop, but since the test's sandbox returns immediately, the result channel is drained on the first select iteration and the ticker never fires — no race condition possible.

The `errOnPutStore` uses `callCount *int` (a shared pointer). In the new tests this is only accessed sequentially, so no data race.

---

### 4. Convention adherence

The new tests follow existing conventions well:
- Same `package main` inline test style
- `pendingJob` / `newFakeMsg` / `newMemStore` / `errOnPutStore` helpers are all reused correctly
- `chatResponse()` from `summarizer_test.go` is used (same package, shared across test files)
- New `failGetStore` type is placed with a clear comment block above it, matching the convention used for `errOnPutStore`
- `NewConsumer(nil, store, sandbox, nil, summarizer, 5*time.Minute, slog.Default())` is used directly — consistent with how `TestProcessJob_SummarizesOnSubmit` constructs the consumer, rather than going through `newTestConsumer` which hardcodes `nil` for the summarizer

Minor style note: `TestSummarizeAndStore_StorePutFailure` sets `failAfter: 0` meaning "fail on the very first Put call" (`callCount > failAfter` is `1 > 0`). The semantics are correct but slightly unintuitive. The existing test `TestProcessJob_StoreUpdateFailureOnStart_Nacks` already has the clarifying comment `// fail on the very first Put (transition to RUNNING)` — the new test would benefit from a similar inline comment.

---

### 5. BUILD file

The BUILD file is **not modified** in this PR. `consumer_test.go` is already listed as a source in both `orchestrator_unit_test` and `orchestrator_test` targets. The new tests live in the same file, so no BUILD change is needed. This is correct.

The new `failGetStore` type and test functions introduce no new external imports. All symbols used (`errors`, `context`, `net/http`, `net/http/httptest`, `slog`, `time`) are already declared at the top of `consumer_test.go`.

---

### Summary

The PR is a clean, focused test-only addition. It correctly exercises the three summarization integration points, covers all error return paths in `summarizeAndStore`, and introduces no new risk. The implementation it tests (`summarizeAndStore`, `summarizeAndStoreOnJob`, and the two terminal-state call sites in `processJob`) is exercised end-to-end through `httptest` servers — a pragmatic and reliable approach for this type of integration.

**Approved with no required changes.**

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

PR #1388, opened 2026-03-20
feat(hooks): add check-recipe-context-forge PreToolUse hook

jomcgi:
## Summary

- Adds `check-recipe-context-forge.sh`, a PreToolUse warning hook that fires when writing/editing a recipe YAML with a `type: builtin` extension but no corresponding `type: streamable_http` extension
- Adds `check-recipe-context-forge_test.sh` with 9 test cases covering the key paths (empty input, non-recipe paths, both extension variants, Edit vs Write tool fields)
- Registers the hook in `.claude/settings.json` at the end of the `Write|Edit` matcher hooks list
- Adds the `sh_test` target to `bazel/tools/hooks/BUILD`

This prevents the class of bug from PR #1377 where 3 recipes were missing context-forge integration by warning at write time rather than at review time.

## Test plan

- [ ] CI runs `bazel test //bazel/tools/hooks:check_recipe_context_forge_test` and all 9 cases pass
- [ ] Writing a recipe YAML with only `type: builtin` triggers the WARNING on stderr
- [ ] Writing a recipe YAML with both `type: builtin` and `type: streamable_http` produces no warning
- [ ] Writing non-recipe YAML files produces no output


jomcgi (commented):
## Code Review

Both the hook implementation and test file look solid. A few things worth noting:

**[MINOR] Edit-tool false positive for partial edits**

The hook reads `.tool_input.new_string` for Edit operations, which contains *only the replacement text*, not the full file. If a developer edits the `description:` field of a recipe that already has both `type: builtin` and `type: streamable_http` elsewhere, the `new_string` snippet won't contain `type: streamable_http` — triggering a spurious WARNING. Not blocking (exit 0), but degrades signal quality over repeated partial edits.

A future improvement: when `file_path` is available and the file exists on disk, read and grep the on-disk content instead of (or in addition to) `new_string`. This would eliminate false positives entirely.

**[MINOR] Missing test case for Edit-tool with both extensions**

Test 6 (`recipe_builtin_and_streamable_http`) only covers the Write path (`content` field). Adding a parallel case with `new_string` containing both extensions would catch the above false-positive scenario in CI.

No blocking issues — looks good to merge.

PR #2881, opened 2026-06-28
feat(agent_platform): transparent egress proxy for Firecracker guests (ADR 023 6a)

jomcgi:
## What

Reworks fc-agentd guest egress (ADR 023 stage 6a) from the planned `HTTP_PROXY` forward-proxy into a **transparent data plane**. The guest is configured with real URLs and needs no per-destination routing config; goose (whose reqwest [ignores `HTTP_PROXY`](https://github.com/block/goose/issues/6835)) works unchanged.

## How

- **fc-agent-init** becomes a dumb funnel: a wildcard DNS responder answers every name with `127.0.0.1`, and per-port loopback listeners (`80/443/8080`) tunnel each connection over vsock to the host sidecar. Drops `HTTP_PROXY` and the old forward-proxy/redirect listeners.
- **egress-proxy** is rewritten as a transparent proxy: it reads the funnel's port preamble, recovers the destination host by **peeking the TLS SNI or HTTP Host header** (without consuming the stream), applies policy, and pipes to the real upstream (the pod resolves cluster + public DNS).
- **`EGRESS_POLICY`**: `allow` (default, open egress in/out of cluster) or `allowlist` (fail closed). The per-secret `egressTo` (6b) remains the exfiltration control, not a global firewall.
- **chart**: replaces the egress channel/allowlist/default-upstream values with `policy` + `harnessEnv` (real URLs), injected as `FC_AGENTD_INJECT_*`.
- Unit tests for the hand-rolled SNI, Host, and DNS parsers.

## Design notes

- The whole transparent model exists because goose ignores `HTTP_PROXY` (confirmed: the fix shipped in goose v1.24.0 but is incomplete; we run v1.27.1 and it still connects directly) and the guest is vsock-only. Routing by SNI/Host in the sidecar means **no per-destination config** and real URLs everywhere.
- Connection capture is **per-port loopback listeners** today (covers the ports the harness uses). Generic any-port capture via `iptables` REDIRECT + `SO_ORIGINAL_DST` is a localized follow-up that needs the guest kernel built with netfilter; ~80% of this code is shared with that upgrade.
- ADR 023 architecture section updated to describe the transparent data plane and the `allow`-by-default posture.

## Validation

- `go test` green locally for the new parsers (SNI ClientHello, HTTP Host, DNS A/NODATA), including peek-non-consumption guarantees.
- `helm template` renders the egress env wiring correctly.
- In-cluster `goose -> Qwen` validation to follow once the chart deploys (chart-version-bot will bump on merge).


jomcgi:
## Public page visual diff
Changed: **4**, New: **0**, commit `765fdd2e`

<details><summary><b>home-desktop</b> (changed)</summary>

| before | after | diff |
|---|---|---|

</details>
<details><summary><b>home-mobile</b> (changed)</summary>

| before | after | diff |
|---|---|---|
|  |  |  |

</details>
<details><summary><b>engineering-desktop</b> (changed)</summary>

| before | after | diff |
|---|---|---|

</details>
<details><summary><b>engineering-mobile</b> (changed)</summary>

| before | after | diff |
|---|---|---|

</details>

PR #2987, opened 2026-07-01
style(campsites): overlay alignment + mobile rework

jomcgi:
Two polish passes on the /app/campsites map.

**Desktop alignment**: the four overlay cards now form a tidy frame. Top row (crumb card + RANKED PARKS) share a top edge and a 52px height; bottom row (legend + detail) share a bottom edge and a 168px height; consistent 16px insets; the detail panel's right inset is dynamic so it clears the zoom controls when the list is collapsed and the list panel when it is open.

**Mobile rework** (<=768px only, desktop unchanged): slim one-line top bar (the verbose "As of" note is hidden on mobile), the legend becomes a pop-out behind a toggle chip instead of a big permanent card, and the bottom RANKED PARKS bar is a fixed 48px bar that nothing overlaps (legend popout, detail sheet, and zoom controls all sit above it).

Deploy: bumps monolith-public to 0.82.6.

### DIFF
--- a/projects/monolith-public/chart/Chart.yaml
+++ b/projects/monolith-public/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: monolith-public
 description: Read-only public tier of the monolith (ADR 004 Phase 5), Linkerd-meshed and pruned
 type: application
-version: 0.82.5
+version: 0.82.6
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

PR #2988, opened 2026-07-01
feat(model-bench): structured-output I/O contract

jomcgi:
Ran hard tasks directly against OpenRouter and found the emit-file-as-text I/O contract conflates output *format* with capability. Opus wrote an excellent null-content fix, but its response leaked a prose line with a unicode arrow into the extracted file (invalid character), and qwen3-coder renamed the function - both format failures, not capability, and temp-0 nondeterministic.

Fix: file-emitting tasks now use a structured contract. The prompt asks for a JSON object `{"files":[{"path","content"}]}` and the runner parses it, falling back to lenient text extraction when a model does not return JSON. A JSON content field is a discrete string, so prose/markdown/unicode cannot leak into the file.

Validated live across 14 models: Opus and the rest pass cleanly now (only qwen3-coder-30b needs one repair round). `response_format` is deliberately NOT forced - it returns null for glm/deepseek and would regress passing models.

Also adds `null-content-fix-01` (a representative code-fix task from the real content:null crash) and bumps HARNESS_VERSION 0.1.3 -> 0.1.4.

Generated with Claude Code

### DIFF
--- a/projects/model-bench/bench/cache.py
+++ b/projects/model-bench/bench/cache.py
@@ -1,7 +1,7 @@
 import hashlib
 from pathlib import Path
 
-HARNESS_VERSION = "0.1.3"
+HARNESS_VERSION = "0.1.4"
 
 
 def cell_key(
--- a/projects/model-bench/bench/runner.py
+++ b/projects/model-bench/bench/runner.py
@@ -13,12 +13,45 @@ Safety invariants:
 from __future__ import annotations
 
 import hashlib
+import json
 import shutil
 import tempfile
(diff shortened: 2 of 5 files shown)


jomcgi:
## Public page visual diff
Changed: **2**, New: **0**, commit `da248aa`

<details><summary><b>engineering-desktop</b> (changed)</summary>

| before | after | diff |
|---|---|---|

</details>
<details><summary><b>engineering-mobile</b> (changed)</summary>

| before | after | diff |
|---|---|---|

</details>

PR #3022, opened 2026-07-01
feat(git-mirror): mirror-side token to mirror private repos (loom)

jomcgi:
Answers 'don't we have creds for loom': yes. Gives the mirror a read token (reused monolith-chat-secrets/GITHUB_TOKEN) so it can clone the private loom repo, and re-adds loom. The token stays in the trusted mirror pod only, written to a credential store in the ephemeral HOME (/tmp emptyDir): never on the PVC (remote URLs stay token-less), never in a guest (guests clone over git:// with no cred, preserving ADR 023). Clone failure stays non-fatal. Chart 0.1.2 -> 0.1.3.

Follow-up: a fine-grained read-only PAT scoped to just these repos would be tighter than the broad token.


jomcgi:
## Public page visual diff
Changed: **2**, New: **0**, commit `fd20493`

<details><summary><b>llm-leaderboard-desktop</b> (changed)</summary>

| before | after | diff |
|---|---|---|

</details>
<details><summary><b>llm-leaderboard-mobile</b> (changed)</summary>

| before | after | diff |
|---|---|---|

</details>

PR #3092, opened 2026-07-02
fix(goosecracker): research search must POST, router must dispatch research

jomcgi:
Two fixes from live smoke tests of the research sub-recipe shipped in #3087 + #3090:

1. **SearXNG JSON needs POST.** On our deployment (searxng 2026.4.4) `GET /search?...&format=json` returns HTML; only POST form data returns JSON. The direct research-recipe run recovered by discovering this itself, but the prescribed incantation should work first try on a 32k-context model. The pipeline in `research.yaml` now uses `curl -s -X POST ... -d "q=...&format=json"`.

2. **Router HARD GATE for research dispatch.** In both router-level smoke tests the qwen router answered the research task inline instead of dispatching the sub-recipe: it has no SEARXNG_URL guidance (that lives in research.yaml), so it flailed through public SearXNG instances (403s), DuckDuckGo, and Wikipedia scraping for ~30 turns, then reported `mode: query`. Added dispatch-mandate language to the research route mirroring the artifact route's HARD GATE, which fixed the identical failure mode for artifacts.

Verification: the direct-invocation run (thread t-c2454650cc97) proves the full path works once the sub-recipe executes: SEARXNG_URL resolved, egress sidecar allowed the in-cluster SearXNG call, primary sources fetched with bounded reads, and the answer correctly separated confirmed facts from unconfirmed ones. Post-merge I'll rerun the router-level smoke test.

Includes the manual fc-invoke chart bump to 0.4.21 (`Chart.yaml` + `targetRevision`); the chart-version bot still does not scope guest content (see #3090 notes).

### DIFF
--- a/projects/firecracker/goosecracker/guest/recipes/agent.yaml
+++ b/projects/firecracker/goosecracker/guest/recipes/agent.yaml
@@ -1,4 +1,4 @@
-version: "1.7.0"
+version: "1.7.1"
 title: "Agent"
 description: "Routing agent for a snapshot-managed microVM thread (ADR 022): classifies the task and dispatches the matching sub-recipe."
 instructions: |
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

PR #3126, opened 2026-07-03
feat(chat): live per-search status in the Discord checklist

jomcgi:
## What

Builds on #3124. The `🔍 Searching...` progress message now shows each search as a checklist row that flips from pending (`⏳`) to done (`✅`), mirroring the ADR-035 agent-stage markers.

Before (post-#3124): flat `• query` bullets, no completion signal.
After: `⏳ query` while the search runs, `✅ query` once it returns.

## How

- New `elif isinstance(event, FunctionToolResultEvent)` branch in `_stream_response`. PydanticAI emits this when a tool returns, carrying the same `tool_call_id` as the originating call event.
- `search_status: dict[query -> done]` keeps the dedup identity (a model-unavailable retry replays the same query, so it stays one row).
- `id_to_query: dict[tool_call_id -> query]` maps every seen call id (including a retry's fresh id) back to its query, so whichever call completes flips the single visible row.
- `_search_content()` renders the checklist; it reuses the raw-emoji marker style already used by `render_checklist`.

The checklist is transient: the final answer text (or the fallback) overwrites the message, so a search still pending when the reply lands is never the terminal visible state (per the agreed "assume done" behavior).

## Tests

- `test_tool_result_flips_search_to_done` — a call event then a matching result event renders `⏳ trail cafes` and later `✅ trail cafes`.
- Updated `test_retried_tool_call_shows_single_bullet` to be marker-agnostic (asserts the query row never repeats).
- Existing "Searching" / query-decoding tests still hold (the header and query text are unchanged).

https://claude.ai/code/session_01PpvtXSA9vTVd9Bm4qrApsu

### DIFF
--- a/projects/monolith/chat/bot.py
+++ b/projects/monolith/chat/bot.py
@@ -14,6 +14,7 @@ from pydantic_ai import (
     AgentRunResultEvent,
     BinaryContent,
     FunctionToolCallEvent,
+    FunctionToolResultEvent,
     PartDeltaEvent,
     TextPartDelta,
     ThinkingPartDelta,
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

PR #3144, opened 2026-07-03
fix(grimoire): batch embeddings by content size, self-heal missing vectors

jomcgi:
## What

Fixes the grimoire chunk-embedding failure found while loading the Monster Manual: the load embedded only 64/1,224 chunks, then failed.

## Root cause

The loader embedded in **fixed batches of 64**. The embedding server (`voyage-4-nano` via llama.cpp) is **token-throughput-bound**: reproduced directly, 64 small front-matter chunks embed in ~4s, but 64 stat-block chunks (~32,365 tokens) take **77.9s**, over the client's **60s `EMBED_BATCH_READ_TIMEOUT`**. The batch times out, retries forever, and the load fails. Worse, the loader only re-embeds *changed* content, so the stuck chunks would never retry on a re-run.

It's not a size-limit or rejection, the endpoint returns all 64 vectors; the client just gives up first.

## Fix

- **Size-budget batching** (`EMBED_CHAR_BUDGET`, ~60k chars ≈ 36s) + an item cap, so every request finishes well under the timeout regardless of chunk size. Total embed time is throughput-bound (~13 min for the book) either way, so this is about reliability, not speed.
- **Self-heal**: also (re-)embed chunks that exist but have no vector for the model, so a plain re-run finishes an interrupted embed (recovers the 1,160 stuck chunks). Each batch commits, so partial progress is durable and convergent across runs.
- **Env-overridable timeouts** (`EMBED_BATCH_READ_TIMEOUT`, `EMBED_RETRY_TIMEOUT`) in the shared client (defaults unchanged); grimoire jobs set 300s/900s as margin.
- chart 0.268.2 → 0.269.0.

## Verification

Reproduced the 78s/60s timeout against the live endpoint with the real batch-2 content. Unit tests for the batcher (size + count bounds, oversized-row-alone) and self-heal (delete vector → re-run re-embeds). ruff clean; `helm template` renders both grimoire CronWorkflows with the new env.

After deploy, a re-run of `grimoire-load-chunks` will embed all 1,224 chunks.

### DIFF
--- a/projects/monolith/chart/Chart.yaml
+++ b/projects/monolith/chart/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: monolith
 description: Consolidated homelab web services
-version: 0.268.2
+version: 0.269.0
 type: application
 dependencies:
   - name: cf-ingress
--- a/projects/monolith/chart/templates/cronworkflows.yaml
+++ b/projects/monolith/chart/templates/cronworkflows.yaml
@@ -124,6 +124,15 @@ spec:
               value: {{ $.Values.stars.s3SecretAccessKey | quote }}
             - name: EMBEDDING_URL
               value: {{ $.Values.chat.embeddingUrl | quote }}
+            # Bulk sourcebook loads embed thousands of chunks; the embedding
+            # server is token-throughput-bound, so a batch of large stat-block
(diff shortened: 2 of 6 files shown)


jomcgi:
## Helm Manifest Diff

PR #3735, opened 2026-07-20
fix(ember): default the semgrep demo to javascript

jomcgi:
The per-scan cost is the language rule pack (pro-javascript 1.9MB vs
pro-python 15MB), so javascript scans in ~200ms vs ~2s for python. Lead
with the instant one; python stays one toggle away.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


jomcgi:
## Helm Manifest Diff

PR #4027, opened 2026-07-26
docs(adr): embervm/022 domain composition and access fabric

jomcgi:
## Summary

Stacked on #4025 (ADRs 019–021). Adds **ADR embervm/022** capturing the composition model we converged on:

- **Service composition, not groups:** multi-component apps are independent Workloads wired by bindings (DSN/URL inject), not R5 composite private nets.
- **Domain** as the boundary seam (required even with one operator domain) so multi-customer enforcement is a later policy tighten; hard multi-tenancy stays ADR 009 Recorded.
- **Mediated access fabric**, not a flat guest mesh: opt-in dial re-enters Envoy/policy so wake, auth, and metering stay on-path; default zero-egress remains.
- **Independent wake** + floor/cap for min avail vs density.
- **Horizontal scale is class-specific:** serving/task HPA yes; stateful singleton stays; **Postgres RO replica autoscaling recorded and deferred** (special case, small near-term value).
- **Near-term simplify:** singleton stateful, bankable embeddings serving, domain+bindings; composite and RO HPA off the critical path.

Lambda MicroVMs noted as session prior art, not a replacement for R4 volume durability.

## Verification

- Docs-only plus chart bumps for regenerated `docs-manifest.json` / `repo_docs_manifest.ndjson` (same coupling as #4025).
- Index row for 022; manifests regenerated with canonical generators.
- Chart bumps: monolith `0.285.359`, monolith-public `0.89.285`.
- Local pre-push `ci test` hit a BuildBuddy runner disk-full (`No space left on device` in apko fetch), unrelated to this diff; PR Workflows will re-run the real checks. Parent #4025 is already green.

## Base

`claude/kloak-vsock-coprocessor-70fsbz` (#4025). Merge after that PR lands, or rebase onto main once 019–021 are on main.


jomcgi (commented):
## Review: ADR embervm/022

Two independent passes (Opus, plus an adversarial Fable review). CI is green on `b85a691`, all 7 statuses. **Good ADR, hold at Draft.** The composition-not-groups decision is well argued and the three-legged framing is the insight that makes it worth having. Three load-bearing claims are currently aspirational, and one hard capacity limit isn't mentioned.

Findings are ranked. Everything is verified against the tree unless marked as judgement.

---

### High

**1. The ADR 020 compatibility claim skips how a caller acquires credentials.**

Decision 3 claims compatibility because the fabric is "policy and name resolution on dials." But ADR 020 makes session endpoints reachable only via CP-minted short-TTL tokens verified at edge Envoy, specifically so "a client cannot name an arbitrary brick and probe it." A bound VM→VM dial to a session target (this ADR's own example, `access.in: workloadRef: {name: agent-session}`) must therefore either bypass that verification, breaking 020's probe-prevention property, or mint and refresh tokens through admission at `O(caller instances × bindings / TTL)` — load 020's scaling argument never budgeted, and where TTL is a *correctness* bound, not something tunable upward to relieve pressure.

Open Question 3 asks which auth primitive to use, not how callers obtain and refresh it. Those are different questions and the second is the load-bearing one.

Related: ADR 020 is Draft with its own decision 3 marked UNDER REVISION, so "compatible with ADR 020" is currently compatibility with an unsettled document.

**2. "Reuses ADR 023's interception point" is substrate slippage.**

023's shipped data plane is the `fc-agentd` sidecar under `projects/agent_platform`. EmberVM's own daemon has only a stub: `noded/server/server.go` carries "the egress forwarder (present but unused: task VMs get no NIC, egress disabled)", and `noded/config/config.go` carries an unused `EgressSidecarAddr` "so a future egress-enabled workload needs no config reshape."

No Ember class has mediated egress today. The "Path today (partial)" cell for VM→Service attributes agent-platform capability to Ember, and since R8 retires `fc-agentd`, "reuse" actually means porting the terminate/swap/allowlist sidecar into `noded`. That's substantial new work presented as existing plumbing. Worth stating plainly, with 023 as the ported design rather than the running one.

**3. The three legs share vocabulary, not an enforcement mechanism.**

VM→VM and VM→Service enforce at the guest egress hop. Service→VM enforcement is Cilium NetworkPolicy or Envoy authn on the cluster network: a different plane, different identity primitives (ServiceAccount and namespace versus domain), and per the ADR's own risk row, documentation-only at High likelihood.

Unifying the schema is right. The ADR should say the *mechanism* is not unified, and define how a consumer discovers whether `access.in` is actually enforced yet. There's no status signal for that today.

**4. The stateful class is hard-capped at ten workloads, and the ADR doesn't mention it.**

```yaml
statefulTcpPortRange:   { start: 5400, end: 5409 }   # 10 ports
compositeTcpPortRange:  { start: 5410, end: 5419 }   # 10 ports
```

Enforced, not advisory: the CRD watcher validates each stateful workload's `listenPort` falls in range and is unique.

This lands exactly where composition hurts most. The serving leg routes by Host through Envoy and scales fine. The L4 leg has the ceiling, and the L4 leg is **App → DB**, the canonical binding in this ADR's own motivating example. "Compose App + DB by binding" is not a general model at N=10.

The fixes are known (per-workload ClusterIP Services, or a few L4 entry ports with SNI or PROXY-protocol in-band naming). The ADR should at least name the ceiling and the intended direction.

---

### Medium

**5. Domains never confront cells.** ADR 007 assigns workloads to cells and orders nothing across them; ADR 020 makes sampling, capacity, and copy sets cell-scoped and vendor-scoped. This ADR does not contain the word "cell." A domain whose members land in different cells needs cross-cell endpoint resolution the ADR never defines, and same-domain-by-default bindings can silently cross the one boundary the scaling architecture actually enforces. State whether a domain is cell-confined or bindings are cell-transparent, and who resolves cross-cell.

**6. Domain versus principal granularity, which is security-relevant.** The Lineage row ("snapshots and volumes stay in-domain") restates ADR 001's *per-principal* rule at coarser grain; an implementer keying snapshot lineage by domain would violate 001. Separately, `access.in` names workload *definitions* while session instances are per-principal, so `workloadRef: agent-session` grants access to **every principal's** session instance. Same-domain-by-default plus a multi-principal domain means cross-principal dial by default, and ADR 001's isolation rule covers VMs and lineage, not dials, so data can cross a principal boundary through a binding while lineage isolation holds perfectly. Needs instance and principal granular ACL semantics.

**7. Worst-case first-request latency is never stated.** A cold chain (client → API → embeddings → DB, all floor 0) is a *serial* wake cascade ordered by application dial order: a bank restore in the good case, an object-store restore or a stateful cold boot from volume in the bad case. Nothing bounds or acknowledges the sum; the only related risk row is probe thrash at Low impact. Decision 5 presents independent wake purely as an upside, and the compounding cost is what will eventually argue for the wake-ahead-on-bind option it currently rules out. Also unresolved: whose per-principal wake-rate budget does a binding-triggered wake debit?

**8. Binding inject versus snapshot pinning.** Sessions ride their birth env until max-lifetime TTL, and R4 secrets ride a first-boot boot-args seam. Injected URLs and DSNs are therefore frozen into banked lineages, so rotating a binding endpoint or credential cannot reach a banked guest without forcing a cold restart. Either require injected names to be indefinitely stable (which constrains the addressing scheme, and interacts with finding 4) or define rebind-forces-roll semantics.

---

### Low

**9. Decision 6 overstates serving horizontal scale as "Existing product."** Multi-node endpoint fan-out is R7, Decided with an explicit hardware prerequisite, on a fleet with one Firecracker-capable AMD node. Single-node today at best.

**10. `agents/023` is a listed foundation but is still marked Draft**, despite its own 2026-07-01 update recording that it shipped in PR #3010 and deployed at fc-invoke chart 0.4.2. Its status is stale rather than this PR's problem, but building on a Draft is worth knowing.

**11. Numbering nit.** In the Builds-on line, "023 - Egress Secret Proxy" sits among embervm-numbered ADRs (001, 009, 004) and reads as embervm/023, which is literally the next number in this category. Prefix it `agents/023`.

---

### Verified correct

Several claims could plausibly have been wrong and are not: `platform/012-cilium-replaces-linkerd.md` exists (there is also a Cilium ADR at `networking/003`, and both are real); the scratch-k8s "warmth-only, no durable member volumes" claim is verbatim in the template; the monolith scratch-postgres consumer is real at `chart/values.yaml:854`; ADR 009's R6 Facade demotion to Recorded is accurate; ADR 004's session-as-external-sandbox framing is accurate; 023's split-horizon `internal.allowlist` matches its 2026-07-01 update; chart bumps at `0.285.359` / `0.89.285` are correctly one past the base branch; and 022 is the right next number.

The rejections of composite-as-composition and of a flat tap mesh are both well grounded in 001's shipped R5 shape.

---

### Note on the 100k-workload target

If the goal is 100k+ workload *definitions*, two more things move from later to foundational, neither of which is this ADR's fault but both of which it depends on:

- **Workload CRs are the wrong home at that count.** `WorkloadWatcher` is an informer, and its own docs say an error triggers "a fresh LIST." etcd compaction routinely invalidates an old resourceVersion, so a 100k-object relist becomes a recurring hazard rather than a boot cost. Definitions likely need to leave etcd for the CP datastore, with GitOps CRs kept for the subset that wants to live in Git.
- **xDS workload cardinality is distinct from the session cardinality ADR 020 solved.** 100k workloads implies on the order of 10k serving endpoints that genuinely do belong in Envoy config, which needs delta xDS plus scoped resources so a node holds only routes it can serve.

Finding 4's port ceiling is the same class of problem and the most immediate of the three.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #4140, opened 2026-07-29
embervm: adopt the k8s resource model for placement (honest filter + score, then a reservation ledger)


jomcgi:
## Problem

VM placement and brick scaling have accumulated complexity that traces to a single
root: **EmberVM schedules against observed cgroup usage, where Kubernetes schedules
against declared, reserved requests.** Most of the machinery we have added to
compensate (reject floors, reject/retry rails, per-tick headroom debits, denial
counters with seven tuning knobs) is machinery k8s does not need, because it never
schedules on a measurement.

The component mapping we reasoned from is correct. The fourth row is the gap:

| Kubernetes | EmberVM | Status |
| --- | --- | --- |
| pod | VM | present |
| node | brick | present |
| kube-scheduler | CP placement | present, forked six ways |
| Karpenter | `Embervm.BrickController` | present, reads an event rate not a level |
| **requests/limits + allocatable** | **nothing** | **missing, and it is the root** |

k8s is battle-tested here not because filter/score is clever, but because
`NodeInfo.Requested` is incremented at *assume* time (before the pod runs) and
kubelet admission checks the same arithmetic. Observed usage appears only in
eviction, never in placement.

### 1. Placement gates on a measurement, and CP and noded disagree on the formula

`noded/server/budget.go:141` computes headroom as `memory.max - working_set`.
`noded/server/pressure.go:108` admits or rejects on it.

The CP gate and the noded gate are **different arithmetic**:

| Side | Predicate |
| --- | --- |
| CP (`Embervm.Placement.mem_eligible?/2`) | `headroom >= need` |
| noded (`memPressured`, `pressure.go:116`) | `headroom >= need + memRejectFloorMib` |

A 2gi brick reports a ~1933 MiB budget. semgrep needs 1536 MiB. It passes the CP
gate (1933 >= 1536) and can never pass noded's (1933 < 2048). That is #4137: an
outage, structural rather than transient, with two public demos down.

`#4101` is the same disagreement between advertise and admit inside noded:
`SlotCeiling()` advertises `memBudget / 512` (3 on a 2gi brick) while every
admission site gates on the raw `cfg.MaxLiveVMs` (`server.go:924`, `:1337`,
`serving.go:37`, `group_member.go:50`). In k8s that gap cannot exist: `maxPods`
is one number both sides read.

Downstream compensations that exist only because the number is not a reservation:

- `memRejectFloorMib`, a 512 MiB fudge cushion.
[...]

PR #4150, opened 2026-07-29
fix(embervm): attribute denials by usable capacity plus the admission floor

jomcgi:
Closes #4145. Phase B PR B1 of #4140. The unfixed half of #4137.

## The defect

`brick_controller.ex` `class_for_need/2` picks which brick class to scale when a
placement is denied. It compared `need_mib` against the class **nameplate**
parsed out of the label (`"2gi"` -> 2048). Two errors stacked:

1. **No admission floor term.** noded admits only at `headroom >= need +
   memRejectFloorMib` (`pressure.go:116`). This is the same defect #4141 fixed in
   the placement filter, still present in the scaling decision.
2. **Nameplate is not usable memory.** `MemBudgetMib()` subtracts
   `DaemonReserveMib` (default 512, `config.go:386`, passed at `server.go:459`).

Correcting the arithmetic, because earlier write-ups on #4140 and #4145 had this
wrong:

| 2gi brick | Value | Subtracts the reserve |
| --- | --- | --- |
| nameplate (what the old code parsed) | 2048 | n/a |
| `MemBudgetMib()` (usable) | **1536** | yes |
| `MemHeadroomMib()` (observed) | ~1933 | **no** |

Cross-check: `SlotCeiling` = 1536/512 = exactly **3**, which is the
`max_live_vms=3` observed in #4101.

So semgrep at 1536 MiB attributed to **2gi**, autoscale added 2gi bricks up to
`maxReplicas: 4`, and **not one of them could ever admit it**. The 16gi class,
the only one that can, was never scaled. The loop could not converge.

#4141 made those 2gi bricks correct non-candidates for *placement*, so they stop
receiving doomed primes. This is the other half: the autoscaler no longer answers
those denials by buying more of the shape that cannot work.

## The fix

**The chart declares usable capacity.** `bricks.daemonReserveMib: 512` is new, and
renders **both** `EMBERVM_NODED_DAEMON_RESERVE_MIB` into the noded pod and the
`usable_mib` arithmetic into `EMBERVM_BRICK_CLASSES`, so the two cannot drift.
Hardcoding 512 in a Helm template would have been the same duplicate-constant
mistake that caused #4141.

Rendered with the default values:

```json
[{"name":"2gi","usable_mib":1536,...},{"name":"4gi","usable_mib":3584,...},
 {"name":"8gi","usable_mib":7680,...},{"name":"16gi","usable_mib":15872,...}]
```

**Declared rather than observed, deliberately.** Denial attribution needs class
capacity **at zero replicas**, when no brick of that class exists to report
anything. A live brick's reported `mem_budget_mib` becomes a drift *check* against
the declared value instead: declared intent, reconciled against node truth, which
is invariant 5's shape.

**The floor comes from the bricks that report it** (#4141's published fact), taking
the largest any registered brick of the class reports and defaulting to 512 when
the class has no replicas. Conservative on purpose: a class that *might* refuse is
not a scale target.

## The subtle one: proto3 has no null, so 0 is not "unset"

The floor is read with an explicit positive check, never `||`:

```elixir
floor when is_integer(floor) and floor > 0 -> floor
_ -> 512
```

`||` catches only `nil` and `false`, and **0 is truthy in Elixir**. That case is
live, not hypothetical: `node_registry.ex:646` copies the value straight off the
decoded proto, and **proto3 decodes an unset `uint64` as 0**, so every daemon
predating that field reports a *present* zero. With `||` the floor would read 0
and attribution would revert to exactly the defect this PR closes.

noded's own `memRejectFloorMib()` (`pressure.go:136-141`) already treats `<= 0` as
"unset, use 512". Reading it literally on the CP side would have been a **fourth**
advertise-versus-enforce divergence, after #4101, #4141 and #4145 itself. The
regression test is named after the bug so a future simplification back to `||`
fails loudly.

## Also

- A class with no `usable_mib` falls back to the nameplate parse, so a CP running
  new code against an older ConfigMap behaves exactly as today rather than
  attributing nowhere.
- A transition-logged warning when a class's declared `usable_mib` disagrees with
  what its bricks report. Observability only, no behaviour change on drift.
- A need no class can serve is now a distinct logged and traced outcome instead of
  a silent `nil`, so "scaling genuinely cannot help" is distinguishable from "no
  denial happened".

## Out of scope, called out rather than slipped in

This deletes `projects/embervm/chart/values.yaml.bak`, an editor artifact
committed by `b3a4ad107`. It is unreferenced (nothing in the repo mentions it),
there is no `.helmignore`, so it currently ships inside the published OCI chart.
Zero risk and trivially revertible, but it is not part of the fix.

## Verification

`ci lint` clean. `ci test` on the rebased tree:

```
MIX TEST OK on the executor
1037 tests, 0 failures, 6 skipped
Executed 1 out of 743 tests: 743 tests pass.
```

The second line is not evidence here and is included only to show it is not:
`Executed N out of M` counts **bazel test targets**, and the control-plane Elixir
suite is not one, it runs inside the `//bazel/erlang:mix_test_smoke` genrule. An
earlier revision of this branch printed `743 tests pass` while the suite reported
`1022 tests, 1 failure`. Only the non-zero exit and a full-log grep caught it. The
count moving 1029 -> 1037 is what shows the new tests actually registered and ran.

That earlier failure was a test of mine that could not discriminate: it asserted
"a positive reported floor overrides the 512 fallback" using a need equal to the
full usable capacity, where **no** positive floor can admit the class, so both
branches produced the same answer. Fixed by choosing a need inside the window
where 256 fits and 512 does not, with the arithmetic recorded in a comment.


jomcgi:
## Helm Manifest Diff

PR #4464, opened 2026-08-07
feat(monolith): show guest VM state in agents console, fix qwen turn status

jomcgi:
Follow-up to #4463, demo-driven.

## Guest VM state chip
The transcript header now shows `vm awake` / `vm asleep` / `vm off`, joined against the EmberVM control plane's own session listing via a new `/api/agents/vms` endpoint (`transport.list_sessions`, management auth), polled every 5s by the console. The control plane parks a session VM 20s (`idleBankSeconds`) after its last invoke without telling the monolith, so the local binding cannot be trusted for liveness; the chip reflects CP truth and degrades to "off"/last-known when the CP is unreachable. Session payloads expose `ember_session_id` for the join.

CP state collapse: `creating/running/relighting` → awake, `banking/banked/parking/parked` → asleep, everything else (incl. absent) → off.

## Qwen turns no longer render as failed
The pi lane passes the model's raw `stopReason` through as `terminal_reason` — `"stop"` for a normal qwen turn — and both the turn card (red box + failed badge) and the backend `_turn_status` (session `warn` pill, amber dots) treated anything but `"completed"` as failure. Both now compare against a clean-reason allowlist (`completed`, `end_turn`, `stop`). `None` still warns: it means the transport died mid-turn (existing test pins this).

Chart bumps: monolith 0.285.460, monolith-public 0.89.385 (shared image).

`ci test`: Executed 61 out of 761 tests: 761 tests pass.

### DIFF
--- a/projects/monolith-public/chart/Chart.yaml
+++ b/projects/monolith-public/chart/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: monolith-public
 description: Read-only public tier of the monolith (ADR 004 Phase 5) and pruned
 type: application
-version: 0.89.384
+version: 0.89.385
 appVersion: "0.1.0"
 maintainers:
   - name: homelab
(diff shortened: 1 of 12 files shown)


jomcgi:
## Helm Manifest Diff

No manifest changes detected across 20 application(s).

PR #4903, opened 2026-08-15
perf(embervm): punch holes in memfiles to reclaim snapshot disk

jomcgi:
Firecracker writes a guest's whole RAM image on snapshot, **fully allocated**. A 4096 memMib workload produces a 4294967296-byte memfile whose `du` reports 4.0G no matter how little the guest touched.

Measured on a live brick, that is nearly all of the snapshot storage:

```
167 memfiles totalling 143.6 GiB
whole snapshots dir:      145.0 G      (~99% is memfiles)
```

Sampled zero fractions, 4 KiB blocks:

| base | memMib | zero | absolute touched |
|---|---:|---:|---:|
| claude-runtime | 4096 | **92%** | ~330 MB |
| sandbox | 512 | **49%** | ~260 MB |

Guests touch a roughly **constant absolute amount** (~250-350 MB) whatever their ceiling, so the zero fraction is `1 - touched/memMib`. Declared `memMib` is billed as disk immediately and in full, which is the opposite of the usual intuition that over-provisioned memory is free until used.

## Why this is safe

`digHoles` uses `FALLOC_FL_PUNCH_HOLE | FALLOC_FL_KEEP_SIZE`, so the **logical size and contents-as-read are unchanged** and only block allocation drops. `KEEP_SIZE` is load-bearing: without it the file is truncated and the snapshot destroyed.

Restore needs no change. The driver restores via `MemBackend{BackendType: "File"}`, which the fcclient documents as "mmaps the image and faults pages lazily", so holes read back as zeros and untouched pages were never faulted anyway. The mapping cannot be writing back, because many microVMs already restore **concurrently from one shared base memfile** (see `RootfsReadOnly` and `CanonicalVsockDir` at `driver.go:45-60`).

A new VM therefore still maps its full `memMib`. Same RAM, less disk.

## Wiring

Every point a memfile is finalized: snapshot, base build, session / serving / stateful / group-member banks, and the stateful checkpoint. All call sites go through a wrapper that **returns nothing**, so a punch failure structurally cannot fail a bank, fail a base build, or strand a paused VM. Bytes freed are logged at info so the win is observable in prod.

Punching happens after `Resume` rather than before: the guest is paused for less time, and a finalized memfile is not read by the running guest either way.

## Verification

Run with the focused target under `--nocache_test_results --test_output=all`, so the tests **provably executed** rather than being inferred from an aggregate count:

```
Executed 1 out of 1 test: 1 test passes
```

Tests assert the property that matters most, that a mixed file's bytes are **identical** after punching, plus unchanged logical size and strictly decreasing `st_blocks`, with all-zero / all-non-zero / sub-block edge cases. Full suite: `Executed 1 out of 716 tests: 716 tests pass`.

## Scope

This does **not** help the S3 path. `store.Put` takes the file's logical size and `io.Copy` reads holes back as zeros, so an upload still ships the full byte count. Compressing that path is separate follow-up work, and it also fixes hydration bandwidth which punching cannot.

Savings degrade over a session's life, since Linux does not zero freed pages, so touched-then-freed memory stays non-zero. Base bundles are the best case (pristine boot, built once) and are also the largest measured cost.

## Related

While measuring this I found **149 stateful bundles totalling 69.9 G (48% of the snapshots dir), all created Jul 17-22 and never reaped**, against a 7-day `bankedTtlSeconds`. That is a separate GC issue, not addressed here, and is likely a larger and cheaper win than this PR.

### DIFF
--- a/projects/embervm/noded/fcvm/driver/BUILD
+++ b/projects/embervm/noded/fcvm/driver/BUILD
@@ -8,10 +8,10 @@ go_library(
         "launcher.go",
         "mountns_linux.go",
         "mountns_other.go",
+        "provisioner.go",
         "punchholes.go",
         "punchholes_linux.go",
         "punchholes_other.go",
-        "provisioner.go",
     ],
     importpath = "github.com/jomcgi/homelab/projects/embervm/noded/fcvm/driver",
     visibility = ["//projects/embervm/noded:__subpackages__"],
@@ -19,8 +19,15 @@ go_library(
         "//projects/embervm/noded/fcvm/fcclient",
         "//projects/embervm/noded/substrate",
         "@io_opentelemetry_go_otel//:otel",
-        "@org_golang_x_sys//unix",
-    ],
+    ] + select({
+        "@rules_go//go/platform:android": [
+            "@org_golang_x_sys//unix",
+        ],
+        "@rules_go//go/platform:linux": [
+            "@org_golang_x_sys//unix",
+        ],
+        "//conditions:default": [],
+    }),
 )
 
 go_test(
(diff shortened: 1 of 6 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4182, opened 2026-07-31
embervm: eager-local rootfs (OCI, Account chunk store, ublk), ADR embervm/028


jomcgi:
Implements [ADR 028](https://github.com/jomcgi/homelab/blob/main/docs/decisions/embervm/028-demand-loaded-rootfs-oci-chunk-store.md). The eager-local and Account-scoped amendment is in #4902.

The destination is: resolve OCI refs to immutable per-architecture digests, convert each image to a deterministic flattened filesystem, split it into reusable chunks, encrypt private chunks at Account scope, hydrate the complete active manifest onto brick NVMe before READY, and present the local chunks as a read-only ublk device. ublk never reads from the object store for a live guest. Mutable memory, workspaces, and volumes remain randomly encrypted per artifact and never deduplicate across principals.

## Phase 0: measure and freeze the representation

- [x] Produce deterministic candidate EROFS images for the current apko workloads and prove byte-for-byte reproducibility with the exact tool version and flags pinned.
- [x] Run the `projects/embervm/rootfs/measure_chunks.py` harness from #4902 over a one-package rebuild, comparing `fixed-v1` and `gear-v1`, then choose chunk parameters.
- [x] Measure per-image, per-principal, Account, and global physical footprints across the image set. Use the result to decide whether the optional published platform universe earns its provenance and GC machinery.
- [x] Inventory the active manifest set per brick and size the local encrypted chunk cache from complete manifests, not the full catalogue.
- [ ] Measure SeaweedFS hydration throughput and tail latency at bounded fresh-brick concurrency. This sizes preparation and admission, not a live block-fault path.
- [ ] Run the Firecracker digest-symlink experiment and a live ublk backend failure experiment before freezing ublk presentation.
- [ ] Probe `CONFIG_BLK_DEV_UBLK`, `/dev/ublk-control`, capabilities, io_uring, and Firecracker jailer compatibility on homelab nodes and the EKS AMI.

## Pre-EKS work items

- [x] Version the bank-bundle schema and record the ext4 rootfs digest. Shipped in #4199.
- [ ] Finish rootfs provenance threading through the control plane and make GC liveness include registry entries, READY bases, live VMs, and bank bundles.
[...]

PR #4904, opened 2026-08-15
docs(embervm): record ADR 028 Phase 0 measurement results

jomcgi:
## Summary

Discharges the ADR embervm/028 Phase 0 measurement gate with reproducible evidence, and commits the raw harness reports beside the harness.

**Stacked on #4902**, which adds the harness this PR runs. Base is `feat/embervm-eager-rootfs-chunks`, so the diff here is two commits. Retarget to `main` once #4902 merges.

## What was measured

Six x86_64 guest images, re-resolved live from all nine noded brick Deployments rather than taken from the digests recorded on 2026-08-14. All bricks carry the same set, and the compressed total matches the earlier 1,761,806,814 bytes exactly.

## Results

**Determinism (gate 1): PASS.** Every candidate EROFS built twice from the same tar is byte-identical. `mkfs.erofs` 1.8.2, with `--tar=f -b 4096 -U <fixed> -T 0 --mkfs-time --sort=path` and no compression. Dropping `-U` alone breaks byte equality.

**The ADR's estimate holds (gates 2 and 3): PASS.** Account-scope Gear CDC saves 626.5 MiB, 16.771%, of the 3.648 GiB flattened set. The planning estimate was 550 to 850 MiB, 13% to 20%. Fixed-offset chunking saves 0.147%, so content-defined chunking is the whole mechanism rather than an optimisation.

Changing format alone, sparse ext4 to flattened EROFS, recovers a further 507.1 MiB (11.95%) before any dedup. That is reported separately and not counted as a dedup result.

**The one-package rebuild found the result that matters most (gate 4).** Gear re-stores 47.6% of the rebuilt image where fixed re-stores 73.4%. That is a bad number for a one megabyte change, so it was decomposed rather than accepted:

| What changed | Re-stored |
| --- | ---: |
| content only, same name and size | 0.36% |
| content plus 1,111 bytes of growth | 47.60% |
| content plus growth plus rename | 47.60% |

The chunker is sound (98.2% shared after a pure insertion, 2.3% of cuts at maximum) and the rename is free. **File size growth is the cause**: `mkfs.erofs` inlines each file's tail into the inode region, so growing one file re-packs the layout of everything after it, non-uniformly, which is what CDC cannot resynchronise against.

Building with `-Enoinline_data` drops the rebuild cost from **47.60% to 0.47%**. The whole six-image set was rebuilt with the flag and re-measured, so the trade is quantified rather than assumed:

| Layout | Average | Logical | Account stored | Saved |
| --- | ---: | ---: | ---: | ---: |
| default | 256 KiB | 3,917,045,760 | 3,163,875,389 | 19.228% |
| `-Enoinline_data` | 256 KiB | 4,017,766,400 | 3,199,529,586 | **20.365%** |
| default | 512 KiB | 3,917,045,760 | 3,260,105,750 | 16.771% |
| `-Enoinline_data` | 512 KiB | 4,017,766,400 | 3,313,293,738 | 17.534% |

Alignment improves the ratio but does not fully repay the padding: in absolute stored bytes it costs 35,654,197 more (1.13%) at the recommended parameters. **That buys back roughly 193 MiB on every single-image rebuild, forever.** A real package upgrade almost always changes file sizes, so the converter must set it, and that decision belongs in the ADR before the format freezes.

**Active-set inventory (gate 5): PASS, and it is the larger number.** Every brick builds bases for six workloads that resolve to five distinct images. `runtime-python` is baked on every brick and activated on none, costing 508.4 MiB each. Bricks also retain superseded rootfs files forever, so a long-lived brick holds 6.934 GiB rather than the 4.143 GiB recorded for a fresh one, and four of its ten roots are digests GHCR no longer serves.

**CDC parameters (gate 6): 64 KiB / 256 KiB / 1 MiB recommended**, from a monotonic sweep (21.068% at 128 KiB down to 14.487% at 1 MiB) with clearly diminishing returns. The ordering survives block alignment.

**Published platform universe (gate 7): recommend against for now.** One Account exists, so `global` and `account` are identical by construction and there is no measured cross-Account demand. The full 16.771% is Account-internal and does not depend on the universe existing.

**ublk (gate 8): OPEN, inconclusive.** `ublk` is in no node's `/proc/modules`, but so is `loop` while `/dev/loop7` is mounted, so absence there proves nothing. Recorded as open rather than inferred. The Firecracker digest-symlink experiment needs a live VM and stays open; `PatchDrive` exists but is only ever called for the `volume` drive, never `rootfs`.

## Validation

- `git diff --check`
- pre-commit changed-file lint passed on both commits
- pre-push `ci test` (bb remote, 1:1 Workflows Test) passed
- PR CI green on the first commit: pr-checks, GitGuardian, Semgrep
- documentation and JSON only, no code changes

Bazel was not run directly on the development machine, per repository AGENTS.md.

Relates to #4182.


jomcgi:
## Helm Manifest Diff

PR #4905, opened 2026-08-15
fix(embervm): let the warmth reaper reclaim pre-sidecar stateful orphans

jomcgi:
The WarmthReaper skipped every stateful bundle whose on-disk workload binding was empty:

```elixir
defp evict_entry(_state, %{kind: :stateful, id: ref, workload: workload}) when workload in [nil, ""] do
  # ...the bundle would be locally gone yet its S3 copy stranded FOREVER with no
  # handle to find it again (Fable F2, worse than nothing)... Reclaiming these
  # pre-sidecar orphans is a separate decision (task #39).
```

The reasoning is sound. The **premise was never tested**: it assumed a remote copy *might* exist and never looked.

## Measured in production today

| | |
|---|---|
| orphaned bundles on one brick | **149**, totalling **69.9 GiB** |
| share of that brick's snapshots dir | **48%** |
| created | Jul 17-22, never reclaimed |
| stateful bundles in S3, total | **7**, all with proper bindings |
| **overlap with the 149** | **ZERO** |

So the skip protected nothing while costing 70 GiB indefinitely, and it could never resolve itself, because nothing else was going to supply the missing binding.

## The store has what the bundle lost

A bundle with an empty binding cannot **compose** a remote prefix, but it can still be **searched for by ref**, and a hit at `stateful/<vendor>/<workload>/<ref>` yields exactly the vendor and workload the local copy lost.

Three branches:

| case | action |
|---|---|
| **found remotely** | recover the binding, evict **both** copies, remote only after local succeeds. Repairs the binding rather than working around it. |
| **not found** | local copy is the only copy: evict local only. Nothing can be stranded. |
| **lookup failed** | keep the previous skip exactly. An unavailable store must never read as "no remote copy exists". |

## Safety properties

- **Fail-safe preserved.** A store error, timeout, or unset endpoint takes the skip branch, never the delete branch.
- **One listing per sweep**, and only when some entry actually has an empty binding, so 149 orphans do not become 149 listings.
- **A raising or throwing store** becomes the lookup-failed branch rather than crashing the GenServer.
- **The destructive gate is unchanged.** With `EMBERVM_WARMTH_RETENTION_SWEEP` off, the sweep still deletes nothing and logs what it would do in each branch.
- **Group sets keep the conservative skip.** Their remote layout is keyed by `group_instance_id` rather than the set ref, so the same search does not apply. Deliberately untouched.

## Verification

`ci test` green. Note the counter reads `Executed 0 out of 716 tests` because the Elixir suite runs as the `//bazel/erlang:mix_test_smoke` **genrule** (a build action) rather than a bazel test target: the log confirms `From Executing genrule //bazel/erlang:mix_test_smoke` with the build completing successfully, and a failure there fails the build.

The 70 GiB on the affected brick was reclaimed by hand before this PR (verified: all workloads still `READY`, `demo-postgres` healthy, zero control-plane errors). This change is what stops it recurring.

### DIFF
--- a/projects/embervm/control/lib/embervm/warmth_reaper.ex
+++ b/projects/embervm/control/lib/embervm/warmth_reaper.ex
@@ -72,7 +72,7 @@ defmodule Embervm.WarmthReaper do
   use GenServer
   require Logger
 
-  alias Embervm.{GroupState, GroupStore, NodeCapacity, StatefulState, StatefulStore, WakeInstance}
+  alias Embervm.{GroupState, GroupStore, NodeCapacity, S3Client, StatefulState, StatefulStore, WakeInstance}
 
   alias Embervm.Node.V1.{
     ArtifactRef,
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

PR #4906, opened 2026-08-15
perf(embervm): zstd store objects to cut S3 bytes and hydration time

jomcgi:
Nothing on the store path was compressed: `Export` passed a file's logical size and `io.Copy` streamed it verbatim.

That matters because memfiles dominate this system's storage and transfer. Measured on a live brick:

```
167 memfiles totalling 143.6 GiB   (~99% of a 145 GB snapshots dir)
claude-runtime base (4096 memMib):  92% zero
sandbox base        (512 memMib):   49% zero
```

Guests touch a roughly constant ~300 MB whatever their ceiling, so most of what is stored and shipped is zeros.

Hole-punching (#4903) does **not** help here: the upload reads holes back as zeros and ships the full byte count. Compression is the layer that does, and this path suits it. The local memfile must stay mmap-able for lazy-fault restore, but a store object is always downloaded to disk before use, so decoding during that download costs little. It cuts stored bytes **and cross-brick hydration latency**, felt every time a workload lands on a node without the base.

## The integrity invariant

`meta.json` keeps describing the **plaintext**. `restoreFile` verifies received bytes against `want.Size` and `want.Sha256`, so the decoder wraps the body **before** the hashing `MultiWriter`:

```go
if want.Compression == "zstd" { decoder, _ = zstd.NewReader(body); decoded = decoder }
n, err := io.Copy(io.MultiWriter(f, h), decoded)
```

Those checks therefore still verify the real artifact rather than its encoding, and what lands on local disk is byte-identical to what was exported.

## Two design choices worth flagging

**The marker lives in `FileMeta`, not the object key.** `s3_warmth_gc.ex` and the warmth reaper match on key prefixes and refs, so a `.zst` suffix would silently break the GC (see #4905 for how much depends on that matching).

**Encode to a temp file.** S3 needs Content-Length up front and a streaming encoder does not know the compressed size in advance, so the source is encoded alongside itself, stat'd, uploaded, and removed, with cleanup on every error path.

## Rollout: the writer ships OFF

`EMBERVM_NODED_STORE_COMPRESS` / `noded.store.compress`, default **false**.

This is deliberately two-phase. A daemon that does not understand the marker cannot restore a compressed object, so phase 1 puts the **reader** everywhere and leaves the writer disarmed. Arming both at once creates a mixed fleet where new objects are unhydratable by old nodes, which is a self-inflicted outage. Same shape as the WarmthReaper's `EMBERVM_WARMTH_RETENTION_SWEEP` gate and the suspended synthetic in #4901.

Objects written before this change carry no marker and restore exactly as today.

## Verification

`ci test`: `Executed 6 out of 716 tests: 716 tests pass`. Tests cover compressed and uncompressed round trips, a **legacy object with an empty marker** (the rollout-safety case), a truncated compressed body failing rather than writing a partial file, and both zero-heavy and incompressible payloads.


jomcgi:
## Helm Manifest Diff

Issue #4893, opened 2026-08-14
embervm: snapshot_ref conflates "present on this node" with "fetchable from the store", at three layers


jomcgi:
Consolidates the two defects from #4865 that #4891 does NOT fix, plus the reason they keep recurring.

## The pattern

`snapshot_ref` is a single string that different call sites read as two different claims:

- **present here**: this node has the base on local disk and can boot it
- **fetchable anywhere**: the store holds it, so any node can hydrate it

Those came apart on 2026-08-14 and stayed apart for six hours. Three independent sites, written at different times, each picked one meaning and was right about half the time.

## Site 1: the base existed and was never exported

`demo-postgres__ec8abc5c1149` was built on node-4 at 14:36 during the chart 0.12.9 roll:

```
snapshots/bases/demo-postgres__ec8abc5c1149/
  imageref  memfile  rootfspath  snapfile
  <- no meta.json, so never exported
```

Present here: true. Fetchable anywhere: false. The control plane recorded it as the workload's current base regardless.

## Site 2: `export_reconcile/1` only inspects the anchor

```elixir
cur = current_instance_on_node(acc, w.node_id, w.name)
if current_base_present_but_unexported?(acc, cur, w) do
```

`w.node_id` is the workload's ANCHOR (node-3). The base was on node-4. The self-healing export backstop therefore never looked at the node actually holding the unexported base, and could not have fixed this no matter how long it ran.

The fix has to consider every node reporting the current ref, not just the anchor. Note the export path also failed at the transport layer when driven manually (`{:server_closed_request, :cancel}` on Mint.HTTP2, with node-3 logging nothing), which is worth understanding before trusting a broader reconcile to succeed.

## Site 3: vendor-blind selection

The scalar collapses a per-vendor map. The live CR carried:

```
snapshotRef:  demo-postgres__ec8abc5c1149
snapshotRefs: {"amd": ["demo-postgres__ec8abc5c1149"], "intel": ["demo-postgres__943a4c135461"]}
```

The scalar is the **amd** ref; the anchor node-3 is **Intel** (`cpu-model.vendor_id = Intel`). So even a correct hydrate would have asked for the wrong vendor's artifact. Worse, the ref it should have wanted, `943a4c135461`, exists on no node either.

A branch exists for this (`fix/embervm-vendor-matched-hydrate`). It is necessary and, on its own, would not have restored service: the correct ref was equally unobtainable.

## Site 4, mine, for the record

[...]

Issue #4865, opened 2026-08-14
embervm: demo-postgres never rehydrates after a fleet restart, jomcgi.dev/health has been 503 since 14:57


jomcgi:
## Symptom

`jomcgi.dev/health` returns 503 with:

```json
{"status":"unhealthy","backendStatus":503,"failing":["ember_postgres"]}
```

Continuously from roughly 14:57 UTC on 2026-08-14. It was 200 earlier the same session. Twelve consecutive polls over 17 minutes, spanning four synthetic probe cycles, all 503. It is not flapping and it is not recovering on its own.

## What is actually happening

The control plane is in a wake-fails-then-rehydrate loop for `demo-postgres`, roughly every 15 seconds:

```
embervm stateful wake failed                 reason=:no_eligible_instance  workload=demo-postgres
embervm base builder: hydrating demo-postgres onto node-3/535f32b4... after a wake found no base on its anchor node node-3
embervm base builder: hydrate demo-postgres/demo-postgres__ec8abc5c1149 fell back, forcing a rebuild (:not_present_in_store)
embervm base builder: forcing rebuild for demo-postgres, cleared snapshot ref "demo-postgres__ec8abc5c1149"
```

Each pass clears the snapshot ref, and the next wake finds no base again. The loop does not converge.

The placement line names the reason:

```
embervm wake: no cold-placement candidate on node need_mib=512 capacity_denial=false brick_count=1
bricks=[{id=node-3/535f32b4... class=2gi free=16 hr=2038 budget=1536 eligible=true base_ready=false workload_facts=1 base_state=nil}]
```

`base_ready=false`, `base_state=nil`. Note also `free=16` against `budget=1536` with `need_mib=512`: the brick is nearly out of budget even though `capacity_denial=false`, so if the base problem is fixed the placement may still not fit.

## Probable trigger

Every EmberVM pod restarted within a 16 minute window today:

| pod | started |
|---|---|
| control plane | 14:50:46 |
| brick-16gi | 14:38:35 |
| brick-16gi-node-4 | 14:50:38 |
| brick-2gi | 14:35:18 |
| brick-2gi-node-1 | 14:38:46 |
| brick-2gi-node-2 | 14:42:50 |
| brick-2gi-node-3 | 14:46:35 |

[...]

PR #4907, opened 2026-08-15
fix(embervm): let retention collect orphans when the current base is unexported

### LINKED ISSUE #4364: embervm: prerequisites before re-arming local base retention (banked-generation keep-set, manifest fidelity)
Blockers that must clear before `baseRetention.sweepEnabled` is set back to `"1"`. PR #4362 disarms it and makes the sweep emit a readable manifest; this issue is the list of things to fix or verify before that gate goes back on. Found in the #4362 review.

## 1. A banked session's base can be evicted after a control-plane restart (the important one)
noded's `evictBaseLocal` guards only against (a) a live VM restored from the ref and (b) a BUILDING base; `store.go:876` states outright that there is deliberately no registry-membership or current-base guard. A banked session has no live VM, so guard (a) never fires.

Control-plane protection is `w.base_refs`, which is populated ONLY on in-process turnover, initialised empty, and never rehydrated on boot. `report_base_refs/3` has no production caller at all, so refcounts stay nil forever.

Failure path: base A turns over to B; a session banks pinned to generation A (the durability ladder pins warm banks to CPU-vendor + base-generation for 7 days); the CP restarts for any reason (deploy, chart bump, OOM); `base_refs` is empty so the desired set is just [B]; A is READY, not BUILDING, and older than the min-age rail, so it is a candidate and gets evicted. That session cannot warm-restore for its remaining ~6 days.

Fix direction: derive the keep-set from durable state (the session store / op-log) rather than in-process memory, so a CP restart cannot forget which generations are still pinned. The 1-hour min-age rail is ~168x shorter than the 7-day ladder it needs to respect, so age alone cannot substitute.

## 2. Reclaimable-bytes understates, often to zero, for exactly the orphans this targets
`server.go:1885-1889`: a disk-only entry that fails `isCompleteBase` reports `SizeBytes: 0`, and a `.tmp` orphan from a build that died mid-write always fails it. The summary can read "N candidates, 0 bytes reclaimable" while gigabytes are staged, which reads as "not worth arming" and is the opposite of true. Pre-existing sizing behaviour, but it lands in the new headline number.

## 3. `reason_unreferenced` is a constant string
It does not distinguish "superseded by a newer generation on a known workload" from "the control plane has no Workload entry at all", and the second is materially riskier. Per-base specificity is the point of the manifest.

## 4. "covers N nodes" overstates
`log_retention_summary/2` counts every node in `capacity_facts_by_node/1`, including nodes skipped for an unexported current base or for not reporting the workload.

## 5. Unknown-workload path has no newest-generation protection
`unknown_workload_retention/1` targets workloads absent from `state.workloads`, where ALL refs including the newest are candidates, protected only by `protected_refs` and the age rail. Arguably by design, since the workload is gone, but it should be a deliberate decision recorded here rather than an emergent one, and the manifest should say when a candidate came from this path.

## Note on the #4290 framing
The "0 candidate(s)" log quoted in #4290 comes from `log_remote_retention_summary/2`, which is the S3 STORE sweep, not the node-disk sweep. On-disk enumeration already existed for the local sweep (`localBasesStatus` has scanned `bases/` since dc934d507). The genuine delta in #4362 is a min-age rail, a per-sweep cap, and the first per-sweep summary the local sweep has ever had. Worth keeping straight when judging whether arming actually reclaims the scratch that #4290 describes.

Refs #4290, #4362.

jomcgi:
Closes #4364.

## The defect

The base retention sweep has reported `0 candidates, 0 bytes evicted` on every cycle for weeks, while orphaned bases accumulated on disk. It is armed, it runs, it sees the garbage, and it collects nothing.

The new per-node accounting from the 0.13.7 work makes it arithmetically visible. One sweep:

```
node-1  seen_on_disk=1  desired=1  too_young=0  refcount_protected=0  candidates=0   consistent
node-2  seen_on_disk=3  desired=3  too_young=0  refcount_protected=0  candidates=0   consistent
node-4  seen_on_disk=6  desired=6  too_young=0  refcount_protected=0  candidates=0   consistent
node-3  seen_on_disk=3  desired=2  too_young=0  refcount_protected=0  candidates=0   CONTRADICTION
```

node-3 sees three bases and wants two. The surplus is excluded by neither the age rail nor a refcount, and is still not selected. Every stated reason to spare a base reads zero and it is spared anyway.

## Cause

The node-level `:skip_unexported` arm hardcoded the candidate list to empty:

```elixir
get_in(fact, [:workloads, w.name, :exported]) != true ->
  {:skip_unexported, fact[:instance_id], retention_accounting(local_bases, desired, w.base_refs, [])}
```

So one workload's current base being unexported spared **every** local base on that node.

That guard was deliberate (dc934d507: "durability-before-eviction"). The problem is it never releases: `export_reconcile/1` only inspects the anchor node rather than whichever node holds the base (#4893), so the current base frequently never gets exported at all. On node-3 both `demo-postgres` bases are unexported, including `ba4ef0bad08b` which the #4865 recovery had just built.

Guard plus export bug equals a leak that grows without bound.

## Why removed rather than narrowed

A candidate already passes four per-base filters:

```elixir
b.base_state != :BASE_BUILD_STATE_BUILDING,
b.ref not in desired,                        # never the CURRENT base
not base_protected?(w.base_refs, b.ref),     # no session or primed VM rides it
base_age_seconds(b) >= @retention_min_age_seconds
```

So a candidate is a superseded artifact that nothing references. **Deleting it cannot reduce the durability of the current base, because they are different files.** The guard was protecting a roll-back-to-a-superseded-base scenario that is not a supported operation, and charging a permanent leak for it.

This conflation is the same one tracked in #4893: "the current base is not in the store" was being used to answer "may I delete a different, unreferenced, superseded file from local disk". Two unrelated facts, one field. That is now its third confirmed instance, after the builder looping on an unfetchable base and `export_reconcile` checking the anchor instead of the holder.

## The safety argument now rests entirely on the per-base filters

So each one gets a test asserting it still spares its case **while the current base is unexported**: the current ref, a refcounted ref, a BUILDING base, and a too-young base. None of the filters were touched.

## Verification

Red first, against unmodified `base_builder.ex`:

```
1) test retention sweep and export per node when fleet has mixed vendors
2) test retention sweep reclaims an orphan when the current base is not exported
3) test retention filters protect every non-current base safety rail when current is unexported
1303 tests, 3 failures, 6 skipped
```

Exactly the two tests that encoded the removed guard, plus the new safety-rail test, and nothing else. Green with the fix: `1303 tests, 0 failures, 6 skipped`. Both runs executed on this branch.

Two existing tests are inverted rather than weakened. `"retention sweep skips a workload whose current base is not yet exported (durability floor)"` asserted the behaviour this PR deliberately removes, so it had to change; leaving it green would have meant the guard was still there.

## Expected effect after deploy

node-3's `demo-postgres__8dde24352f92` becomes a candidate and is evicted, and the masters' scratch should fall for the first time since the sweep was armed on 2026-08-05. Worth watching the first armed sweep after rollout and reading `bytes evicted` rather than assuming.

### DIFF
--- a/projects/embervm/control/lib/embervm/base_builder.ex
+++ b/projects/embervm/control/lib/embervm/base_builder.ex
@@ -1350,8 +1350,7 @@ defmodule Embervm.BaseBuilder do
   # scratch and survives a noded pod restart, but the build-time placement instance_id
   # churns with the pod. Without this remap `find_capacity_fact`/`node_addr` miss after
   # any pod restart (the churned pod_uid no longer matches w.node_id), which silently
-  # parks the durability export AND the local retention drain: the current base is
-  # never seen as exportable, so the sweep skips every workload as unexported.
+  # parks the durability export and local retention drain until the node reports again.
   defp current_instance_on_node(state, stale_id, workload) do
     target = node_name(stale_id)
 
@@ -2101,13 +2100,13 @@ defmodule Embervm.BaseBuilder do
   # gated action. The plan is returned for tests/visibility; the action is the
   # spawned evictions (gate on) or the dry-run log (gate off).
   #
-  # For each node with a current, exported local base, candidates are local bases
+  # For each node with a reported current local base, candidates are local bases
   # minus the node's current ref and still-refcounted superseded refs. Base ownership
   # is a NODE property because bases live on shared hostPath scratch, so each node is
   # considered once even when several bricks report the same local inventory. BUILDING
-  # local bases are never candidates, and an unexported current base only blocks
-  # eviction on that same node. noded's in-use/current/BUILDING guard is the final
-  # backstop beneath this.
+  # local bases are never candidates. The per-base desired, refcount, BUILDING,
+  # and age filters make each candidate non-current and unreferenced by
+  # construction. This avoids the old durability guard deadlocking against #4893.
   defp retention_sweep_plan(state) do
     state = %{state | retention_sweep_evictions: 0}
     {plan, state} = Enum.reduce(state.workloads, {[], state}, fn {name, w}, {plan, acc} ->
@@ -2121,11 +2120,6 @@ defmodule Embervm.BaseBuilder do
           {:skip, node_id, accounting} ->
             entry = %{workload: name, evict_refs: [], evict_bytes: 0, candidates: [],
               skipped_unexported: false, node_id: node_id, retention_accounting: accounting}
-            {[entry | p], a}
-
-          {:skip_unexported, node_id, accounting} ->
-            entry = %{workload: name, evict_refs: [], evict_bytes: 0, skipped_unexported: true,
-              node_id: node_id, retention_accounting: accounting}
             {[entry | p], a}
 
           {:evict, node_id, refs, bytes, manifest, accounting} ->
@@ -2154,7 +2148,6 @@ defmodule Embervm.BaseBuilder do
 
   # Decide one retention outcome per node from the representative reported facts:
   #   :skip                          - no local base or nothing to evict
-  #   {:skip_unexported, node_id}    - this node's current base is not exported
   #   {:evict, node_id, refs, bytes} - this node's local bases to evict + bytes
   defp workload_retention(state, w) do
     for fact <- representative_facts_by_node(state, w.name),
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

PR #4908, opened 2026-08-15
fix: converge the five ServerSideApply apps and unblock the tokenbroker scrape

jomcgi:
Clears the 5 OutOfSync ArgoCD applications and the firing Hubble POLICY_DENIED alert. Both were long-standing and neither was breaking anything user-visible, which is why they had persisted.

## The OutOfSync five

`argocd`, `authentik`, `kargo`, `context-forge-gateway` and `kyverno` were all OutOfSync-but-Healthy, with selfHeal re-syncing forever and never converging. Every differing field turned out to be cosmetic.

The five share exactly one property that no Synced app has, which is what identified the cause: **`ServerSideApply=true`**.

SSA diffs on field *ownership*. A field the chart never declares is absent from ArgoCD's predicted state but present in live once the apiserver defaults it, so the gap can never close. Client-side apply hides this: its 3-way merge treats an undeclared field as someone else's. That is why `longhorn`, `signoz` and `monolith` stayed Synced against byte-identical live objects rendered from the same library.

Two classes, two fixes:

| Class | Where | Fix |
|---|---|---|
| Chart omits a field the schema defaults | Gateway API `parentRefs.group/kind`, `backendRefs.group/kind/weight`; Envoy Gateway `remoteJWKS.cacheDuration` | Declare it in the chart |
| Chart emits `{}` the apiserver drops | upstream Kyverno CRD `annotations` / `labels` | Global normalizer (nothing to fix at source) |

`SecurityPolicy.targetRefs` already declared its `group`/`kind` and showed no drift, while `cacheDuration` on the same object drifted. That is what proved declaring is sufficient under SSA, before any change was written.

### Declaring, not ignoring

The `argocd` Application already carried an app-level `ignoreDifferences` for exactly these fields, and it worked. It is removed here anyway, because its list included `.spec.rules[].backendRefs[].weight`, which is live traffic shaping: a deliberate weight change in a chart would have been silently dropped while the Application still reported Synced. That is the same failure mode this repo already hit with `/spec/replicas` (PR #4168, where a merge went green and changed nothing). Suppression is quiet, not safe.

This also corrects the runbook, which stated app-level `ignoreDifferences` is inert. That was observed on v3.1.6; on v3.1.8 it demonstrably works.

## The Hubble alert

[...]

PR #4909, opened 2026-08-15
fix(argocd): put the CRD normalizer under cm: where the chart reads it

jomcgi:
Follow-up to #4908. Four of the five apps converged; `kyverno` did not, because the normalizer I added landed at the wrong nesting level.

It sat as a sibling of `cm:` instead of a child, so it rendered as `configs.resource.customizations...`, a key the argo-cd chart reads nothing from. No ConfigMap entry was produced at all.

## Why nothing caught it

This is worth recording because every signal was green:

- valid YAML
- `ci lint` green, `ci test` green (716 pass)
- all three PR checks green
- the `argocd` Application reported **Synced**, and correctly so, because the manifest it applied genuinely did match git

The only observable was the absent key in the live ConfigMap. `kubectl get cm argocd-cm -n argocd` is the check that would have caught it, and "the app is Synced" is specifically not.

## Verification

Rendered the ConfigMap this time rather than trusting sync status:

```
RENDERED KEY: resource.customizations.ignoreDifferences.apiextensions.k8s.io_CustomResourceDefinition
```

And checked the jq expressions against the real `managed-resources` states pulled from the cluster:

| Input | Expected | Result |
|---|---|---|
| predicted (chart renders `{}`) | match, so the empty map is deleted | matches |
| live (key absent) | no match, no-op | no match |
| map with a real annotation | no match, drift still visible | no match |

That last row is the safety property: the map is deleted only when already empty, so this cannot mask real annotation or label drift.

### DIFF
--- a/projects/platform/argocd/values.yaml
+++ b/projects/platform/argocd/values.yaml
@@ -60,21 +60,30 @@ argo-cd:
           #
           # Fleet-wide was never safe, only quiet. Keep replica ignores scoped
           # to the specific Application whose controller owns the field.
-    # The same {} != absent asymmetry as the Deployment annotations above, but
-    # arriving from the opposite direction. The upstream Kyverno chart renders
-    # `annotations: {}` and `labels: {}` on its CRDs; the apiserver drops an
-    # empty map on write, so live has no key at all while the predicted state
-    # keeps {}. Eleven policies.kyverno.io CRDs sat OutOfSync on nothing else.
-    #
-    # Unlike a field-level ignore this cannot mask real drift: the map is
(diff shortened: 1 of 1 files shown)


jomcgi:
## Helm Manifest Diff

<details>
<summary><code>argocd</code></summary>

```diff

data  (v1/ConfigMap/argocd/argocd-cm)
  + one map entry added:
    resource.customizations.ignoreDifferences.apiextensions.k8s.io_CustomResourceDefinition: |
      jqPathExpressions:
        - '.metadata.annotations | select(. != null and length == 0)'
        - '.metadata.labels | select(. != null and length == 0)'
      
    
  
```

</details>

PR #4911, opened 2026-08-15
fix(embervm): stream store compression so it never writes into the artifact dir

jomcgi:
Follow-up to #4906, which merged before its review returned. The review found one blocker, latent because the writer ships off (`noded.store.compress` defaults false). This fixes it before anyone arms the writer.

## The defect

`compressFile` staged each compressed file with `os.CreateTemp(localDir, "."+base+".zstd-*")`, writing **into the artifact directory**. `enumerateArtifactFiles` (`noded/server/store.go:379`) decides what a file IS and skips only a `.tmp` suffix, `meta.json`, and the retirement intent file. The temp matches none of them.

| # | Consequence |
|---|---|
| 1 | A crash, OOM, eviction, or reboot during a multi-minute base export leaves the temp with no sweeper. The next export hashes it into `meta.Files`, uploads it, and writes it into `meta.json`, after which it is a **permanent member of the artifact**, restored onto every fresh node, inflating exactly the bytes compression exists to cut. |
| 2 | **No crash needed.** The async export queue dedupes by prefix, but the synchronous `ExportArtifact` path (every non-BASE kind) never consults that set. A sync export can enumerate the async worker's temp mid-write, so `fileMeta` hashes a moving file and the recorded sha256 mismatches the uploaded bytes, making the artifact copy **unrestorable** until content changes. |
| 3 | A crash between the file PUT and the meta PUT orphans an S3 object under a unique key that no later export overwrites and no GC ever names. |
| 4 | The `ErrStaleGeneration` fence does not help: a poisoned meta carries an equal generation and passes cleanly. |

## The fix

Stream the zstd encode through an `io.Pipe` straight into the PUT. No temp file exists to leak, enumerate, or sweep. `Put` already supports this: its contract documents that a negative size lets net/http chunk the body.

**Why streaming rather than a skip pattern in `enumerateArtifactFiles`.** The invariant then holds by construction instead of by two packages continuing to agree. The producer lives in the store package and the filter in the server package, and that drift is what created the defect. Renaming the temp to `*.tmp` was rejected separately: `isCompleteBase` returns false on any `.tmp` file in a base dir, so it would report a base incomplete with `sizeBytes: 0` for the whole export window and could trigger a spurious rebuild.

An audit read `noded/server/store.go` and `noded/store/store.go` end to end and confirmed `enumerateArtifactFiles` is the **only** file-level artifact enumerator in noded, so no other package needs changing. Every other walker was checked against the pattern: `isCompleteBase` ignores it, the `Reconcile*`/`Scan*` walkers skip non-dir entries and stat fixed names, rootfs GC requires a `rootfs-` prefix, the retirement sweep matches only `.retirement-intent`, and volume `Scan` keys on `vol.img`.

## `countingReader` implements `Close` deliberately

net/http wraps a plain `io.Reader` body in `io.NopCloser`, so a counting reader that only implemented `Read` would **swallow the underlying Close**. On the compressed path the body is an `*io.PipeReader`, and the transport closing it is the only thing that unblocks the encoder goroutine when a request is aborted mid-body. Without that method the goroutine blocks forever on `pw.Write`, leaking itself and its encoder buffers on every failed export. `TestCompressedExportPutFailureLeavesDirClean` exercises that exact path and passes either way, since Go tests do not observe goroutine leaks, so the reasoning is recorded on the type.

## Other changes

- Encoder and decoder concurrency pinned to 1. The default fills a pool of `GOMAXPROCS` encoders, each holding a 1 MiB long table plus a 256 KiB short table, while exactly one is ever used: roughly 20 MiB of dead allocation per export against a 512 MiB daemon reserve.
- Level `SpeedFastest`. With a 92% zero profile the ratio difference is negligible and the CPU on multi-GB memfiles is not.
- `New(endpoint, bucket, compress)` drops its variadic, so a future forgotten wiring is a compile error rather than a silent `compress=false`.

## Tests

- **`TestCompressedExportLeavesArtifactDirUntouched`** is the regression guard: the artifact directory listing, **including dotfiles**, is identical before and after an export, and again after the already-durable skip path. The leaked temp was a dotfile, so a test ignoring dotfiles would not have caught it.
- **`TestCompressedExportPutFailureLeavesDirClean`** covers the pipe error path: Export errors, no `meta.json` is PUT, listing unchanged.
- **`TestCompressedIncompressibleRoundTrip`** used a payload with period 256 that compressed roughly 194:1 and never exercised the case its name claims. It now uses `crypto/rand` and asserts the stored object exceeds the plaintext size.
- The other compressed tests (round trip, legacy empty-marker, truncated object) pass unchanged, which is the point: streaming must not alter observable behaviour.

Verified with the focused target under `--nocache_test_results --test_output=all --test_arg=-test.v`, so the tests provably executed:

```
Executed 1 out of 1 test: 1 test passes.
21 test functions passing, 0 failing
```

## Known unknown

SeaweedFS accepting a Content-Length-less PUT is not provable in CI (httptest accepts chunked trivially). It resolves at arming time. The failure mode is loud and safe: the export errors, the artifact stays local-durable, and the pre-agreed fallback is staging in a scratch dir outside every artifact root.

## Note on CI

Full `ci` currently fails on `main` for an unrelated reason: MODULE.bazel pins `gazelle` at `0.47.0` while `go.mod` requires `bazel-gazelle` `0.52.2`, so `@bazel_gazelle` cannot resolve. That is a separate one-line fix and is deliberately not folded in here.

https://claude.ai/code/session_01DTQ29PfFZShnxQUBAsW977

### DIFF
--- a/projects/embervm/noded/store/store.go
+++ b/projects/embervm/noded/store/store.go
@@ -120,19 +120,46 @@ type Store struct {
 // empty, which every caller reads as "the store is disabled" (export is skipped,
 // restore-on-miss is impossible, and the continuity verbs refuse). The endpoint
 // is normalised to have no trailing slash so key joins are unambiguous. The
-// optional compress argument enables zstd for newly exported objects; omitting
-// it preserves the legacy writer behavior.
-func New(endpoint, bucket string, compress ...bool) *Store {
+// compress enables zstd for newly exported objects.
+func New(endpoint, bucket string, compress bool) *Store {
 	if endpoint == "" {
 		return nil
 	}
-	compressEnabled := len(compress) > 0 && compress[0]
 	return &Store{
 		endpoint: strings.TrimRight(endpoint, "/"),
 		bucket:   bucket,
 		client:   &http.Client{},
-		compress: compressEnabled,
+		compress: compress,
 	}
+}
+
+// countingReader counts the bytes actually read out of a request body, so an
+// export can report compressed bytes moved without knowing the encoded size in
+// advance.
+//
+// Close is NOT optional and must not be dropped. net/http wraps a plain
+// io.Reader body in io.NopCloser, so a counting reader that only implements
+// Read would swallow the underlying Close. On the compressed path the body is
+// an *io.PipeReader, and the transport closing it is the ONLY thing that
+// unblocks the encoder goroutine when a request is aborted mid-body (a failed
+// PUT, a cancelled context). Without this method that goroutine blocks forever
+// on pw.Write, leaking itself and its encoder buffers on every failed export.
+type countingReader struct {
+	io.Reader
+	n int64
+}
+
+func (r *countingReader) Read(p []byte) (int, error) {
+	n, err := r.Reader.Read(p)
+	r.n += int64(n)
+	return n, err
+}
+
+func (r *countingReader) Close() error {
+	if c, ok := r.Reader.(io.Closer); ok {
+		return c.Close()
+	}
+	return nil
 }
 
 // url composes the object URL for a key under the bucket. The key is used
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

PR #4926, opened 2026-08-15
fix(agents): keep the composer above the iOS software keyboard

### LINKED ISSUE #4613: agents console: iOS keyboard occludes the composer on mobile
The mobile layout for `/agents` landed in #4603 and is live. On iOS Safari the
software keyboard occludes the message composer when it opens.

## Symptom

Focusing the composer raises the keyboard, which does not resize the visual
viewport the way it resizes the layout viewport. The composer stays where it
was laid out, behind the keyboard, so the user types into an input they cannot
see.

## Why the usual fix is not enough

`100vh` and `100dvh` both describe the layout viewport, so neither moves when
the keyboard opens. The affordance that does track it is `visualViewport`
(`height` and `offsetTop`, plus its `resize` and `scroll` events), which is the
only source that reports the occluded region.

Worth confirming before building: whether the composer needs to be pinned to
the visual viewport, or whether the scroll container simply needs its padding
adjusted so the composer scrolls into view. The second is much less invasive
and may be sufficient.

## Constraints

- The transcript must stay scrolled to its anchor when the keyboard opens and
  closes. Jumping to the top on focus is worse than the occlusion.
- Desktop and Android must not regress. `visualViewport` handling that assumes
  a keyboard will misbehave on a resized desktop window.
- The `max-width: 760px` full-view sidebar from #4603 is the layout this has
  to work inside.

jomcgi:
Closes #4613.

On iOS Safari, focusing the `/agents` composer raised the software keyboard over it, so you typed into an input you could not see.

## Why the cheap fix does not apply

The issue floated adjusting the scroll container's padding as the less invasive option. That cannot work here. The composer is not `position: fixed`: it is the last normal-flow child of `.transcript`, a flex column inside `.console { height: 100dvh }`. Padding on `.turns` moves content *inside* the scroll box, and the composer sits *below* that box. What the keyboard covers is the container's own bottom edge.

So `100dvh` is the bug. `dvh` resolves against the layout viewport, which iOS does not shrink when the keyboard opens, so the whole flex column keeps extending under it.

## What this does

`.console` height becomes `var(--console-h, 100dvh)`, and a new `visual-viewport.js` writes `--console-h` from `visualViewport.height`. The `100dvh` fallback stays, so SSR, no-JS, and browsers without `visualViewport` render exactly as before.

**Scroll anchor.** Shrinking a scroll container preserves `scrollTop`, not the bottom anchor, so a height change alone would leave the transcript parked one keyboard-height above the newest turn. Concretely: at the bottom the near-bottom distance is 0; the keyboard takes 336px of `clientHeight` without moving `scrollTop` or `scrollHeight`, so the distance measures 336 and fails the 200px threshold. The measurement therefore straddles the write: `measure()` runs before, `apply()` re-pins after. The near-bottom test is split out of `autoScroll` so the 200px threshold is still defined exactly once.

**Desktop and Android.** Gated on the existing `MOBILE_MEDIA_QUERY` constant rather than applied everywhere, because desktop pinch-zoom also shrinks `visualViewport.height` and binding the app height to it there would shrink the console for a zoomed desktop user. The gate re-evaluates on `MediaQueryList` `change`, and removes `--console-h` when it stops matching so the `100dvh` fallback resumes.

**Zoom.** `scrollTo(0, 0)` on a non-zero `offsetTop` runs on `resize` only, never on `scroll`. A non-zero `offsetTop` is also what panning a pinch-zoomed page looks like, and resetting there would fight the user on every pan frame. Page zoom is an accessibility affordance (WCAG 1.4.4).

## Verification

- `ci lint` clean.
- `//projects/monolith/frontend:test` with `--nocache_test_results`: `Executed 1 out of 1 test: 1 test passes`, including `visual-viewport.test.js (10 tests)`.
- `//projects/monolith/frontend:build_test` with `--nocache_test_results`: `Executed 1 out of 1 test: 1 test passes`. The vitest target does not compile the app, so this is the check that catches a Svelte template error.

## Not verified

The iOS behaviour itself has not been observed on hardware. The fix is derived from `visualViewport` semantics and covered by unit tests, but neither CI nor a headless browser exercises a real software keyboard. Worth an eyeball on a phone once it deploys.

### DIFF
--- a/projects/monolith/frontend/src/routes/private/agents/+page.svelte
+++ b/projects/monolith/frontend/src/routes/private/agents/+page.svelte
@@ -24,6 +24,7 @@
   import { firstLine, fmtCost, joinMeta } from "./run-format.js";
   import { partitionRuns, recentRuns, runActivityAt } from "./run-history.js";
   import { crumbTrail, sessionLineage } from "./lineage.js";
+  import { setupVisualViewport } from "./visual-viewport.js";
   import { RUN_LEXICON as P } from "./run-lexicon.js";
   import PaneHeader from "./PaneHeader.svelte";
   import SessionWalkthrough from "./SessionWalkthrough.svelte";
@@ -120,6 +121,7 @@
   let renderedPending = $state({});
   let vms = $state({});
   let turnsEl = $state(null);
+  let consoleEl = $state(null);
 
   const selectedSession = $derived(
     sessions.find((session) => String(session.id) === String(selectedId)) ??
@@ -366,11 +368,16 @@
     return index === 0 ? "starting" : "waiting";
   }
 
+  function nearBottom() {
+    if (!turnsEl) return false;
+    return (
+      turnsEl.scrollHeight - turnsEl.scrollTop - turnsEl.clientHeight < 200
+    );
+  }
+
   function autoScroll(force = false) {
     if (!turnsEl) return;
-    const nearBottom =
-      turnsEl.scrollHeight - turnsEl.scrollTop - turnsEl.clientHeight < 200;
-    if (force || nearBottom) turnsEl.scrollTop = turnsEl.scrollHeight;
+    if (force || nearBottom()) turnsEl.scrollTop = turnsEl.scrollHeight;
   }
 
   function syncPendingPartials(entries) {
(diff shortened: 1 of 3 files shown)


jomcgi:
## Helm Manifest Diff

PR #4927, opened 2026-08-15
fix(monolith): move the dev refresh out of container args and verify every phase

jomcgi:
Round four of this job failed because the wipe never ran:

```
ERROR:  syntax error at or near "$"
LINE 1: DO $
```

kubelet expands `$(VAR)` over `command`, `args` and `env` values at container start, and `$$` is its escape for a literal `$`. The inline PL/pgSQL block was mangled before the shell ever saw it. Then, because `psql` exits 0 even when its statements fail, the script carried on into the restore against a still-populated database: 664 "already exists" errors.

**`helm template` plus `sh -n` could not catch this.** The rendered manifest still contains `$$`; the mangling happens later, inside kubelet. Fable confirmed the diagnosis independently by porting kubelet's expansion algorithm and running it over the rendered manifest.

## Why restructure rather than patch again

| Round | Cause | Fix |
|---|---|---|
| 1 | dump from a hot standby, cancelled by WAL replay | dump from the primary |
| 2 | `DROP EXTENSION` ownership; barrier both mis-sequenced and unqualified | filter list, arm EXIT trap, qualify |
| 3 | `DROP CONSTRAINT` on inherited partition keys | drop `--clean`, wipe first |
| 4 | `$$` mangled by kubelet; wipe failed silently | this PR |

Four causes, one shape: **a step reports success without having done anything.** Patching instance five was the wrong move. Two changes make the class impossible.

### Transport

The script moves to `chart/scripts/dev-refresh.sh`, mounted from a ConfigMap via the same `.Files.Get` idiom as `migrations-configmap.yaml`. kubelet never expands file contents, so the bytes in git are the bytes the shell runs, and static checks on the file are meaningful. This retires the whole class, not the instance: any future `$(WORD)` in an inline script where WORD matches a defined env var would be substituted silently.

### Verification

Every phase asserts what it did, because neither exit codes nor rendered manifests can be trusted here:

- `ON_ERROR_STOP=1` and `--no-psqlrc` on every `psql` call
- wipe via `\gexec`: no dollar quoting to mangle, one connection, ON_ERROR_STOP covers every generated statement
- **ASSERT EMPTY before restoring.** An incomplete wipe is now loud, and loud *before* the restore. Nothing stood here in round four
- schemas restored are `diff`ed against the schemas the archive carries, which self-calibrates as prod grows rather than rotting like a hardcoded count
- chunk count compared against prod, read back from dev

### Safety

The barrier no longer depends on running at all: `dbos` and the two outbox tables' **data** are excluded from the restore list (the tables are still created, empty), so dev holds no actionable queue state at any instant.

This closes a hole nobody had hit yet: `activeDeadlineSeconds` SIGKILLs the pod, and **a fatal signal does not run an EXIT trap**, so the trap alone never guaranteed the barrier. Confirmed first that no foreign key from outside `dbos` points into `dbos` or either outbox, so the exclusion cannot break a referencing table.

The barrier is also now called explicitly in the main flow and then verified, with the trap kept as an idempotent backstop. A trap-only barrier would have run *after* the verification, not before it.

Plus: terminate dev app backends and `SET lock_timeout = '60s'` so the wipe cannot block on their locks until the hour deadline, which would read as a hang rather than a conflict.

## Verification done

- `sh -n` on the real artifact, re-run after `shfmt` reformatted it
- ConfigMap data key extracted from `helm template` and diffed against the git file: **byte-identical**, proving the Helm transport
- kubelet expansion emulator over the rendered CronWorkflow: **changes nothing**
- `defaultMode` renders as decimal 365, i.e. `0555` octal, not 555
- the exclusion grep and the schema-name parser exercised against realistic `pg_restore --list` input: `SCHEMA - public` excluded while `chat_public` survives, only the outbox data dropped
- the ASSERT EMPTY query run against currently-populated dev: returns 33 objects, so it would correctly refuse
- `ci test`: `Executed 2 out of 717 tests: 717 tests pass`

## Predicted round five, pre-empted

Fable predicted a `SCHEMA - public` TOC entry failing "already exists" once `--clean` is gone, `public` being app-owned here. Not present in this archive (checked round four's log), filtered anyway: harmless if absent, and the failure it prevents is another whole round.

## Still unproven

A green run. That is the only thing that closes this, and I will trigger one after rollout rather than claiming it here.

### DIFF
--- a/projects/monolith/chart/scripts/dev-refresh.sh
+++ b/projects/monolith/chart/scripts/dev-refresh.sh
@@ -1,0 +1,220 @@
+#!/bin/sh
+# Refresh the dev database from production.
+#
+# This lives in a file, mounted from a ConfigMap, rather than inline in the
+# CronWorkflow's container `args`. That is not tidiness, it is correctness.
+# kubelet runs its own `$(VAR)` expansion over `command`, `args` and `env`
+# values at container start, in which `$$` is the escape for a literal `$`. An
+# inline PL/pgSQL block therefore reached the shell as `DO $` and failed with a
+# syntax error, while the whole wipe was skipped:
+#
+#   ERROR:  syntax error at or near "$"
+#   LINE 1: DO $
+#
+# `helm template` plus `sh -n` could not catch that, because the rendered
+# manifest still contains `$$`; the mangling happens later, inside kubelet.
+# Mounted file contents are never expanded, so what is in git is byte for byte
+# what the shell executes, and static checks on this file are meaningful. It
+# also retires the rest of the class: any future `$(WORD)` in an inline script
+# where WORD matches a defined env var would be silently substituted.
+#
+# Every phase asserts what it did. Four consecutive nightly failures were each a
+# different cause with one shape in common: a step reported success without
+# having done anything. Exit codes lie here (psql exits 0 on failed statements
+# unless ON_ERROR_STOP is set) and so do rendered manifests, so the job trusts
+# neither and reads the database back instead.
+#
+# Deliberately no `set -x`: it would print PGPASSWORD into the Workflow logs.
+set -eu
+
+dump_file=/tmp/monolith.dump
+list_file=/tmp/monolith.list
+
+case "$DEV_PGHOST" in
+*monolith-dev*) ;;
+*)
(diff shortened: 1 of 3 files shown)


jomcgi:
## Helm Manifest Diff

PR #4928, opened 2026-08-15
feat(agents): follow the private tier's time-of-day palette in the console

### LINKED ISSUE #4683: The agents console does not follow the private tier's time-of-day palette
## What

`/agents` renders on its own light-only palette (`.console` in
`projects/monolith/frontend/src/routes/private/agents/agents-theme.css`).
The private landing page at `/private` renders on a clock-driven palette
(`.shell.dawn` / `.day` / `.dusk` / `.night` in
`src/lib/private/dashboard-theme.css`, period chosen at
`src/routes/private/+page.svelte:152-161`).

At night the landing page is dark and clicking through to the console
throws a white page at you.

## What is not true

A UX critique of this console asserted that `agents-theme.css` is
"light-only (`color-scheme: light`), while the rest of the private tier
has a real dark theme". Neither half holds, and it is worth writing down
so the next reader does not act on it:

- there is no `color-scheme` declaration in `agents-theme.css`;
- there is no `prefers-color-scheme` or `data-theme` anywhere under
  `projects/monolith/frontend/src`;
- the private tier has a time-of-day palette on one page, not a dark mode.

## Options

1. Give `.console` the same period classes as `.shell`, so the tier moves
   together. Costs: the console changes colour mid-run, so a state chip
   learned as one shade is another an hour later. That is a real hazard
   on a monitoring surface.
2. Add a `prefers-color-scheme: dark` mapping for the existing `.console`
   token names. Self-contained and stable during a run, but it makes the
   console the only surface in the app answering to the OS setting: a new
   convention rather than an existing one.
3. Leave it light.

## Status

Deferred deliberately. The copy migration and first-line title that were
bundled with this in the console UX work shipped separately; only the
palette question is open.

jomcgi:
Closes #4683.

At night the private landing page is dark and clicking through to `/agents` threw a white page at you. The console now carries the same clock-driven palette as the rest of the tier.

## Corrections to the issue body

Three things in the issue were wrong, and are worth recording since its "What is not true" section was itself partly wrong:

- `color-scheme: light` **did** exist, at `+page.svelte:1732`, in the component style block rather than `agents-theme.css`. The critique the issue rebuts named the wrong file but was right about the fact.
- `agents-theme.css:14` already claimed the neutral tokens "vary per period via `.shell.dawn`/`.shell.dusk`/`.shell.night` overrides". No such overrides existed and `.console` carried no period class.
- `lib/styles/shared/tokens.css:49` defines `body.dark`, applied by nothing anywhere in `src/`. `.impeccable.md:155-157` describes the private tier as having "a real `body.dark` theme" on the strength of it. It is dead CSS, so there was no existing dark-mode convention to join.

## Approach

Option 1 from the issue, scoped so the hazard it names is designed out rather than accepted.

**Hue is period-invariant.** Green stays green, amber amber, red red. Only lightness moves, and only as far as WCAG AA requires. The issue rightly flags "a state chip learned as one shade is another an hour later" as a real hazard on a monitoring surface; holding hue fixed means the channel carrying the meaning never moves.

**Only night goes dark.** `day` is unchanged and is the default. `dawn` and `dusk` shift the `--page-bg` surround only, so panels stay white and the console's stated "white working surfaces" direction is preserved. That confines the state-token work to one period.

**Soft fills derive from the surface** via `color-mix(in srgb, var(--ok) N%, var(--panel-bg))` rather than four hand-tuned sets, so they track the panel automatically. Body text on them lands between 7.06:1 and 9.78:1.

**The inverted pair.** `--ink` / `--ink-text` paint a near-black chip with white text for `.chip.on` and `.send-button`. At night `--ink` (#1c2024) would have sat on `--panel-bg` (#1c1e26) and dissolved into it, so the pair inverts rather than darkens: 13.20:1.

## Mechanism: an `<html>` attribute, not a class

`data-agents-period` is stamped by a blocking inline script in `app.html`, following the existing `data-agents-rail` precedent, and kept current after hydration by a 60s clock.

A Svelte class would only apply at hydration, so a night visitor would get a white flash first, which is the exact jolt this issue exists to remove. SSR cannot fix it either: the server clock is UTC while the period is a property of the viewer's local time.

Period boundaries are extracted from `routes/private/+page.svelte` into `lib/private/period.js` so the console and the landing page cannot drift. The landing page's `greeting` uses different boundaries and is deliberately left alone.

## Contrast

Every value computed with the standard WCAG formula, measured against `--panel-bg` (#1c1e26) unless noted:

| Token | Ratio |
|---|---|
| `--text` #e8e5dc | 13.20:1 |
| `--text-soft` #a5a297 | 6.50:1 |
| `--muted` #9a9890 | 5.75:1 (6.30:1 on `--page-bg`) |
| `--line-strong` #7a7870 | 3.76:1 (UI boundary, 3:1 floor) |
| `--dot-idle` #7a7870 | 4.12:1 on `--page-bg` |
| `--ok` #6fbf8d | 7.53:1 |
| `--attn` #d0a050 | 6.99:1 |
| `--attn-text` #f0e4c0 | 7.34:1 on `--attn-soft` |
| `--info` #7eb3f0 | 7.60:1 |
| `--err` #d9776a | 5.38:1 |

## Verification

- `ci lint` clean.
- `//projects/monolith/frontend:test` and `:build_test` with `--nocache_test_results`: `Executed 2 out of 2 tests: 2 tests pass`, including `period.test.js (10 tests)`.
- `color-scheme` lives in `agents-theme.css` rather than the component. An unscoped `html[data-*]` selector inside a Svelte `<style>` block is deleted at compile time with a `css_unused_selector` warning that `build_test` does not surface, so the rule would have shipped as a no-op behind green CI. Moving it to the plain stylesheet removes the whole class of bug; the 15 remaining `html[...]` rules in the component are all `:global()`-wrapped.

## Not verified

The rendered result has not been looked at on a real screen in any period. Contrast is computed, not observed, and "clears AA" is not the same as "reads well at 2am".

### DIFF
--- a/projects/monolith/frontend/src/app.html
+++ b/projects/monolith/frontend/src/app.html
@@ -18,6 +18,28 @@
         // localStorage blocked or quota exceeded; continue with expanded
       }
     </script>
+    <script>
+      /* Set the initial period attribute before first paint.
+         This runs inline and blocking to avoid a white flash when switching
+         to night mode. The period module (src/lib/private/period.js) is the
+         source of truth; this duplication must remain inline because app.html
+         has no module graph. The agents console updates this attribute after
+         hydration in response to clock ticks every 60s. */
+      try {
+        const hour = new Date().getHours();
+        const period =
+          hour >= 5 && hour < 11
+            ? "dawn"
+            : hour >= 11 && hour < 17
+              ? "day"
+              : hour >= 17 && hour < 22
+                ? "dusk"
+                : "night";
+        document.documentElement.setAttribute("data-agents-period", period);
+      } catch (e) {
+        // Gracefully degrade if Date is not available
+      }
+    </script>
     <!-- Declared in the static shell (not just the layout) so it is present at
          initial parse: the /app/* pages are ssr=false, so a layout-injected
          icon arrives after hydration and the browser has already 404'd on the
--- a/projects/monolith/frontend/src/lib/private/period.js
+++ b/projects/monolith/frontend/src/lib/private/period.js
@@ -1,0 +1,18 @@
+/**
+ * Period computation for the private tier's time-of-day palette.
+ * Period boundaries must remain in sync with the agents console and the
+ * landing page's `+page.svelte`. The landing page's `greeting` message uses
+ * different boundaries and is a separate concern.
+ */
+
+/**
+ * Compute the current period from an hour value.
+ * @param {number} hour - Hour value from 0-23
+ * @returns {"dawn" | "day" | "dusk" | "night"} The period for the hour
+ */
+export function periodForHour(hour) {
+  if (hour >= 5 && hour < 11) return "dawn";
+  if (hour >= 11 && hour < 17) return "day";
+  if (hour >= 17 && hour < 22) return "dusk";
+  return "night";
+}
--- a/projects/monolith/frontend/src/lib/private/period.test.js
+++ b/projects/monolith/frontend/src/lib/private/period.test.js
@@ -1,0 +1,44 @@
+import { describe, it, expect } from "vitest";
+import { periodForHour } from "./period.js";
+
+describe("periodForHour", () => {
+  it("returns 'night' for hour 0", () => {
+    expect(periodForHour(0)).toBe("night");
+  });
+
+  it("returns 'night' for hour 4", () => {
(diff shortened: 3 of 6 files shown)


jomcgi:
## Helm Manifest Diff

PR #4929, opened 2026-08-15
feat(agents): surface stream connection state in the VM chips

### LINKED ISSUE #4612: agents console: surface stream connection state in the VM chips
The VM state stream landed in #4610: `/agents/vms/stream` (SSE) replaced the
100ms poll, with a 2s poll fallback that arms on `EventSource.CLOSED` or three
consecutive errors.

Nothing in the UI says which of the three modes is live.

## The gap

Three states are visually identical today:

1. Stream connected, frames arriving. VM chips are within ~1s of truth.
2. Stream dead, fallback polling at 2s. Chips are correct but staler, and the
   page is back to per-viewer control plane load.
3. Both dead, or the refresher published an error. Chips hold their last
   values and simply stop changing.

State 3 is the dangerous one: a frozen chip and a genuinely idle fleet render
the same. The console has no "nothing has changed" versus "nothing is being
reported" distinction.

This is not a regression from the polling version, which had the same blind
spot. The stream makes it easier to be silently wrong for longer, because a
poll loop failing is self-correcting on the next tick and a closed EventSource
is not.

## What exists to build on

The backend already folds the outage into the payload rather than dropping it:
`/vms` and each SSE frame carry an `error` field when the control plane refresh
fails, and `_publish_vm_map` treats an error change as a change, so an outage
and a recovery both wake subscribers. The client currently ignores the field.

## Scope

- Render the `error` field when present, rather than discarding it.
- Distinguish fallback-polling from streaming, at whatever weight suits the
  page. This should be quiet: it is an availability hint, not an alert.
- Give a frozen state a visible age, so "last updated 4 minutes ago" is
  readable when the stream is gone.

## Not this issue

Reconnect strategy changes, and any backoff tuning on the fallback. The
current escalation works; this issue is only about saying which mode is live.

jomcgi:
Closes #4612.

Three states rendered identically in the console: stream connected, stream dead with the 2s fallback polling, and both dead. The third is the dangerous one, because a frozen chip and a genuinely idle fleet looked the same.

## Why liveness cannot come from message timing

Two properties of the existing stream rule out the obvious approach, and both are load-bearing:

- `_publish_vm_map` wakes subscribers **only when the map or the error actually changed**. A healthy idle fleet legitimately sends zero frames for an unbounded time, so "no message in N seconds means stalled" would fire constantly on a working console.
- The 15s heartbeat is `": ping\n\n"`, an SSE **comment**. It keeps the connection warm at the transport layer and never fires `onmessage`. `_run_vm_refresher` documents this exact trap: the connection looks healthy, no message arrives, and the fallback never engages.

So connection state is derived from `EventSource.readyState` and whether the fallback is armed, never from timing. A test asserts that a long idle period while streaming does not change the mode; that is the assertion that stops a time-based heuristic being reintroduced later.

## What this adds

A pure reducer in `vm-stream-status.js` (`nextStatus(prev, event)` plus `formatAge`), with the component holding the state and feeding it events from the stream handlers. Four modes: `connecting`, `streaming`, `polling`, `stalled`.

The `error` field that both `/vms` and every SSE frame already carried is now read instead of discarded, and `source.onopen` is wired as the positive signal for `streaming`.

Rendering is quiet, as the issue requires: an availability hint, not an alert. While streaming, only a muted dot renders. The mode label and the "updated 4m ago" age appear only when `polling` or `stalled`, because showing an age during healthy idle operation would read as a problem and reproduce the very confusion this issue is about. An `sr-only` span carries the full state for screen readers in all four modes.

## The map is applied unconditionally, on purpose

An earlier revision of this change gated `vms = body.vms ?? {}` on the connection being healthy. That inverts a deliberate backend guarantee. On a control plane failure `router.py` publishes an **empty** map alongside the error, and says why:

> Drop the map rather than serving the last known one. A stale entry renders as a confident "awake" chip for a guest that may be long gone, where an absent entry renders "off" alongside the error. Losing the CP means we do not know, and the honest rendering of not knowing is the one the console already had.

Gating the assignment would have kept painting confident "awake" chips for guests that may be gone, during exactly the outage this change exists to surface. Both assignment sites carry a comment saying so. The status indicator decides what to say; it never decides whether the map is applied.

## Not this issue

Reconnect strategy and fallback backoff are untouched, per the issue.

## Verification

- `ci lint` clean.
- `//projects/monolith/frontend:test` and `:build_test` with `--nocache_test_results`: `Executed 2 out of 2 tests: 2 tests pass`, including `vm-stream-status.test.js (21 tests)`.

## Not verified

The degraded modes have not been exercised against a real control plane outage. The reducer is covered by unit tests, but nothing here proves the console renders sensibly during an actual EmberVM failure.

### DIFF
--- a/projects/monolith/frontend/src/routes/private/agents/+page.svelte
+++ b/projects/monolith/frontend/src/routes/private/agents/+page.svelte
@@ -25,6 +25,7 @@
   import { partitionRuns, recentRuns, runActivityAt } from "./run-history.js";
   import { crumbTrail, sessionLineage } from "./lineage.js";
   import { setupVisualViewport } from "./visual-viewport.js";
+  import { formatAge, nextStatus } from "./vm-stream-status.js";
   import { RUN_LEXICON as P } from "./run-lexicon.js";
   import PaneHeader from "./PaneHeader.svelte";
   import SessionWalkthrough from "./SessionWalkthrough.svelte";
@@ -140,6 +141,13 @@
   let requestSequence = 0;
   let renderedPending = $state({});
   let vms = $state({});
+  let vmStreamStatus = $state({
+    mode: "stalled",
+    lastUpdateAt: null,
+    error: null,
+  });
+  let vmStreamNow = $state(Date.now());
+  let vmFallbackArmed = false;
   let turnsEl = $state(null);
   let consoleEl = $state(null);
 
@@ -625,6 +633,11 @@
     });
   }
 
+  function updateVmStreamStatus(event) {
+    vmStreamStatus = nextStatus($state.snapshot(vmStreamStatus), event);
+    return vmStreamStatus;
+  }
+
   function selectRun(runOrId) {
     const id = typeof runOrId === "object" ? runOrId?.workflow_id : runOrId;
     if (id == null) return;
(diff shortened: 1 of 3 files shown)


jomcgi:
## Helm Manifest Diff

PR #4930, opened 2026-08-15
feat(agents): explain a degraded workspace restore in the transcript

### LINKED ISSUE #4329: agent sessions: surface workspace-not-recovered as first-class state, not a log line
When a restore degrades (lineage_live_heir, unknown_lineage, export race) the transport falls back to a blank session with only logger.warning (transport.py:343-353, :432-445) and the turn succeeds: the user meets an agent with amnesia and no explanation. Persist restored/degraded on the turn (usage_json or a column), return it from /api/agents session detail, and render a clear banner in the /agents UI. Also: the UI statusClass tests for error/failed/failure, strings the backend never emits (real statuses: running/completed/warn/needs_input).

jomcgi:
Closes #4329.

When a restore degrades, the transport falls back to a blank session and the turn still succeeds, so the user meets an agent with amnesia and no explanation.

## Most of this issue already shipped

The issue body is stale. Scoping it found:

- **Persistence: already done.** `store.py:547-548` writes `usage["workspace_recovery"]` into `usage_json`, assembled in `transport.py` and covered by four tests at `transport_test.py:804-879`.
- **API exposure: already done.** `router.py:466` returns the whole `usage` blob, so `workspace_recovery` already reached the client.
- **The dead `statusClass` branches: already fixed.** `status.js` no longer tests for `error` / `failed` / `failure`.

So this is frontend only. No migration, no schema change, no Python. The column-versus-`usage_json` question in the issue is moot for the banner; the only thing a column would add is queryability, which is filed separately.

## Placement and register

Rendered inline per turn inside the existing turn loop, not as a session-level banner. A degraded restore is a property of the specific turn whose workspace was lost, so the explanation belongs at the point in the transcript where the amnesia starts.

Attention register, not error. The turn succeeded: `transport.py` deliberately falls back to a blank session rather than failing it, so styling it like `.turn-error` would misreport a working degradation as a failure.

## Contrast

The banner uses `color: var(--attn-text)`, not `var(--attn)`. Measured against `--attn-soft`:

| | `--attn` | `--attn-text` |
|---|---|---|
| day | 4.09:1 | 6.01:1 |
| night | 3.91:1 | 7.34:1 |

`--attn` fails the 4.5:1 AA floor for body text in both periods. `agents-theme.css` documents this exact pairing as the reason `--attn-text` exists, and `.session-state.warn` already uses the correct combination. The border keeps `--attn`, where the 3:1 UI-boundary floor applies.

## Copy

One visible sentence covers all three degrades, because the user consequence is identical in each:

> The agent started with an empty workspace and cannot see files or changes from earlier turns.

The differing cause is carried in the `title`, in operator register, since this console is a repo-facing surface. Three distinct strings naming `restore_denied`, `restore_fallback`, and the unrecognised case. An unknown degrade renders the generic message rather than nothing, so a new backend state cannot silently reintroduce the invisibility this issue is about.

Joe writes final wording; treat the sentence above as a draft.

## Verification

- `ci lint` clean.
- `//projects/monolith/frontend:test` and `:build_test` with `--nocache_test_results`: `Executed 2 out of 2 tests: 2 tests pass`, including `workspace-recovery.test.js (14 tests)`.
- Separately verified that every lexicon reference the template makes actually resolves, and that the three titles are distinct. `build_test` cannot catch this: an undefined property lookup is a runtime miss, not a template syntax error, so a missing lexicon key compiles green and renders `undefined`.

## Not verified

The banner has not been seen against real degraded data. No session in the current database is known to carry `workspace_recovery.degraded`, so this is verified by unit test and by reasoning about the payload shape, not by observation.

### DIFF
--- a/projects/monolith/frontend/src/routes/private/agents/+page.svelte
+++ b/projects/monolith/frontend/src/routes/private/agents/+page.svelte
@@ -27,6 +27,10 @@
   import { setupVisualViewport } from "./visual-viewport.js";
   import { formatAge, nextStatus } from "./vm-stream-status.js";
   import { RUN_LEXICON as P } from "./run-lexicon.js";
+  import {
+    workspaceRecoveryMessage,
+    workspaceRecoveryTitle,
+  } from "./workspace-recovery.js";
   import PaneHeader from "./PaneHeader.svelte";
   import SessionWalkthrough from "./SessionWalkthrough.svelte";
   import { periodForHour } from "$lib/private/period.js";
@@ -1490,6 +1494,15 @@
                   >{/if}
               </div>
             </article>
+            {@const degradedCause = workspaceRecoveryMessage(turn)}
+            {#if degradedCause}
+              <div
+                class="turn-degraded"
+                title={P.workspaceRecoveryCauses[degradedCause]}
+              >
+                {P.labels.workspaceRecovery}
+              </div>
+            {/if}
           {/each}
 
           {#each detail?.pending_queue ?? [] as entry, index (entry.seq)}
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

PR #4931, opened 2026-08-15
fix(tools): stop shipping aspect-gazelle as the PATH BUILD generator

jomcgi:
## What

The `gazelle` binary in the developer tools image came from the multitool
lockfile, which pins [aspect-build/aspect-gazelle](https://github.com/aspect-build/aspect-gazelle)
2025.40.99. That is a different program from this repo's `//:gazelle_binary`,
which is bazel-gazelle plus the helm, semgrep, bzl, go, proto and python
extensions.

So anything calling bare `gazelle` regenerated BUILD files by aspect-gazelle's
own defaults. Among other things it rewrites external Go deps to
`:go_default_library` targets that do not exist, which the root `BUILD` already
documents as failing analysis rather than anything obvious.

Two callers did exactly that:

- `ci regen`, guarded only by `command -v gazelle`
- `bazel/tools/format/fast-format.sh`, where the call was `gazelle 2>/dev/null || true`, so the damage was silent and landed in whichever directory you were already editing

`//bazel/tools/BUILD` has excluded this binary from `bazel_env` for the same
reason since before this image existed:

```starlark
# Multitool tools (excluding gazelle - use `bazel run //:gazelle` instead)
```

ADR tooling/001 moved tool distribution from `bazel_env` to the OCI image, and
that exclusion did not come across.

## Why not vendor the correct binary instead

`//:gazelle_binary` links a cgo tree-sitter through
`@rules_python_gazelle_plugin//python`, and MODULE.bazel registers no cross C
toolchain (no `hermetic_cc_toolchain`, no `apple_support`). The linux/amd64
executors that build this image therefore cannot produce a darwin/arm64 copy,
or a linux/arm64 one. Every other non-Linux entry in the image is a downloaded
prebuilt, never a built one.

BUILD generation stays where it already worked: CI's Format stage, which runs
`bazel run //bazel/tools/format:format`, whose multirun includes `//:gazelle`,
and auto-commits on PR branches.

## Consequence

A PR that adds Go imports or Python deps now receives its BUILD update as a
`style: auto-format` bot commit rather than in its own commit. That is the
known non-fast-forward strand: fetch and rebase before the next push.

Developer machines keep the old binary in `~/.cache/homelab-tools` until this
merges and `./bootstrap.sh` runs again (it `rm -rf`s the tools dir on a digest
change). Nothing in the repo calls it in the meantime.

## Also

- Drops `ORION_EXTENSIONS_DIR` from `.envrc`. It pointed at `.aspect/gazelle/`, a directory this repo does not have, and aspect-gazelle logged `No orion plugins found` on every run.
- Drops the `SKIP_GAZELLE` gate, which nothing ever set, and the `BUILD_FILES` array in `fast-format.sh`, which existed only to decide whether to run gazelle.

Not touched: `env = {"ENABLE_LANGUAGES": ...}` on the `gazelle` rule in the root
`BUILD`. Very likely aspect-gazelle residue that bazel-gazelle ignores, but that
is unverified, so it is a separate cleanup rather than a guess in this PR.

## Test

`ci test`: `Executed 4 out of 717 tests: 717 tests pass`. The point of the full
run is analysis, which has to keep resolving with the three
`multitool.gazelle.*` repos removed from `MODULE.bazel`.

Verified `ci regen` from the branch copy prints the new line and runs no
gazelle. Worth noting for reviewers: typing bare `ci` does not test this, since
`.envrc` puts the **main checkout's** `bazel/tools/ci` on `$PATH` from any
worktree.

https://claude.ai/code/session_01DTQ29PfFZShnxQUBAsW977


jomcgi:
## Helm Manifest Diff

PR #4932, opened 2026-08-15
test(embervm): harden noded failure coverage

jomcgi:
Summary:
- make noded status and activator assertions fail when expected entries are absent
- cover partial Firecracker launch and restore cleanup, stalled store and guest calls, and probe failures
- strengthen header filtering and volume idempotency checks

Validation:
- gofumpt v0.9.2 applied to all changed files
- git diff --check passed
- bazel/tools/ci/ci test -- //projects/embervm/noded/... attempted, but no tests ran because BuildBuddy rejected the unauthenticated remote request with API key not set

This PR is draft because the required Linux Bazel test gate has not yet been verified.

### DIFF
--- a/projects/embervm/noded/egress/forwarder_test.go
+++ b/projects/embervm/noded/egress/forwarder_test.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"net"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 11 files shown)


jomcgi:
Closing unmerged. This came out of a session started to exercise the walkthrough feature rather than to land a test refactor, and its changes were never verified: the guest had no BuildBuddy API key, so `ci test //projects/embervm/noded/...` never ran and none of the 11 files' test changes are known to pass.

The session did its job. It surfaced three real defects in the session walkthrough, fixed in #4933, and its own output was honest about being unverified.

If the noded test gaps are worth pursuing, they should be scoped as their own issue rather than carried on an unverified branch.

PR #4933, opened 2026-08-15
fix(agents): stop the walkthrough contradicting an unavailable diff

jomcgi:
Three defects in the session walkthrough (ADR agents/056), found while using it on a real session rather than by reading the code.

## What was wrong

**A compare against the base branch is not a compare.** `compare_stats` resolved rung 2 whenever `session.branch` named an existing branch. A session started from `main` has `session.branch` `"main"`, so it computed `main...main`, found zero files, and every path the agent named became a contradiction.

Live evidence, session 190 turn 1:

```
resolution_rung: 2   diff_type: branch_ephemeral
base_sha: null   commit_sha: null   branch: "main"
files: []   total_files: 0   contradicted_paths: 5
```

The console rendered `0 points, 0 files changed` above five rows, each reading "Contradicted path. The agent named this file; the compare does not contain it." The agent had really changed 11 files (+335/-34) on its own branch. Session 191 reproduced it independently on an unrelated task.

**The header count ignored contradictions.** The list is `[...points, ...contradictions]` but the header counted only points and files, which is how `0 points` appeared above five rows.

**The compare routes were served at a doubled prefix.** `compare_router` declared `prefix="/api/swarm/compare"` and was included into a router already prefixed `/api/swarm`, so live OpenAPI served:

```
/api/swarm/api/swarm/compare/{session_id}/{turn_seq}
/api/swarm/api/swarm/compare/{session_id}/{turn_seq}/patch
```

`walkthrough.js` builds patch URLs at the single-prefix path, which 404s in production, so every "view patch" link was dead. `walkthrough.test.js` asserted the same broken path, which is why CI was green on it.

## What changed

- A base branch compared with itself resolves rung 3 and leaves the diff unresolved, so the composer renders the agent's account as testimony instead of contradicting it. `compare_patch` gets the same guard.
- The header count describes the list actually rendered: accounted, unexplained, contradicted, touched.
- Router prefix is `/compare`, composing to `/api/swarm/compare/...`, and the test pins the composed path.
- The panel leads with a factual summary above the existing walkthrough. The composer emits counts and provenance only; every visible string comes from `run-lexicon.js`.

The two channels stay unmerged: the points are the agent's testimony, the diff is the artifact. The composer never narrates the change itself.

## Verified

- `ci lint` clean.
- `ci test //projects/monolith/frontend:test //projects/monolith/frontend:build_test //projects/monolith:swarm_compare_router_test //projects/monolith:swarm_walkthrough_composer_test` -> `Executed 3 out of 4 tests: 4 tests pass`.
- `build_test` included deliberately: vitest does not compile the app, so a Svelte template error would otherwise pass green.
- All 37 `P.labels` / `P.punct` keys used by `SessionWalkthrough.svelte` resolve against the lexicon, checked mechanically. A missing key renders `undefined` with `build_test` still green.
- Server to client summary field mapping checked by hand (`files_changed` to `files`, `accounted_files` to `accounted`, `unexplained_files` to `unexplained`).

## Not verified

The rendered result has not been seen against a live session in production. The reasoning is grounded in the payloads above, but nobody has looked at the fixed panel in a browser.

## Not fixed here

This stops the false accusation. It does not populate the diff. The session lane never records the branch or SHAs an agent pushed, so rung 1 stays unreachable for session-lane walkthroughs and the diff half remains empty. That needs new plumbing (the guest reporting its branch, or the monolith discovering it) and is filed separately.

### DIFF
--- a/projects/monolith/frontend/src/routes/private/agents/SessionWalkthrough.svelte
+++ b/projects/monolith/frontend/src/routes/private/agents/SessionWalkthrough.svelte
@@ -55,6 +55,28 @@
   );
   const point = $derived(items[selected] ?? null);
 
+  function countLine(currentWalk) {
+    const parts = [
+      `${currentWalk.counts.accounted} ${P.labels.walkAccountedCount}`,
+      `${currentWalk.counts.unexplained} ${P.labels.walkUnexplainedCount}`,
+      `${currentWalk.counts.contradicted} ${P.labels.walkContradictedCount}`,
+    ];
+    if (currentWalk.counts.touched > 0) {
+      parts.push(`${currentWalk.counts.touched} ${P.labels.walkTouchedCount}`);
+    }
+    return joinMeta(...parts);
+  }
+
+  function headerCountLine(currentWalk) {
+    const parts = [countLine(currentWalk)];
+    if (currentWalk.rung <= 2) {
+      parts.push(
+        `${currentWalk.counts.files} ${currentWalk.counts.files === 1 ? P.labels.walkFileChanged : P.labels.walkFilesChanged}`,
+      );
+    }
+    return joinMeta(...parts);
+  }
+
   function basename(path) {
     return String(path || "")
       .split("/")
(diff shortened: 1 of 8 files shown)


jomcgi:
## Helm Manifest Diff

PR #4935, opened 2026-08-15
perf(embervm): punch holes on the store restore path and arm compression

jomcgi:
## What

Two changes to noded, both about snapshot disk and store bytes.

1. Hole punching moves from package `driver` into a new `noded/sparse` package, and `store.restoreFile` punches hydrated files.
2. `noded.store.compress` is armed.

## Why

Firecracker writes a guest's whole RAM image on snapshot, fully allocated, so a workload declaring `memMib: 4096` costs 4 GiB of snapshot disk however little the guest touched. `digHoles` already reclaims that, and it works. Measured on a live brick that built its own base:

| memfile | declared | allocated | freed |
|---|---|---|---|
| `sandbox-session` | 2048 MiB | 288 MiB | **85%** |

But measured across the live fleet, five of six bases on every brick were still at 0% freed:

| memfile | declared | allocated | freed |
|---|---|---|---|
| `claude-runtime` | 4096 MiB | 4096 MiB | 0% |
| `semgrep` | 1536 MiB | 1536 MiB | 0% |
| `bazel-query` | 1024 MiB | 1024 MiB | 0% |
| `sandbox` | 512 MiB | 512 MiB | 0% |
| `demo-postgres` | 150 MiB | 150 MiB | 0% |

`sandbox-session` was built locally on that brick. The other five arrived by hydration, and `store.restoreFile` copies the object into an `O_TRUNC` file, writing every zero back out densely. None of the seven `digHoles` call sites covered the restore path, so the punch only ever applied to the path that central artifacts are designed to replace. `claude-runtime` samples 92% zero, so that is roughly 3.7 GiB per brick left on the floor.

## How

`store` cannot import `fcvm/driver`, so the implementation moves to `noded/sparse` with `DigHoles` and `BestEffort` exported. The scan loop is unchanged. The seven driver call sites become `sparse.BestEffort` with the same path and kind arguments.

`restoreFile` punches the temp file after its size and sha256 checks pass and before the rename that publishes it, matching the pattern the driver already uses. `FALLOC_FL_PUNCH_HOLE` stays paired with `FALLOC_FL_KEEP_SIZE`, so logical size and contents-as-read are unchanged and only block allocation drops. Punching is best effort: a failure logs and does not fail the restore.

## On arming compression

The writer shipped inert on purpose, because a daemon that cannot read the marker cannot hydrate a compressed object mid-rollout. Every brick now runs a daemon that honors it, so the stated arming precondition is met.

The two changes are complementary rather than redundant. Punching is local-only: holes read back as zeros, and `Export` uploads the full logical extent, so a punched 4 GiB memfile still uploads as 4 GiB. Compression is what turns those same zero regions into a small object in the store.

Existing objects are never rewritten, by design: the idempotency fence compares sha256 and size, both properties of the uncompressed content, and `alreadyDurable` short-circuits before `Export` is called. That is fine without a backfill. `stateful` carries a 0.3 day TTL and `session-workspace` 7 days, together 93% of the bucket, so they migrate to compressed within one retention window. `base` is keyed by guest image digest, so it migrates on the next image bump.

## Testing

`ci lint` clean, `ci regen` no drift, `ci test` green:

```
Executed 62 out of 718 tests: 718 tests pass.
```

Both new targets re-run uncached to prove they execute:

```
bazel test //projects/embervm/noded/sparse:sparse_test \
           //projects/embervm/noded/store:store_test --nocache_test_results
Executed 2 out of 2 tests: 2 tests pass.
```

`TestRestorePunchesHoles` writes an 8 MiB payload with non-zero bytes only at the first and last offset, exports, restores, then asserts the contents are byte-identical and that allocation is under half the logical size. Removing the punch call fails the second assertion.

## Known gap

The new test exercises the uncompressed restore path only, so the compressed-object restore plus punch combination is not covered together. The punch sits after `io.Copy` regardless of whether a zstd decoder was in the chain, so I do not believe it can diverge, but it is the one interaction this change introduces and does not test.


jomcgi:
## Helm Manifest Diff

PR #4936, opened 2026-08-15
feat(agents): capture each turn's diff in the guest and render from it

### LINKED ISSUE #4934: monolith: capture the turn diff at completion instead of deriving it from GitHub at render time
The session walkthrough derives its diff from the GitHub compare API at render time. The diff is a fact about a past event, but it is being read live from a mutable source, so it decays, truncates, and is unavailable exactly when the work was never pushed.

PR #4933 stopped the walkthrough from misreporting a missing diff as agent dishonesty. This issue is about making the diff exist.

## Why the current source is wrong

All of this is visible in `projects/monolith/swarm/compare_router.py`:

- **It expires by construction.** The payload carries `ephemeral: compare_rung == 2` and the console says "This walkthrough becomes unavailable once this branch is deleted". The record of what an agent did should outlive the branch it did it on.
- **It truncates at 300 files.** There is a dedicated `truncation` step type to report "GitHub files truncated".
- **The expensive path is deliberately uncached.** `_compare` guards both `_cache_get` and `_cache_put` with `if not include_patches`, so per-file patch fetches always hit GitHub with a 15s timeout. That is the path a user triggers repeatedly by clicking through points.
- **Availability is coupled to GitHub.** Any non-200 becomes `HTTPException(502, "GitHub compare unavailable")`.
- **It cannot see unpushed work at all.** Session 191 committed `94a5b6b97` inside its guest and never pushed. There was nothing for GitHub to compare, so no walkthrough was possible even in principle.

## Decision

Capture the diff at turn completion and store it compressed alongside the turn, in Postgres.

The guest is the natural capture point: it holds the diff at the moment it makes it, before anything can be deleted or rewritten.

```
TURN COMPLETION (in guest)
  git diff <base>...<head>  ->  zstd  ->  typically 2-20 KB
            |
            v  recorded with the turn
  agent_turn.diff_blob      bytea
  agent_turn.diff_truncated bool

RENDER
  one row read, no network
```

Properties this buys: survives branch deletion, works when nothing was pushed, no 300-file ceiling, no GitHub dependency at render time, and the walkthrough becomes atomic with the turn row rather than a distributed read.

## Open work

- Agree the completion-payload field for the compressed diff. This is a cross-service contract between the guest and the monolith, so it needs writing down.
- Schema migration for the new columns, plus the size cap and what `diff_truncated` means when it trips. An uncapped blob column has an unbounded tail: a turn that regenerates a lockfile can be megabytes.
- Read path in `compare_stats` and `compare_patch`, preferring the stored diff and falling back to the GitHub path for turns predating capture.
- Retention: how long a stored diff lives, and whether it is deleted with the session.

## Supersedes

The original version of this issue asked how to record the agent's branch and SHAs so the compare could find them. Capturing the diff removes the need to know the branch at all, so that framing is retired. The evidence that prompted it is kept below.

Session 190 turn 1 carried the fields but never the values, because `update_turn_shas` is called only from `swarm/workflows.py` (the DBOS run lane) and never from the session lane:

```
base_sha: null   commit_sha: null
usage keys: activities, cache_read_tokens, cache_write_tokens,
            input_tokens, output_tokens, workspace_recovery
```

Meanwhile the agent had pushed `codex/noded-test-audit` at `dcdde14c9` with 11 files changed.

jomcgi:
Closes #4934.

## Why

The session walkthrough derived its diff from the GitHub compare API at render time. That has never once worked in production. Scanning every session:

```
sessions=191   with repo=68   swarm-lane=7
rung distribution: {3: 6, 5: 97}
```

Rung 1 and rung 2 have never occurred, so no walkthrough has ever shown a diff. The reasons are structural, not incidental:

- The compare expires with the branch (`ephemeral: rung == 2`, and the console said so out loud).
- GitHub truncates compare at 300 files.
- Work that was never pushed is invisible. Session 191 committed `94a5b6b97` inside its guest and never pushed, so no compare was possible even in principle.

## What changed

The guest records the checkout HEAD when a turn starts, then captures `git diff <base>` at the end. Diffing the working tree against the recorded base picks up committed and uncommitted work in one command, which is what "what this turn changed" means. The result is zlib compressed and stored beside the turn.

**Compression is zlib, not zstd.** The guest shim is deliberately stdlib-only Python (`runtimes/claude/apko.yaml` says so) and `zstandard` is absent from the monolith requirements. A dependency in the guest runtime would have cost more than the compression ratio is worth.

A stored diff resolves as `resolution_rung` 1 with `diff_type: "stored"` and `ephemeral` false, because it is authoritative and does not expire. `compare_stats` and `compare_patch` both prefer it and fall back to the GitHub path for turns predating capture. Parsing is a tested pure function, not a shell out to git.

**Capture cannot fail a turn.** A missing checkout or a git error yields a turn with no diff, and a guest predating this change is unaffected: an absent or malformed `diff` payload behaves exactly as before.

The two nulls are kept distinct deliberately. `diff_blob` null with `diff_truncated` true means the diff was too large to keep; null without it means nothing was captured. This feature has already been bitten once by reporting a missing artifact as a positive claim, and collapsing those two would do it again.

## Review found a real defect in the parser

The first implementation parsed `---` / `+++` header lines regardless of whether it was inside a hunk. Inside a hunk, content impersonates a header: removing a line reading `-- comment` produces `--- comment`, and adding one reading `++ x` produces `+++ x`. Since those headers decide the reported path, this misattributed files. Demonstrated against the real parser:

```
DELETED SQL   -> path: 'Store the server-owned intent separately.'   (should be the .sql path)
ADDS ++ LINE  -> path: 'some/other/file.py'                          (should be docs/example.md)
```

The second is the dangerous one: a real hunk served under a completely different file's name. Every migration in this repo opens with `-- `, and `run-fixtures.js` embeds diff text, so both shapes are reachable.

Fixed by only reading headers before the first hunk, with two regression tests. The pre-existing test used `----` and `++++` (four characters), which do not match `"--- "`, so it passed while the bug was live.

## Verified

- `ci lint` clean.
- `ci test` across `swarm_compare_router_test`, `swarm_unified_diff_test`, `agent_sessions_transport_test`, `embervm shim_test`, `frontend:build_test` -> `Executed 2 out of 5 tests: 5 tests pass`.
- Re-run after the parser fix -> `Executed 2 out of 2 tests: 2 tests pass`.
- Acceptance checked by hand, not from the implementer's report: no zstd anywhere, shim imports only `zlib`/`base64`, exactly one migration, all three columns on `AgentTurn` and persisted by `create_turn`, `Turn.diff` defaults to None.
- The rung-1 test monkeypatches both `_github_get` and `_compare` to raise, so any GitHub call fails the test. The capture-failure test asserts the turn still returns `{"ok": True}` with git exiting 1.

## Not verified

The end-to-end path has not run against a live session, because until this lands there is nothing that can produce a stored diff. The first real session after rollout is the actual test: parser correctness is proven, but the guest capturing the right base, the blob round-tripping through Postgres, and the patch endpoint serving the right file's hunk are not.

## Not in scope

Retention and deletion of stored diffs. No backfill, because no turn has ever had a diff to backfill.

### DIFF
--- a/projects/embervm/runtimes/claude/shim.py
+++ b/projects/embervm/runtimes/claude/shim.py
@@ -53,8 +53,7 @@ def _capture_turn_base(checkout_dir):
         if not os.path.exists(os.path.join(checkout_dir, ".git")):
             return None
         result = subprocess.run(
-            ["git", "rev-parse", "HEAD"],
-            cwd=checkout_dir,
+            ["git", "-C", checkout_dir, "rev-parse", "HEAD"],
             stdout=subprocess.PIPE,
             stderr=subprocess.DEVNULL,
             check=False,
@@ -76,8 +75,7 @@ def _capture_turn_diff(checkout_dir, base_sha):
         return None
     try:
         result = subprocess.run(
-            ["git", "diff", base_sha],
-            cwd=checkout_dir,
+            ["git", "-C", checkout_dir, "diff", base_sha],
             stdout=subprocess.PIPE,
             stderr=subprocess.DEVNULL,
             check=False,
--- a/projects/embervm/runtimes/claude/shim.py
+++ b/projects/embervm/runtimes/claude/shim.py
@@ -1,9 +1,9 @@
 #!/usr/bin/python3
 """HTTP over vsock shim for a long-lived Claude Code CLI session."""
 
-import http.server
 import base64
 import collections
+import http.server
 import json
 import math
 import os
@@ -17,6 +17,7 @@ import sys
 import threading
 import time
 import urllib.request
+import zlib
 
 
 GUEST_HTTP_PORT = 1027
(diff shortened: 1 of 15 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4937, opened 2026-08-15
embervm: a failed claude-runtime base build is invisible, because Ready stays True on the stale base


jomcgi:
## What happened

PR #4936 changed the guest shim (`projects/embervm/runtimes/claude/shim.py`) to capture each turn's diff. It merged, the chart published, embervm promoted to `0.14.1`, and ArgoCD reported Synced and Healthy. The pinned runtime image genuinely contains the new code:

```
crane export ghcr.io/.../runtimes/claude@sha256:c8a38ea... | grep -c
  _capture_turn_diff    2
  _capture_turn_base    2
  MAX_TURN_DIFF_BYTES   2
```

But session 192, started after the rollout, captured nothing. `diff_blob` null, `diff_base_sha` empty, for that turn and every other.

The reason is in the Workload status:

```
Ready:     True   BaseReady    "a restorable base snapshot is available"   17:29:17Z
BaseBuilt: False  BuildFailed  "{:connect, :timeout}"                      17:29:17Z
snapshotRef: claude-runtime__e847f766b497
```

The base rebuild failed with a connect timeout, most likely because the 16gi bricks were mid-roll from the chart bump at 17:27. Sessions kept working, restoring the previous base, which was built from the previous image. So the new shim shipped everywhere except into the thing that actually runs it.

Fifteen minutes later the condition had not changed, so it does not appear to retry promptly.

## Why this is worth fixing

`Ready: True` is not wrong on its own terms: a restorable base genuinely is available. But it means a failed base build produces no alert, no degraded status, and no user-visible symptom. The only way to notice is to observe that a guest-side behaviour change did not take effect, which is exactly what happened here and only because the change was one whose absence was easy to query.

This is the advertise-versus-enforce shape. The Workload advertises readiness against the old base while the desired digest has moved, and nothing reconciles the gap loudly.

## Suggested direction

Not prescribing the fix, but the options seem to be:

1. Surface the mismatch: a condition or metric for "restorable base was built from a digest other than the desired one", which is the fact that actually matters and is cheap to compute.
2. Retry a failed base build with backoff rather than leaving `BuildFailed` terminal until something else re-keys it.
3. Alert on `BaseBuilt: False` persisting beyond some window, since `Ready` will not carry it.

[...]

PR #4938, opened 2026-08-15
fix(embervm): re-key the claude-runtime base so diff capture reaches guests

jomcgi:
Unblocks #4936. Related: #4937.

## Why

PR #4936 added turn diff capture to the guest shim, and the runtime image built with it. Verified against the pinned digest:

```
crane export ghcr.io/.../runtimes/claude@sha256:c8a38ea... | grep -c
  _capture_turn_diff    2
  _capture_turn_base    2
  MAX_TURN_DIFF_BYTES   2
```

The base rebuild that would carry that into guests failed:

```
Ready:     True   BaseReady    "a restorable base snapshot is available"   17:29:17Z
BaseBuilt: False  BuildFailed  "{:connect, :timeout}"                      17:29:17Z
snapshotRef: claude-runtime__e847f766b497
```

17:29 is two minutes after the 16gi bricks started rolling for chart `0.14.1`, so the timeout is almost certainly collateral from that roll.

Nothing surfaced it. `Ready` stayed `True` on the older restorable base, so sessions kept succeeding while running the previous shim. Session 192 committed locally and its walkthrough still resolved rung 3 with `diff_unavailable`; `diff_blob` is null for every turn.

## Why an epoch bump rather than waiting

Both alternatives were tested rather than assumed:

- **Waiting**: the condition was unchanged 25 minutes later, same reason and same `lastTransitionTime`. A failed build is not retried.
- **Starting another session**: a qwen probe (session 193) created and served a guest. `BaseBuilt` stayed `False` with the same timestamp across six consecutive polls while the session went live and then banked. Creating a session does not trigger a rebuild.

The base signature deliberately excludes noded config, which is why `claudeRuntimeWorkload.initEnv` exists as a manual rebuild epoch. It is the documented mechanism and the only one that applies here.

## Verified

- `ci lint` clean.
- `helm template` renders the new epoch into the Workload manifest (1 occurrence), so the value actually reaches the object whose signature it re-keys.

## Not verified

That the rebuild succeeds. This re-keys the base and forces a build; if the underlying connectivity problem recurs, it will fail again the same way. I will watch `BaseBuilt` after rollout and confirm a session captures a diff before calling diff capture live.

## Note

That a transient timeout can only be cleared by hand-editing a string in a values file is the real problem, and it is filed as #4937 rather than fixed here.

### DIFF
--- a/projects/embervm/deploy/values.yaml
+++ b/projects/embervm/deploy/values.yaml
@@ -562,7 +562,20 @@ claudeRuntimeWorkload:
   # nothing the guest can read. Use it as a rebuild epoch, never as a knob.
   # Making it mean what it says is tracked in #4429.
   initEnv:
-    EMBER_BASE_EPOCH: "warm-restore-3"
+    # diff-capture-1: PR #4936 put turn diff capture in shim.py, and the image
+    # built with it (verified: the pinned runtime digest contains
+    # _capture_turn_diff). The base rebuild that would carry it into guests
+    # failed at 17:29Z with {:connect, :timeout}, two minutes after the 16gi
+    # bricks started rolling for chart 0.14.1. Ready stayed True on the older
+    # restorable base, so nothing alerted and every session kept running the
+    # previous shim, capturing nothing.
+    #
+    # A failed build is not retried, and creating a session does not trigger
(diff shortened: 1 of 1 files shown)


jomcgi:
## Helm Manifest Diff

PR #4939, opened 2026-08-15
fix(embervm): let the warmth reaper reclaim pre-sidecar group orphans

jomcgi:
## What

Gives `:group` the same empty-binding recovery that PR #4905 gave `:stateful`.

## Why

#4905 removed the empty-binding skip for stateful bundles and left the identical clause for group sets, with a stated reason:

> GROUP_SET is not searchable by the same ref in this reaper. Its remote layout is keyed by group_instance_id, so the conservative empty-binding skip remains.

Both sentences are true. The conclusion only follows if the index must be shared. The stateful index parses `stateful/<vendor>/<workload>/<ref>/<file>` into `ref => binding`, and group keys genuinely do not fit that shape. But the store key is `<kind>/<vendor>/<workload>/<ref>`, so a group set is `group_set/<vendor>/<gid>/<set_id>/<member>/<file>`, and a second index keyed by `set_id` recovers exactly the binding the skip calls unrecoverable.

## Measured in production today

On the node-4 brick:

| | |
|---|---|
| local group sets | **6**, 18 memfiles, **60 GiB** |
| created | 2026-07-16 |
| `group_instance_id` binding files present | **0**, so every set hits the skip |
| local group ids | `grp-01KXVQ...`, `grp-01KXVV...`, `grp-01KXW0...` |
| group ids in the store | `grp-01KY8Y...`, `grp-01KY91...`, `grp-01KY92...`, `grp-01KY97...` |
| **overlap** | **ZERO** |

Same shape as #4905's stateful finding: the skip protects nothing while costing 60 GiB indefinitely, and it cannot resolve itself because nothing else will supply the missing binding.

Note what this implies for the reclaim itself. Every one of the six sets misses the index, so they all take the local-only path: this PR's practical effect is a pure local reclaim with nothing deleted from the store. The branch that deletes remotely is the one that has a binding to address it with.

## How

Mirrors the stateful implementation clause for clause:

- `default_remote_group_index/0` lists `group_set/` and folds it into `set_id => %{vendor, group_instance_id}`, with the same conflicting-binding `:halt`
- `remote_group_index/2` with the same "only look up when some entry has an empty binding" gate and the same `rescue`/`catch` wrappers
- three-way `evict_entry`: recovered binding evicts local then remote; index miss evicts local only; lookup failure still skips
- the inline group eviction body is extracted to `evict_group/4`, symmetric with `evict_stateful/4`
- `log_dry_run_entry/2` gains group clauses so a gate-off sweep reports which path each orphan would take

`run_evict/4` is untouched, so the single remote GROUP_SET evict still fires only when **every** member's local eviction succeeded.

`vendor: vendor || ""` on the remote request is a no-op for the existing path: the old request omitted the field, which protobuf renders as `""`.

## Not in this PR

**Arming.** `warmthRetention.sweepEnabled` is unchanged. The dry-run lines now report what would be reclaimed and by which path; reading those and arming is a separate supervised step, the way the S3 GC was armed on 2026-08-04 after its dry-run manifest was read.

## Flagged: is `group` retired or active?

`deploy/values.yaml` sets `warmthS3Gc.allowEmptyKinds: "group"` as the operator statement that the scratch-k8s demo is retired and `GroupStore` legitimately tracks zero instances. But the four group ids in the store are dated 2026-08-14, and those same keys are the 16 entries the S3 sweep reports as `ambiguous_keys`.

That matters for what this PR is. If group is genuinely dead, this is a one-time 60 GiB reclaim. If something is still producing group sets, the skip is an ongoing leak and this fix stops it accruing. I have not resolved which, and it does not block the fix either way, but it should be settled before arming.

## Known behaviour worth recording

`dial_for_group` requires a non-empty gid and otherwise falls open to the bare `node_id`. In the recovered path the entry's workload is rewritten, but the node's fact still reports an empty `group_instance_id` (that is what made it an orphan), so the dial falls back to `node_id` in both new paths. That is safe here because co-located bricks share one scratch mount, so either instance's evict removes the same file. It would misroute on a node whose siblings did not share scratch.

## Testing

`ci lint` clean. `ci test` green:

```
1312 tests, 0 failures, 6 skipped
```

Note the Elixir suite runs as the `//bazel/erlang:mix_test_smoke` **genrule**, so it never appears in Bazel's `Executed N out of M tests` line (that line reads `0 out of 719` here and is not the signal for this change). The ExUnit summary above is.

Test count in `warmth_reaper_test.exs` goes 16 to 20: five added, one removed. The removed test asserted the old skip ("runs no member local and no remote evict"), which is exactly the behaviour this changes; the remaining no-evict case is covered by the index-failure test.

All four paths assert on the captured requests rather than on the return value, because every clause returns `:ok` including the skip, so a return-value assertion would pass against the bug:

- remote hit: local members evicted, then remote with the recovered vendor and gid
- remote miss: `assert_receive` both members, `refute_receive` the remote GROUP_SET evict
- index failure: nothing evicted
- non-empty gid: unchanged behaviour
- plus `{:error, {:conflicting_bindings, set_id}}` for two bindings on one set id

### DIFF
--- a/projects/embervm/control/lib/embervm/warmth_reaper.ex
+++ b/projects/embervm/control/lib/embervm/warmth_reaper.ex
@@ -125,6 +125,9 @@ defmodule Embervm.WarmthReaper do
     remote_stateful_index_fun =
       Keyword.get(opts, :remote_stateful_index_fun, &default_remote_stateful_index/0)
 
+    remote_group_index_fun =
+      Keyword.get(opts, :remote_group_index_fun, &default_remote_group_index/0)
+
     # 0 disables the timer entirely (the unit-test default, so a test drives
     # :sweep_now explicitly and asserts deterministically); production uses the
     # module default.
@@ -144,6 +147,7 @@ defmodule Embervm.WarmthReaper do
       evict_snapshot_fun: evict_snapshot_fun,
       evict_artifact_fun: evict_artifact_fun,
       remote_stateful_index_fun: remote_stateful_index_fun,
+      remote_group_index_fun: remote_group_index_fun,
       sweep_interval_ms: sweep_interval_ms,
       sweep_enabled: sweep_enabled
     }
@@ -258,9 +262,10 @@ defmodule Embervm.WarmthReaper do
   defp apply_plan(state, plan) do
     {stateful, group} = Enum.split_with(plan, &(&1.kind == :stateful))
     stateful_index = remote_stateful_index(state, stateful)
+    group_index = remote_group_index(state, group)
 
     log_and_evict(state, :stateful, stateful, stateful_index)
-    log_and_evict(state, :group, group, nil)
+    log_and_evict(state, :group, group, group_index)
     :ok
   end
 
@@ -309,8 +314,9 @@ defmodule Embervm.WarmthReaper do
   # uses); the single remote EvictArtifact{remote: true, kind: GROUP_SET, workload:
   # group_instance_id, ref: set_id} that drops the whole set prefix at once fires
   # ONLY when EVERY member's local eviction succeeded.
-  # GROUP_SET is not searchable by the same ref in this reaper. Its remote layout
-  # is keyed by group_instance_id, so the conservative empty-binding skip remains.
+  # Empty on-disk group_instance_id bindings are recovered from a remote GROUP_SET
+  # index keyed by set_id. The indexed value supplies the vendor and
+  # group_instance_id needed to address the remote prefix safely.
 
   defp evict_entry(state, %{kind: :stateful, id: ref, workload: workload} = entry, {:ok, index})
        when workload in [nil, ""] do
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

jomcgi:
Post-deploy result: **this fix did not reclaim the 60 GiB**, and the reason is one layer above the clause it changed.

Live on chart 0.14.4 (CP pod restarted 19:58:14Z), and 30 minutes later node-4 still holds all 18 memfiles / 61440 MiB. The reaper logged nothing at all, which is itself the signal: `log_and_evict(_state, _kind, [], _index), do: :ok` returns silently on an empty plan, so silence means these sets never entered the plan.

Why they cannot. On-disk shape of the orphaned sets:

```
set-grp-01KXVQAGWQX81GBMD593QTZME8-1784416305287/
  server/   memfile snapfile
  agent-0/  memfile snapfile
  agent-1/  memfile snapfile
```

Two files per member. The store layout for a healthy set carries four:

```
group_set/<vendor>/<gid>/<set_id>/<member>/{memfile,snapfile,member.json,group_instance_id}
```

No `member.json`, no `group_instance_id`, at any level. These predate the binding sidecar completely, so the node's boot scan has no member metadata to build a `group_bundle_sets` entry from, they never appear in `NodeStatus`, and the reconcile that produces the eviction plan enumerates from `NodeStatus`.

So the empty-binding clause this PR fixed is real and worth having, and #4905's stateful analogue proves the pattern. But it only fires for sets that are *reported* with an empty binding. These are not reported at all.

Reclaiming them needs a different mechanism: either the boot scan learns to emit a minimal entry for a member directory holding only `memfile` + `snapfile`, or they are collected out of band the way the S3 warmth GC handles prefixes the reaper cannot reach. That is a separate change and I have not opened one.

Correcting my own expectation in this PR's body: I said it would reclaim ~60 GiB on deploy. It did not, and the projection was wrong because I reasoned from the eviction clause without checking whether the entries reach it.

Issue #4946, opened 2026-08-15
monolith: GitHub mutation broker as first delegation consumer


jomcgi:
## Approved disposition: gate on a concrete permission gap

Retain needs-thought. #4944 and #4946 share an evidence gate, not a circular dependency requiring one speculative implementation to justify the other.

## Evidence gate

Name an agent operation requiring per-issue permissions that the existing factory path cannot enforce. Record the caller, exact permitted issue/action set, prohibited actions, current authorization path and demonstrated enforcement gap. Define the intended allowed and denied behavior before selecting an implementation.

A generic conductor act tier, a retired merge-queue caller or the existence of the other proposal does not satisfy this gate. #5789 independently verifies existing factory MCP controls.

## Scope after the gate

Choose the smallest enforcement change for the named operation through existing owners. A new broker or delegation token system is not preapproved. If existing enforcement can close the gap, prefer that bounded correction and close the unnecessary proposal.

Preserve existing authentication, actor attribution, target checks, retry/control semantics and secret handling. No blanket ban on existing supported GitHub access is introduced. This decision changes issue readiness only, not credentials, permissions or runtime access.

## Broker-specific boundary

Only if the selected correction requires a broker, implement the named operation and exact issue/action restriction. Keep coarse credentials server-side, enforce authority at the mutation owner and record actor/subject attribution. Do not expose a general GitHub mutation catalogue or rebuild unrelated access paths.

Required acceptance must include authorized operation, wrong issue/verb, unauthorized caller and duplicate/uncertain outcomes. Remove the retired reconciler as a proposed first tenant. Record the selected contract before making implementation agent-ready.

jomcgi:
## Agent brief
Verdict: needs-human (recommend: defer)

### Outcome
The GitHub mutation broker is the right first delegation consumer, but it cannot be built to its stated contract yet: its core check consumes delegation grants whose encoding is owned by still-open #4944, and its alignment target ADR agents/055 is Draft. The actionable path is to defer full delivery until #4944 lands, or narrow scope to a standing-scope slice now.

### Acceptance
Full broker, after #4944:
- MCP tools update_issue, comment, link_pr, and label exposed by the monolith; no agent path touches GraphQL or REST directly.
- The coarse repo-scoped credential stays server-side (1Password Operator, never hardcoded) and is reachable only from the broker handler.
[...]

Issue #4945, opened 2026-08-15
embervm: session delegation delivery via placeholder swap


jomcgi:
## Approved disposition: gate on the selected delegation design

Retain needs-thought. This issue follows the concrete permission-gap decision in #4944/#4946 and proceeds only if the selected design actually requires a new session delegation delivery path.

## Evidence gate

Record the named consumer, selected authorization/delegation contract and why existing credential delivery cannot meet it. The existence of a token proposal alone does not justify placeholder-swap expansion.

Do not require this implementation before choosing the permission solution. If that solution needs no new delivery path, close this issue with the decision.

## Constraints if needed

- Bind session identity and swap ownership through the trusted spawner, never through a guest-asserted identity.
- Keep usable credentials outside the guest; the guest receives only an inert reference if the selected design uses placeholders.
- Deny access after termination. Do not remove an entry if removal would permit uncredentialed fall-through; preserve an explicit deny state where required by the actual current contract.
- Respect selected token lifetime and authority bounds. Add refresh or session-specific swap behavior only where the named consumer requires it, through existing lifecycle owners.
- Validate wrong-session access, stale references, termination and retry behavior against current implementation before making the bounded delivery change agent-ready.

## Boundaries

No speculative token refresh framework, new guest authentication protocol or rollout under this triage decision. Existing credential controls and separately tracked security fixes remain required and are not gated by this optional expansion.

This scope supersedes the earlier unconditional session-level placeholder-swap implementation brief.

jomcgi:
## Agent brief

Verdict: `agent-ready`.

### Outcome

`agent-ready`. Session-scoped delegation delivery through the Ember egress swap is actionable as specified: provision the guest with an inert placeholder, swap it for the real session delegation token at the vsock/egress hop with the usable token living control-plane-side only keyed by session, refresh control-plane-side with no guest refresh protocol, flip the swap entry to a DENY TOMBSTONE on workload death, and bind swap keys at spawn by the spawner.

### Acceptance

1. Guest carries only an inert placeholder and never the usable delegation token; the egress hop replaces the placeholder signal with the real session token, which lives control-plane-side only and is keyed by session.
2. Refresh of the underlying short-TTL token happens control-plane-side and is invisible to the guest; no refresh protocol exists in agent environments.
3. Workload death flips the swap entry to a DENY TOMBSTONE and never removes it, with a test proving a removed entry would fall through to uncredentialed pass-through (the documented catalog inversion).
[...]

Issue #4944, opened 2026-08-15
monolith: delegation minting with attenuation-only grants


jomcgi:
## Approved disposition: gate on a concrete permission gap

Retain needs-thought. #4944 and #4946 share an evidence gate, not a circular dependency requiring one speculative implementation to justify the other.

## Evidence gate

Name an agent operation requiring per-issue permissions that the existing factory path cannot enforce. Record the caller, exact permitted issue/action set, prohibited actions, current authorization path and demonstrated enforcement gap. Define the intended allowed and denied behavior before selecting an implementation.

A generic conductor act tier, a retired merge-queue caller or the existence of the other proposal does not satisfy this gate. #5789 independently verifies existing factory MCP controls.

## Scope after the gate

Choose the smallest enforcement change for the named operation through existing owners. A new broker or delegation token system is not preapproved. If existing enforcement can close the gap, prefer that bounded correction and close the unnecessary proposal.

Preserve existing authentication, actor attribution, target checks, retry/control semantics and secret handling. No blanket ban on existing supported GitHub access is introduced. This decision changes issue readiness only, not credentials, permissions or runtime access.

## Delegation-specific boundary

Only if the selected correction requires delegation, define a bounded grant for the named consumer. It must not widen parent authority or outlive its parent, and must preserve the subject and attributable actor chain. Use existing credential delivery mechanisms where sufficient; do not add a general token-exchange framework, speculative provider migration or standing client per executor role.

Required acceptance must include attempted scope widening, expired or invalid parent authority and wrong-target use. Record the selected contract before making implementation agent-ready.

jomcgi:
## Agent brief

Verdict: `agent-ready`.

### Outcome

`agent-ready`. Delegation minting in `monolith/auth` is actionable as specified: short-lived `(subject, actor, scope)` tokens with RFC 8693-compatible `act` chaining, built on the existing `Principal` model and the reserved verifier slot, with the issue invariants as acceptance criteria.

### Acceptance

1. Minting requires presenting a verified parent credential; the minted scope is a strict subset of the parent scope (attenuation only, enforced at mint time, with a test that equal or wider scope is rejected).
2. Delegated TTL is short and never exceeds the parent remaining lifetime (TTL is the revocation story; the numeric bound is asserted in tests).
3. Re-delegation extends the actor chain and never changes the subject (tests for chain growth and subject immutability).
[...]

Issue #4943, opened 2026-08-15
authentik: add a workload identity only for a named authentication gap


jomcgi:
## Approved disposition: gate on a named caller and authentication gap

Retain needs-thought. Provision a new standing workload identity only when a named factory caller needs authentication or permissions that existing supported authentication cannot adequately provide.

## Evidence gate

Record the caller, required operations, resource scope, existing authentication path and concrete gap. An executor role such as implementer, reviewer or researcher does not by itself require a separate OAuth client or grant permission.

Neither a generic conductor programme nor the existence of delegation/broker proposals satisfies the gate. Remove the retired merge-queue reconciler and hypothetical BOSUN/cron tenants from the active implementation scope.

## Scope after the gate

Provision only the named caller through existing Authentik identity and 1Password secret mechanisms, with the minimum required permissions and explicit ownership. Revalidate current verifier, issuer/audience, credential lifecycle and deployment contracts before implementation; historical source claims are not current verification.

Define focused acceptance for the required caller and rejection of unauthorized callers/operations. Do not create a client per task or workload role, broaden shared credentials or introduce another identity plane.

## Boundaries

No credentials, groups, permissions or live configuration change under this triage decision. Existing supported authentication remains in use.

The bundled blueprint-apply health advisory is removed from this scope. A demonstrated blueprint visibility problem may justify a separate bounded correction; it is not a prerequisite to generic identity work.

This gate supersedes the earlier needs-human provisioning question and speculative tenant list.

jomcgi:
## Agent brief

### Outcome
This issue is still actionable but not agent ready as written, because the first tenant is now a choice between two open paths and no workload blueprint exists yet to copy. The verifier side is confirmed ready in the checkout, the blueprint pattern to model on exists, and the linked history checks out (#4921 is closed, #4946 is open), so the remaining gap is a human scope call on which workload goes first, plus the security relevant visibility hook for blueprint apply failures.

### Acceptance
[...]

PR #4955, opened 2026-08-15
feat(monolith): add the auth domain, Principal type, and token verification

### LINKED ISSUE #4942: monolith: auth/ verification middleware and Principal type
Implement `projects/monolith/auth/`: token verification middleware and the Principal type. Per #4569 item 1: verify RS256 against the authentik JWKS (`https://auth.jomcgi.dev/application/o/mcp-friends/jwks/`), strict `iss`, `aud` must contain `https://private.jomcgi.dev`; expose the Principal (subject, actor chain, scope, groups, email, standing-vs-delegated flag) in request-scoped context. Two verifiers behind one interface: authentik-issued standing tokens and monolith-issued delegations. Wire as FastAPI dependency for routers and FastMCP middleware for `core/mcp_app.py` (currently 9 lines, no auth). Hard constraint from #4569: absent token yields an anonymous least-privilege Principal, never 401 (tool-refresh CronJob and gateway federation call with no user context). No monolith BUILD target currently depends on any JWT library; adding one is part of this work (`@pip//package` via aspect_rules_py, not requirement() syntax).

jomcgi:
Closes #4942. Refs #4940 (design), #4569 (constraints), #4944 (next consumer).

Adds `projects/monolith/auth/`, a domain that turns bearer material into a verified Principal. It makes no authorization decisions: it answers "who is this and what do they carry", and each resource-owning domain decides what that Principal may do. A per-domain rule accumulating here is the smell.

## Fail-mode semantics, which are the point of the change

**Absent bearer material yields an anonymous least-privilege Principal, never 401.** Context Forge's tool-refresh CronJob and gateway federation call with no user context, so requiring a token would empty the tool catalogue. A missing `Authorization` header and a non-Bearer scheme both count as absent.

**Present but invalid always raises, and never downgrades to anonymous.** Caller faults (bad signature, wrong issuer, wrong audience, expired, malformed, unknown kid after refresh, unrecognised by every verifier) render 401 with `WWW-Authenticate: Bearer`. Verifier infrastructure faults (JWKS unreachable, JWKS malformed, identity settings unconfigured) render 503. Both raise the same `AuthError` carrying a `reason` enum, so the split is observable in logs and tests rather than only in the status code. A misconfigured JWKS URL breaks loudly instead of quietly de-scoping every caller to the public tier.

Tests assert `pytest.raises(AuthError)` with the reason, not merely that the result is non-anonymous, so an inversion fails them.

## What ships

- `Principal`: frozen dataclass carrying subject, an ordered actor chain, scope, groups, email, kind (human or workload) and authority (standing, delegated, anonymous), with `has_group` / `has_scope` and no policy.
- `AuthentikStandingVerifier`: RS256 against the authentik JWKS, strict issuer, audience membership, and a required `exp`. PyJWT validates `exp` only when the claim is present, so `options={"require": [...]}` is what stops a token minted without an expiry from being honoured forever.
- An ordered verifier chain. The standing verifier **declines** (returns `None`) a token whose issuer is not ours, which is what leaves a slot control can actually reach for the delegation verifier in #4944. A decline with no verifier left to ask still fails closed as `UNRECOGNIZED`.
- `JwksCache`: a TTL governing the ordinary path and a separate floor governing caller-triggered unknown-kid refreshes, so N requests bearing random key ids cannot each cause an outbound fetch serialized behind the cache lock. Failures are remembered for the same window so a down IdP fails fast rather than paying the full timeout per request. Monotonic clock, explicit httpx timeout.
- A FastAPI dependency for routers and an ASGI middleware for the FastMCP mount, both exposing the Principal request-scoped. No tool handler behaviour changes: result scoping is #4569's own work.
- Env wiring across code, chart template and `deploy/values.yaml` together, since a rename touching fewer than all three silently unrenders the block.

## Known constraint, recorded on #4569

On the SSE transport the middleware authenticates the **stream, not the message**. `handle_sse` runs the whole MCP session loop inside the stream-opening GET, while `handle_post_message` only writes to an anyio memory object stream and returns 202, and contexts do not travel through that stream. So `current_principal()` read inside a tool returns the stream-opener's identity. This is latent here because no tool reads the Principal yet, and it is documented at the mount. It stops being latent when #4569's result scoping lands, which is why it is written up there with options.

Not addressed, deliberately: delegation minting (#4944), result scoping (#4569), the GitHub broker (#4946), authentik blueprints (#4943), the Ember placeholder swap (#4945).

## Verification

`ci` green on this tree. Because a cached pass proves nothing about new targets, the auth and framework targets were force-executed:

```
ci test //projects/monolith:auth_verifier_test //projects/monolith:auth_plumbing_test \
        //projects/monolith:framework_core_test -- --nocache_test_results --test_output=all

Executed 3 out of 3 tests: 3 tests pass.
```

40 auth cases plus `framework_core_test`, including `test_private_app_registers_the_auth_error_handler`, which asserts against an app built the production way rather than one the test wires itself.

`helm template monolith projects/monolith/chart/ -f projects/monolith/deploy/values.yaml` renders all four `AUTH_*` variables into the `backend` container. `Chart.yaml` `version:` and `deploy/application.yaml` `targetRevision:` are untouched.

## Note on the JWT dependency

`pyjwt[crypto]==2.13.0` is already in `bazel/requirements/all.txt`, locked transitively via `mcp` and `semgrep`, so `@pip//pyjwt` resolves with no lock change and this PR touches neither `pyproject.toml` nor `bazel/requirements/`. Promoting it to a direct dependency is a reasonable follow-up: a transitive pin disappears if `mcp` drops it, though that fails loudly at build time rather than silently at runtime.

### DIFF
--- a/projects/monolith/BUILD
+++ b/projects/monolith/BUILD
@@ -1,5 +1,6 @@
 # gazelle:exclude agent
 # gazelle:exclude agent_sessions
+# gazelle:exclude auth
 # gazelle:exclude artifact
 # gazelle:exclude chat
 # gazelle:exclude chat_public
@@ -88,6 +89,7 @@ _BACKEND_SRCS = glob(
         "agent/**/*.py",
         "agent_sessions/**/*.py",
         "app/**/*.py",
+        "auth/**/*.py",
         "core/**/*.py",
         "framework/**/*.py",
         "artifact/**/*.py",
@@ -284,6 +286,29 @@ py_venv_binary(
 # The packages in the agent/chat/home import cycle share one
 # target because Bazel deps must be acyclic; everything else is a DAG.
 py_library(
+    name = "pkg_auth",
+    srcs = glob(
+        [
+            "auth/**/*.py",
+        ],
+        exclude = [
+            "**/*_test.py",
+            "**/test_*.py",
+            "*/tests/**/*.py",
+        ],
+    ),
+    imports = ["."],
+    visibility = ["//:__subpackages__"],
+    deps = [
+        "@pip//fastapi",
+        "@pip//httpx",
+        # PyJWT is currently locked transitively through mcp and semgrep.
+        "@pip//pyjwt",
+        "@pip//starlette",
+    ],
+)
+
+py_library(
     name = "pkg_core",
     srcs = glob(
         [
@@ -370,6 +395,7 @@ py_library(
     imports = ["."],
     visibility = ["//:__subpackages__"],
     deps = [
+        ":pkg_auth",
         ":pkg_core",
         "@pip//certifi",
         "@pip//fastapi",
@@ -1307,6 +1333,7 @@ py_library(
             "agent/**/*.py",
             "agent_sessions/**/*.py",
             "app/**/*.py",
+            "auth/**/*.py",
             "core/**/*.py",
             "framework/**/*.py",
             "artifact/**/*.py",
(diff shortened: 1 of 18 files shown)


jomcgi:
## Helm Manifest Diff

PR #4956, opened 2026-08-15
fix(agents): log why a guest diff payload is discarded

jomcgi:
Supports #4934. Complements the guest-side instrumentation being added separately: this is the same blind spot on the monolith side of the wire.

## Why

`_guest_diff` returns `None` on eight separate conditions and logs nothing on any of them:

- payload is not a mapping
- missing one of `base_sha` / `zlib_b64` / `truncated`
- `base_sha` fails the 40 to 64 char hex check
- `truncated` is not a bool
- a truncated payload carried a blob anyway
- `zlib_b64` is not a string
- compressed over 1 MiB, or uncompressed over 5 MiB
- undecodable base64, or a zlib stream that is incomplete or has trailing data

An absent payload also returns `None`. So the two are indistinguishable in the database: a null `diff_blob` means either the guest sent nothing, or the guest sent something that was rejected here, and the row cannot tell you which.

That matters right now. Diff capture is inert in production and the remaining hypotheses are that the guest's `_capture_turn_diff` returns None, or that the payload is rejected after it arrives. Guest-side instrumentation alone cannot separate those: a guest that captures correctly and trips one of the eight conditions above would log success while the column stays null.

Worth noting what this rules out. The wire itself is not a suspect: the shim sends the whole record (`self._send(200, record)`, no key whitelist), noded carries it as `bytes body` described in the proto as "forwarded verbatim", and the control plane's `send_guest_result` does `send_resp(conn, status_code, body)` and filters headers only. Nothing between the guest and this function can drop a JSON key, which is what makes this function the only remaining place a present payload can vanish.

## What changed

Each rejection logs a distinct reason with the session id. Sizes are included where relevant, so an over-cap diff reports its actual size rather than just failing.

An absent payload stays silent. Guests predating diff capture send nothing, that is not a fault, and warning on every turn they take would be noise that trains you to ignore the log.

Behaviour is unchanged: every path returns exactly what it returned before, including the truncated-with-blob case that returns `None` and the truncated-without-blob case that returns the value.

## Verified

- `ci lint` clean.
- `ci test //projects/monolith:agent_sessions_transport_test` -> `Executed 1 out of 1 test: 1 test passes`.
- New tests assert a distinct reason per rejection shape, that the session id appears in the log, that an absent payload logs nothing, and that a valid payload is returned unchanged and silently.

## Not verified

That this fires in production, because it only fires when a guest sends a diff at all, and no guest is currently known to be doing so. If diff capture starts working, this is silent; if capture works and validation rejects, this is the line that says so.

### DIFF
--- a/projects/monolith/agent_sessions/transport.py
+++ b/projects/monolith/agent_sessions/transport.py
@@ -124,39 +124,66 @@ class Turn(NamedTuple):
     diff: dict | None = None
 
 
-def _guest_diff(value) -> dict | None:
-    """Validate optional guest diff metadata without making it turn-critical."""
+def _reject_guest_diff(session_id, reason: str) -> None:
+    """Log why a present diff payload was discarded, then discard it.
+
+    Every rejection below returns None, which is also what an absent payload
+    returns, so without this the two are indistinguishable: a null diff_blob
+    means either the guest sent nothing or the guest sent something this
+    rejected, and the database cannot tell you which. A guest that captures
+    correctly and fails validation here would look identical to one that never
+    captured at all.
+    """
+    logger.warning("discarding guest diff for session %s: %s", session_id, reason)
+    return None
+
+
+def _guest_diff(value, session_id=None) -> dict | None:
+    """Validate optional guest diff metadata without making it turn-critical.
+
+    An absent payload is silent: guests predating diff capture send nothing and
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

PR #4957, opened 2026-08-15
feat(agents): make the walkthrough the main feature of a finished session

jomcgi:
Follows #4933, which added the summary and fixed the counts but left the panel collapsed. This is the half that was still missing: prominence.

## Why

The walkthrough was a collapsed `<details>` behind a one-line `<summary>`, so a finished session showed a wall of transcript and a small closed strip. Once the work is over, the explanation of what the turn changed is the thing the reader came for.

## What changed

The last turn of a session whose VM is not awake renders open and full width, framed so it carries more weight than the turn separators:

```svelte
prominent={vmState(selectedSession, vms) !== "awake" &&
  turnIndex === (detail?.turns ?? []).length - 1}
```

It stays in the transcript column and in the normal scroll flow. This is not a modal, a takeover, or a route change: the conversation remains the spine of the page and getting back to it is just scrolling.

**Only the last turn is promoted.** A ten turn session that expanded ten two-pane panels would be worse than the disclosure it replaces. Earlier turns and awake sessions are unchanged.

## The subtle part

A declaratively open `<details>` does not reliably emit a `toggle` event after Svelte attaches its listener, so the existing one-shot load in `onToggle` would never fire for a panel that starts open. The load now also fires from `onMount`, sharing the same guards.

Without that, the panel renders open and never fetches, showing a permanent empty state, and `build_test` passes happily. The tests cover it: one asserts the payload loads exactly once when it renders open.

`opened` initialises under `untrack()` so it stays genuine local state the reader's clicks own, while `open={Boolean(fixture || prominent)}` remains reactive. A session that finishes while the page is open therefore still opens the panel, and that path fires the load through the normal toggle route.

## Verified

- `ci lint` clean.
- `ci test //projects/monolith/frontend:test //projects/monolith/frontend:build_test -- --nocache_test_results` -> `Executed 2 out of 2 tests: 2 tests pass`.
- The first run returned `Executed 0 out of 2`, both cached. That was a legitimate hit on the implementer's identical run through the shared action cache, but it proves nothing, so it was re-run uncached. Judging by the summary line rather than the exit code is the rule here for good reason.
- Confirmed the new test is genuinely in the target: `test_lib` globs `src/**/*.test.js` and lists `vitest.config.js` in `srcs`, so both the new file and the config change are real inputs rather than files that merely exist on disk.
- All six CSS custom properties used by the prominent styles resolve in `agents-theme.css`. An undefined `var(--ink)` would render the 4px border as nothing, shipping prominence as an invisible no-op with every test green.
- No new visible strings, so no lexicon changes were needed.

## Blast radius worth naming

`vitest.config.js` gains the Svelte plugin and `conditions: ["browser"]`, which is what makes component-level tests possible at all. That changes module resolution for **every** frontend test, not just the new ones, which is why the full suite was run uncached rather than trusting a targeted pass.

## Not verified

Nobody has looked at the promoted panel in a browser. The behaviour is covered by tests and the tokens are confirmed to resolve, but the visual weight is a judgment call that needs eyes.

Also worth stating plainly: diff capture is still inert, so a promoted walkthrough currently shows the agent's testimony and an honest "the diff is unavailable for this turn". The structure is right; the diff fills in when capture works.

### DIFF
--- a/projects/monolith/frontend/src/routes/private/agents/+page.svelte
+++ b/projects/monolith/frontend/src/routes/private/agents/+page.svelte
@@ -1421,7 +1421,7 @@
       </header>
       <div class="turns" bind:this={turnsEl}>
         <div class="turns-inner">
-          {#each detail?.turns ?? [] as turn (turn.seq)}
+          {#each detail?.turns ?? [] as turn, turnIndex (turn.seq)}
             <!-- {@const} has to be an immediate child of the block, not of the
                  element inside it, so this sits above <article> rather than
                  next to the markup that reads it. -->
@@ -1480,6 +1480,8 @@
                   sessionId={selectedSession.id}
                   turnSeq={turn.seq}
                   model={turn.model || selectedSession.model || "luna"}
+                  prominent={vmState(selectedSession, vms) !== "awake" &&
+                    turnIndex === (detail?.turns ?? []).length - 1}
                 />
               {/if}
               <div class="turn-meta mono">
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

PR #4958, opened 2026-08-15
fix(embervm): parse member-suffixed group_set keys and hold partially-parsed prefixes

### LINKED ISSUE #4334: embervm: S3 GC cannot parse member-suffixed group_set keys, 18 GiB stuck ambiguous
Found supervising the first armed sweep (#4317). The warmth S3 GC's parse_key handles `<kind>/<vendor>/<owner>/<ref>/<file>` (5 segments), but group snapshots write per-member subdirectories: `group_set/<vendor>/<gid>/<set_id>/<agent-N>/<file>` (6 segments). All 16 remaining group_set objects (~18 GiB, the retired scratch-k8s demo backlog that motivated the allowEmptyKinds escape) fail closed to `ambiguous_keys` on every sweep: never deleted, never reclaimed, invisible in both `held` and `eligible_beyond_caps`.

Fix: a parse clause for the member-suffixed group shape producing the `(gid, set_id)` prefix as the candidate (delete unit = the whole set prefix), plus a test with a six-segment fixture. Fail-closed behavior today means there is no urgency and no risk; this is reclaim completeness, not safety.

Evidence: armed manifest `gc-manifests/1785868386571.json` has `ambiguous_total: 16`, all group_set; the one 5-segment group prefix in the plan deleted fine.

jomcgi:
Closes #4334.

`Embervm.S3WarmthGc` is armed and destructive in production. Its `parse_key/3` handled at most 5 path segments, and group sets are the one artifact kind whose files live under per-member subdirectories, so every group key fell through to the `_ -> :ambiguous` catch-all. 16 objects, about 18 GiB, have been skipped on every hourly sweep since 2026-08-04, invisible in both `held` and `eligible_beyond_caps`.

## Verified before the fix, not after

Walked the SeaweedFS filer directly rather than trusting the issue text, because a pre-R7 backlog could plausibly have been un-vendored, which would have made these keys five segments and this fix wrong. All 16 are the vendored member-suffixed shape:

```
group_set/amd/<gid>/<set_id>/<member>/<file>
```

Four group instances, one set each, two members each, files `memfile` / `snapfile` / `member.json` / `group_instance_id`. Six 3 GiB memfiles plus sidecars is 18.0 GiB across exactly 16 objects, matching the `16 ambiguous keys skipped` the sweep still logged at 2026-08-15T22:01Z. The writer side agrees: `store.go` `Export` PUTs to `prefix + "/" + name` and `enumerateArtifactFiles` walks the set dir recursively, so a member's relative name carries its own segment.

## Defect 1: the missing parse clause

One clause for the 6-segment vendored shape, dropping the member segment so the delete unit stays the whole set. Every member's files then group into one candidate alongside the set's own `meta.json`, which lives at 5 segments and already parsed.

The legacy un-vendored member shape is deliberately NOT claimed. It is five segments, indistinguishable from a vendored flat key under a vendor token this build does not know (`@vendors` is `["amd", "intel"]`). Guessing there would compose a delete prefix one level too wide, `group_set/<vendor>/<gid>` instead of `.../<set_id>`, deleting every set belonging to that group instance. In an armed destructive sweep that is a fail-open, so it stays ambiguous and a test pins the gap.

## Defect 2: a partial parse produced a partial delete reported as complete

Not in the issue. Found while verifying, and it is why there is no `meta.json` left anywhere under `group_set/`.

`parse_candidates/6` grouped PARSED keys into candidate prefixes and set ambiguous keys aside separately, while `delete_prefix/2` deletes `entry.files`, only the parsed subset. A prefix with some parsed and some ambiguous keys therefore got a partial delete, logged as `DELETED prefix=... bytes=...` for the whole prefix. That is what the first armed sweep did: each set's 5-segment `meta.json` parsed alone and was deleted, so "the one 5-segment group prefix in the plan deleted fine" reclaimed about 100 bytes while reporting a set.

Nothing wanted was destroyed (the orphan predicate had already cleared those sets), but the module's moduledoc claims it is fail-closed everywhere and here it was not.

Now any candidate prefix that is a parent of an ambiguous key is held WHOLE, with reason `sibling_key_ambiguous`, and surfaced in the manifest instead of silently discarded. The trailing slash in the prefix test is load-bearing: it stops `session/amd/wl/ref` from shadowing on an ambiguous `session/amd/wl/ref-2/...`.

## Tests

The existing group fixtures used `artifact("group_set/amd/grp-1/set-9", ...)`, a FLAT layout the writer never produces, which is why the six-segment case was never exercised. Worse, because that fixture's meta.json and payload always parse together, the partial-delete path was unreachable in test, so an unsound path looked proven. They stay (they still cover the flat and meta paths) and a `group_set_artifact/4` helper builds the real layout alongside.

Seven new tests: whole-set delete with meta.json first, bytes summing every member's files, no ambiguous keys for a well-formed set, a live group instance holding its set, a set with NO meta.json (production's current state) ageing off `newest_modified_ms`, the legacy shape staying ambiguous, and the `sibling_key_ambiguous` guard.

## Verification

`//bazel/erlang:mix_test_smoke` is a genrule, not a test rule, so `bazel test` on it emits no `Executed N out of M` line and exits 4 with "No test targets were found" even when the suite passes. Judge it by the mix summary and the process count instead.

Baseline, unmodified `origin/main`:

```
1315 tests, 0 failures, 6 skipped
5 processes: 358 action cache hit, 4 internal, 1 remote.
```

This branch:

```
1322 tests, 0 failures, 6 skipped
5 processes: 358 action cache hit, 4 internal, 1 remote.
```

Delta is exactly the 7 new tests, and `1 remote` on both runs proves the genrule executed rather than replaying a cached result.

## Expected effect once deployed

The next sweep should report 0 ambiguous keys and plan four group prefixes totalling about 18 GiB. Their `meta.json` is already gone, so created-at comes from `newest_modified_ms` (2026-07-24), well past the 7 day group TTL.

`Chart.yaml` `version:` and `deploy/application.yaml` `targetRevision:` are untouched. No values changed: the flag is already armed.

### DIFF
--- a/projects/embervm/control/lib/embervm/s3_warmth_gc.ex
+++ b/projects/embervm/control/lib/embervm/s3_warmth_gc.ex
@@ -38,6 +38,8 @@ defmodule Embervm.S3WarmthGc do
       uptime is additionally required before any sweep.
     * Ambiguous key parses (a workload literally named like a vendor) are
       skipped, never guessed.
+    * A candidate prefix with any unparsed sibling key below it is held whole,
+      so a parsed subset is never deleted.
     * Deletes: hard allowlist {stateful/, session/, serving/,
       session-workspace/, group_set/}; every DELETE targets one
       fully-qualified LISTED key (never a prefix, never bucket-wide); meta.json
@@ -236,8 +238,9 @@ defmodule Embervm.S3WarmthGc do
          {:ok, group_keys} <- list_or_abort(state, "group_set/"),
          {:ok, snapshot} <- cp_snapshot(state),
          :ok <- check_empty_cp_state(snapshot, stateful_keys, session_keys, serving_keys, workspace_keys, group_keys, state.allow_empty_kinds) do
-      {candidates, ambiguous} = parse_candidates(state, stateful_keys, session_keys, serving_keys, workspace_keys, group_keys)
+      {candidates, shadowed, ambiguous} = parse_candidates(state, stateful_keys, session_keys, serving_keys, workspace_keys, group_keys)
       {eligible, held} = build_plan(state, snapshot, candidates)
+      held = shadowed ++ held
       plan = apply_caps(state, eligible)
       log_plan(state, plan, eligible, held, ambiguous)
 
@@ -462,7 +465,7 @@ defmodule Embervm.S3WarmthGc do
   # segment-2 membership in the known vendor set. Anything else (a workload
   # named like a vendor, an unexpected depth) is AMBIGUOUS: the whole key is
   # held aside and never composed into a deletable prefix (fail-closed).
-  # Returns {candidates_by_prefix, ambiguous_keys}.
+  # Returns {candidates_by_prefix, shadowed_prefixes, ambiguous_keys}.
   defp parse_candidates(state, stateful_keys, session_keys, serving_keys, workspace_keys, group_keys) do
     {s_parsed, s_ambiguous} = split_parsed(state, stateful_keys, "stateful")
     {se_parsed, se_ambiguous} = split_parsed(state, session_keys, "session")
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

PR #4959, opened 2026-08-15
fix(embervm): read the checkout as its owner so turn diff capture works

jomcgi:
Fixes the root cause behind #4934. Diff capture has never produced a diff in production; this is why.

## The evidence

With the outcome instrumentation from `33fe69c8e` in place, the guest named it on the first session:

```
ember-claude-shim: turn-diff phase=base outcome=rev_parse_failed checkout_dir=/workspace/src
ember-claude-shim: turn-diff phase=diff outcome=no_base_sha checkout_dir=/workspace/src
```

`git rev-parse HEAD` exits non-zero, so no base sha is captured and the diff never runs.

## Why that is strange, and what it means

The same guest clones, edits, commits and pushes the same repository without trouble. Session 190 pushed `codex/noded-test-audit` at `dcdde14c9`; 192 and 196 committed locally. Notably the outcome is `rev_parse_failed`, not `no_git_dir`, so `.git` exists.

The difference is **who runs git**:

```python
# hydration's validation, works
["git", "-C", checkout_dir, "rev-parse", "--verify", "HEAD"],
capture_output=True, **_cli_privilege_kwargs(),

# capture, fails
["git", "-C", checkout_dir, "rev-parse", "HEAD"],
stderr=subprocess.DEVNULL,
```

`_cli_privilege_kwargs()` returns uid/gid **only when the shim is root**. The shim is root, hydration clones under those kwargs, so the checkout belongs to the CLI uid. Capture called git as root against it and tripped git's dubious-ownership check. The agent runs as the CLI user, which is why every path except capture worked.

The observed asymmetry is itself the proof the shim runs as root: if it did not, `_cli_privilege_kwargs()` would return `{}`, both calls would be identical, and capture would succeed.

## Why not just drop privileges

That was the first attempt and it failed a test in a way worth recording. `test_process_manager_captures_diff_against_head_at_turn_start` builds a **real** git repository as root, so running capture as uid 65532 reproduced the same ownership refusal with the roles swapped. There is no single uid that owns the checkout in both contexts, so any fix that picks a uid breaks one of them.

Reading as root with `safe.directory` scoped to that one path works in both directions: root can always read, and the ownership check is the only obstacle. It is passed per invocation via `-c`, so nothing is written to any git config in the guest and no persistent weakening of the check occurs.

## git's stderr now reaches the log

It was going to `DEVNULL`. The instrumentation could say which guard fired but never why, which is exactly what made `rev_parse_failed` unactionable. git had already written the reason and it was being thrown away. The failure outcomes now carry it, bounded to 200 chars.

## Test stub

`shim_hydration_test`'s stub recognised the checkout by fixed argv positions (`command[1:3] == ["-C", dir]`), so it rejected the new shape. It now matches `-C <dir>` wherever it appears, which is what "targets our checkout" actually means, and reports the offending command when it fails rather than a bare "unexpected git command".

## Verified

- `ci lint` clean.
- `ci test //projects/embervm/runtimes/claude:shim_test //projects/embervm/runtimes/claude:shim_hydration_test -- --nocache_test_results` -> `Executed 2 out of 2 tests: 2 tests pass`.
- Run uncached and across **both** shim targets deliberately: running only `shim_test` is how a hydration break slipped through earlier today.
- New tests assert both capture calls scope `safe.directory` to the checkout, and that git's stderr reaches the outcome line.

## Not verified

That capture actually produces a diff in production. That needs this image built, a base rebuild to carry it into guests, and a session. The guest instrumentation will say `phase=base outcome=success` when it works, and will name any new reason if it does not.

### DIFF
--- a/projects/embervm/runtimes/claude/shim.py
+++ b/projects/embervm/runtimes/claude/shim.py
@@ -63,21 +63,56 @@ def _emit_turn_diff_outcome(checkout_dir, phase, outcome):
         pass
 
 
+def _git_read_argv(checkout_dir, *args):
+    """git argv for reading a checkout the shim does not own.
+
+    The shim runs as root while hydration clones under _cli_privilege_kwargs, so
+    the checkout belongs to the CLI uid. git refuses to operate on a repository
+    owned by another user ("detected dubious ownership") and exits non-zero,
+    which is why capture reported rev_parse_failed on every turn while the agent
+    used the same repo without trouble.
+
+    safe.directory is scoped to this one path and passed per invocation, so
+    nothing is written to any git config. Dropping to the CLI uid instead would
+    only move the mismatch: the checkout is not always owned by that user
+    either, and a read as root is the one thing that works in both directions.
+    """
+    return [
+        "git",
+        "-c",
+        "safe.directory=%s" % checkout_dir,
+        "-C",
+        checkout_dir,
+        *args,
+    ]
+
+
 def _capture_turn_base(checkout_dir):
     """Best-effort checkout HEAD capture. A failure must not affect the turn."""
     try:
         if not os.path.exists(os.path.join(checkout_dir, ".git")):
             _emit_turn_diff_outcome(checkout_dir, "base", "no_git_dir")
             return None
+        # Run git as the CLI user, exactly as hydration does. The shim is root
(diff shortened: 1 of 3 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4569, opened 2026-08-09
mcp: fine-grained authorization for mcp.jomcgi.dev: monolith token validation, tool tiers, group provisioning


jomcgi:
## Status 2026-08-22

**Landed since this was written**

- Item 1, first half: the monolith verifies authentik tokens itself. #4955 added `projects/monolith/auth/` (`PrincipalMiddleware`, `JwksCache`, `Principal`, `current_principal()` in `auth/dependencies.py`). Absent token resolves to an anonymous principal, never a 401, so the tool-refresh constraint holds.
- The SSE per-stream identity problem in the 2026-08-15 comment is gone: #5036 moved the `/mcp` mount to streamable HTTP with `stateless_http=True` (`projects/monolith/framework/core.py`), so each message carries its own principal. Option 3 from that comment is what shipped.
- A CiliumNetworkPolicy now exists in the `monolith` namespace (`monolith-api-ingress`, from `chart/templates/cilium-ingress-policy.yaml`), so the first bullet of item 5 is addressed.

**Premise change: ADR agents/059 (Draft, 2026-08-21) removes Context Forge as the MCP entry point.** `docs/decisions/agents/059-authentik-federates-monolith-serves-mcp.md` decides that `mcp.jomcgi.dev` goes straight to the monolith `/mcp` mount. If it lands, items 2 (tool tiers in Context Forge), 3 (team provisioning in `provision-mcp-auth.sh`) and most of item 5 (`friends-empty`, `APP_DOMAIN`, hardcoded `serverId`, `tokenReplayDeny`) are deleted rather than finished. Do not start them until 059 is accepted or rejected. Execution of 059 is tracked in #3832 and #3833.

**What remains here regardless of 059**

- Item 1, second half: result scoping by group. No tool reads `current_principal()` for scoping yet (`projects/monolith/knowledge/mcp.py` has no principal check; `agent_sessions/mcp.py` reads it for the voice surface only). The rule stays as written: token absent or anonymous means public tier results, valid token means scope by group.
- Item 4: the trust-boundary ADR was broadened into #4941 under the #4940 auth-domain program. Close item 4 here when #4941 lands.

**Acceptance for what remains**: `search_knowledge` called with an `homelab-admin` token returns rows the same call without a token does not, with a test exercising both principals on one tool; the decision on items 2, 3 and 5 is recorded once 059 resolves.

---

## Current architecture (as of 2026-08-09, PRs #4557 to #4567)

[...]

PR #4961, opened 2026-08-15
fix(monolith): gate verifier ownership before rejection, fetch JWKS in-cluster

jomcgi:
Three hardenings found while reviewing #4955 after it merged. All verified against the live cluster, not just tests.

## 1. The verifier rejected tokens before deciding whether it owned them

`AuthentikStandingVerifier` ran every check that *raises* (`alg != RS256`, missing `kid`, undecodable) **before** the `iss` gate that *declines*. That makes slot 1's opinion final for the whole chain: a token belonging to a later verifier gets a 401 no later slot can revisit.

This matters because `build_default_resolver` states that #4944's delegation verifier registers after this one, and those tokens need not be RS256-with-kid. The chain existed for exactly the token shape that could not reach it.

The ownership gate now runs first, and only `iss` decides ownership. `UNCONFIGURED` deliberately stays ahead of the gate: a verifier that does not know its own settings cannot know what it owns, and declining there would render an operator fault (503) as a caller fault (401).

**Falsified, not assumed.** Against the old verifier with the new tests, 2 of 4 fail with exactly `401: authentication failed: malformed`:

| test | old code | why |
|---|---|---|
| `test_non_jwt_credential_declines_to_the_next_verifier` | FAILED | opaque credential raised instead of declining |
| `test_foreign_issuer_token_of_another_algorithm_declines` | FAILED | HS256 token raised at the alg pin |
| `test_token_claiming_our_issuer_is_judged_here_not_passed_on` | PASSED | guards against over-correcting |
| `test_unconfigured_verifier_raises_even_for_a_token_it_does_not_own` | PASSED | documents the deliberate exception |

## 2. JWKS was fetched over the public internet

`https://auth.jomcgi.dev/...` from a pod hairpins out to Cloudflare and back through the tunnel, so a WAN blip would 503 the entire `/mcp` surface, held for the full `jwksCacheTtlSeconds` by the negative cache. `authentik-server` is a ClusterIP with no such dependency.

Verified in-cluster: `http://authentik-server.authentik.svc.cluster.local/application/o/mcp-friends/jwks/` returns 200 and serves the **same `kid`** that signed a real issued access token. Nothing blocks the path (no clusterwide policy, no netpol in `authentik`, monolith has no egress policy).

`jwksUrl` and `issuer` now deliberately differ, and the values file explains why so nobody "fixes" it: `issuer` is a **string compared** against `iss` and must stay what authentik mints (`issuer_mode: per_provider`); `jwksUrl` is a **URL fetched** and is not required to match.

Plaintext because authentik serves its ClusterIP with a self-signed certificate, so https would mean disabling verification. JWKS is public key material, so integrity is the concern, and it now rests on the cluster network rather than TLS: a strictly smaller attacker set than before.

## 3. PyJWT was never declared

It reached the build only because `mcp` and `semgrep` pulled it in; no `.in` named it. An upstream dropping it would have broken token verification through a dependency this repo never asked for. Now `pyjwt[crypto]>=2.13.0` in `pyproject.toml`. The `[crypto]` extra is required or RS256 has no RSA backend at runtime. Lock regen is a single provenance line, no version churn.

## Not addressed

- No test asserts the `/mcp` mount is actually wrapped by `PrincipalMiddleware`. The wiring is correct today, but a refactor could unwrap it with every test green.
- `projects/mcp/ARCHITECTURE.md` still says token forwarding "is inert until the monolith validates the token" and that `aud` is the client id rather than the monolith. Both were superseded.

Refs #4955, #4942, #4944, #4569

### DIFF
--- a/bazel/requirements/runtime.txt
+++ b/bazel/requirements/runtime.txt
@@ -3086,6 +3086,7 @@ pyjwt[crypto]==2.13.0 \
     --hash=sha256:66adcc2aff09b3f1bbd95fc1e1577df8ac8723c978552fd43304c8a290ac5728
     # via
     #   --override bazel/requirements/overrides.txt
+    #   homelab (pyproject.toml)
     #   mcp
 pyobjc-core==10.3.2 ; sys_platform == 'darwin' \
     --hash=sha256:16644a92fb9661de841ba6115e5354db06a1d193a5e239046e840013c7b3874d \
--- a/projects/monolith/auth/verifier.py
+++ b/projects/monolith/auth/verifier.py
@@ -41,9 +41,44 @@ class AuthentikStandingVerifier:
         )
 
     async def verify(self, token: str) -> Principal | None:
+        # Deliberately ahead of the ownership gate, and the only check that is.
+        # An unconfigured verifier cannot know what it owns, so declining would
+        # turn a deployment mistake into "no verifier recognized this token",
+        # which renders 401 and hides an operator fault behind a caller one.
         if not self._settings.identity_is_configured:
             raise AuthError(AuthErrorReason.UNCONFIGURED)
 
+        # ── Ownership gate ───────────────────────────────────────────────────
+        # This runs BEFORE any check that raises, which is the load-bearing
+        # ordering. A verifier that raises on a token it does not own makes its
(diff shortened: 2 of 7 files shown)


jomcgi:
## Helm Manifest Diff

PR #4965, opened 2026-08-16
feat(agents): give the walkthrough its own full-page view

jomcgi:
Replaces the `prominent` disclosure from #4957. Diff capture works now, so a finished session has a real explanation worth reading, and it was rendering as a bordered panel inside the transcript scroll.

## What changed

A header toggle switches the main pane between **Conversation** and **Walkthrough**:

```
SESSION  vm asleep      [ Conversation | Walkthrough ]
```

- Defaults to Walkthrough when the VM is not awake and at least one turn has a walkthrough; a live session opens on the conversation.
- The composer stays mounted in both views, so a finished session can still be sent a follow-up.
- No route change and no URL state, so no `pushState` and none of the shallow-routing trap where `page.url` never reassigns.

The walkthrough view drops the two-pane split for one column reading top to bottom: a summary sentence, then per file its path and counts, the agent's account, and its diff hunks. One section per turn that has a walkthrough, in order.

## The four defects, all fixed

Visible in the previous panel:

1. **Punctuation** rendered as `Git diff: 1 file changed, +1/-0· Agent accounted for: 1 file· Unexplained: 0 files·` because a separator was appended after every clause including the last. It is a composed sentence now.
2. **Counts printed twice**, in the panel header and again in the points list. Once now.
3. **Diff truncated mid-word** with no affordance. Hunks scroll in their own `overflow-x: auto` container; the page body never scrolls horizontally.
4. **Nested frames**, a heavy border wrapping a panel containing another bordered box. In a full-page view the walkthrough is the page.

## Replaces rather than layers

`prominent` is removed entirely; two mechanisms for making a walkthrough visible would drift apart. In conversation view it is a collapsed disclosure per turn again, as before #4957.

The `onMount` one-shot load from #4957 **stays**. A view that renders open by definition must still fetch, and without it the page would render permanently empty with every test green.

## Reverted implementer churn worth naming

My acceptance criterion said `rg 'html\[data-'` must return nothing. I meant "do not introduce *unscoped* `html[data-*]` in a component style block", which is the real trap: the Svelte compiler silently deletes those. The implementer satisfied the literal grep by rewriting 15 already-`:global()`-wrapped selectors in `+page.svelte` and 3 in `agents-theme.css`, a plain CSS file where the trap does not apply.

That is not cosmetic. `:root[data-x]` has higher specificity than `html[data-x]`, so it changes cascade resolution, and it landed on the night/dawn/dusk palette from #4683. All 18 lines reverted; this PR contains zero theme churn.

## Verified

- `ci lint` clean.
- `ci test //projects/monolith/frontend:test //projects/monolith/frontend:build_test -- --nocache_test_results` -> `Executed 2 out of 2 tests: 2 tests pass`. Uncached deliberately.
- All 138 `P.labels` / `P.punct` keys used across `WalkthroughNarrative.svelte`, `SessionWalkthrough.svelte` and `+page.svelte` resolve against the lexicon. A missing key renders `undefined` and compiles clean.
- No unscoped `html[data-*]` introduced; `prominent` returns no hits anywhere in the agents route.
- Tests cover both default-view cases and the toggle.

## Known gap

`turnHasWalkthrough` decides the default view from parsed rationale, a `base_sha`+`commit_sha` pair, or activities. It cannot see a **stored diff**, because the session detail payload does not expose `diff_base_sha`. A turn with a real captured diff and no RATIONALE trailer would therefore default to Conversation. Session 198 is unaffected since its rationale parses. Worth a follow-up to expose that flag rather than widening this change.

## Not verified

How it looks. The payload is correct, the tokens resolve, and the tests cover behaviour, but the narrative layout is a judgment call that needs eyes on a real session.

### DIFF
--- a/projects/monolith/frontend/src/routes/private/agents/+page.svelte
+++ b/projects/monolith/frontend/src/routes/private/agents/+page.svelte
@@ -33,6 +33,13 @@
   } from "./workspace-recovery.js";
   import PaneHeader from "./PaneHeader.svelte";
   import SessionWalkthrough from "./SessionWalkthrough.svelte";
+  import WalkthroughNarrative from "./WalkthroughNarrative.svelte";
+  import {
+    defaultSessionView,
+    SESSION_VIEW_CONVERSATION,
+    SESSION_VIEW_WALKTHROUGH,
+    walkthroughTurns,
+  } from "./session-view.js";
   import { periodForHour } from "$lib/private/period.js";
 
   const MOBILE_MEDIA_QUERY = "(max-width: 760px)";
@@ -89,6 +96,8 @@
   let runDetail = $state(null);
   let runRequestSequence = 0;
   let detail = $state(null);
+  let sessionView = $state(SESSION_VIEW_CONVERSATION);
+  let sessionViewInitializedFor = $state(null);
   let searchQuery = $state("");
   let searchResults = $state(null);
   let searchLoading = $state(false);
@@ -145,6 +154,7 @@
   let requestSequence = 0;
   let renderedPending = $state({});
   let vms = $state({});
+  let vmSnapshotReceived = $state(false);
   let vmStreamStatus = $state({
     mode: "connecting",
     lastUpdateAt: null,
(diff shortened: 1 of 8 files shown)


jomcgi:
## Helm Manifest Diff

PR #4966, opened 2026-08-16
docs(mcp): correct the token-validation state the monolith now provides

jomcgi:
`projects/mcp/ARCHITECTURE.md` is the source of truth for current state, and two of its claims stopped being true. One of them actively argued against the code that ships.

## What was wrong

| Claim | Verdict |
|---|---|
| "This is **inert** until the monolith validates the token" | False since #4955 |
| "The token's `aud` is the MCP client id **rather than** the monolith, so the monolith must validate `iss` strictly and knowingly accept that audience" | False since `c341f0118` |
| "Per-caller data scoping does not work today" | **Still true**, kept |

The audience claim is the one that mattered. As written it told a future reader the monolith cannot check audience and must accept somebody else's. The shipped code checks membership, and `mcp-auth.yaml` binds a scope mapping returning `{"aud": [provider.client_id, "https://private.jomcgi.dev"]}`. Verified against a real issued access token: `aud` is a list carrying both.

## The part that needed care

The conclusion stayed true while both of its premises went false. A pass looking only for wrong statements would confirm "scoping doesn't work" is still accurate and leave the false premises in place.

So the conclusion stays and the reasons are replaced:

- **No token has been observed arriving** on the claude.ai path, so the principal is anonymous and there is no identity to scope by. Measured, not assumed: on a freshly rolled pod with a cold JWKS cache, an MCP call produced zero monolith to authentik flows, and a first verification would have forced a fetch. The doc notes that this signal expires once the cache warms, so nobody reads later silence as proof.
- **On SSE the principal belongs to the stream, not the message**, already recorded on #4569 with the mechanism.

## Also

Regenerates `repo_docs_manifest.ndjson`, which embeds this file's text and hash. Without it CI's Format stage would auto-commit the drift as `style: auto-format`.

Docs-only, no behaviour change.

Refs #4955, #4569


jomcgi:
## Helm Manifest Diff

PR #4967, opened 2026-08-16
feat(agents): strip the walkthrough to the explanation and its diff

jomcgi:
Follows #4965. The walkthrough is **not a pull request**: it is the model saying it wants to explain a change, with the reason and the evidence. Per-file review belongs on a PR. Everything else on the page was bookkeeping *about* the explanation rather than the explanation.

## Removed

- `TURN N` on single-turn sessions (kept when there is more than one, where it is the only way to tell sections apart)
- The counts sentence: *"1 file changed with 1 insertion and 0 deletions; the agent accounted for 1 file, leaving 0 files unexplained"*
- The `Agent-authored` provenance banner
- The `AGENT'S ACCOUNT · TURN 1 · ATTEMPT 1 · SONNET` chrome. The session header already names the model; attempt numbers are operator trivia.

## Leads with

The model's prose, at reading size with a 760px measure, **before** the path and the diff. It is the one thing the page exists to show and it was rendered as a quoted aside.

```
added a one-line comment above parse_unified_diff clarifying
it parses guest-captured diff text and never shells out to git.

projects/monolith/swarm/unified_diff.py                +1 -0

@@ -32,6 +32,7 @@
+ # Parses diff text already captured in the guest; ...
  def parse_unified_diff(diff: str) -> list[dict]:
```

## The honesty property, moved rather than dropped

The banner existed so a reader could not mistake a claim for verified fact. That guarantee now lives in the layout: prose reads as prose in the model's voice, the diff is monospace with its own background and `+`/`-` colouring, unmistakably machine output.

**Accounting appears only when it is non-zero.** This is the organising rule. "0 files unexplained" is noise that trains a reader to skip the exact region where "2 files the agent never mentioned" would appear. Removing the zero state makes the non-zero state more visible, not less.

When something is unexplained or contradicted, the line names those files in the fact register **and renders their diffs**. That asymmetry is deliberate: a file changed without explanation is the one you most need to see.

## Verified

- `ci lint` clean.
- `ci test //projects/monolith/frontend:test //projects/monolith/frontend:build_test -- --nocache_test_results` -> `Executed 2 out of 2 tests: 2 tests pass`.
- Tests cover single vs multi-turn labels, that nothing renders about unexplained or contradicted when both are zero, that the unexplained line does render and names files when there is at least one, and prose-before-diff ordering.
- All 119 lexicon keys across the three components resolve.
- `--err` measured against every period and background it renders on: 5.48:1 and 4.80:1 day, 5.38:1 and 5.90:1 night. Body text needs 4.5:1, so it clears everywhere. This matters because the console has a live trap here: `--attn` fails at 4.09:1 and `--attn-text` exists for exactly that reason.
- `agents-theme.css` unmodified and zero `:global(` selector rewrites, after a previous pass churned 18 working selectors to satisfy an over-broad grep.

## Not verified

How it reads. Layout is a judgment call and needs eyes on a real session.

### DIFF
--- a/projects/monolith/frontend/src/routes/private/agents/+page.svelte
+++ b/projects/monolith/frontend/src/routes/private/agents/+page.svelte
@@ -1350,7 +1350,6 @@
           <h2>{P.labels.walkSummary}</h2>
           <WalkthroughNarrative
             turnSeq={fixture.walkthrough.turnSeq}
-            model={fixture.walkthrough.model}
             fixture={fixture.walkthrough}
           />
         </div>
@@ -1639,7 +1638,7 @@
               <WalkthroughNarrative
                 sessionId={selectedSession.id}
                 turnSeq={turn.seq}
-                model={turn.model || selectedSession.model || "luna"}
+                walkthroughTurnCount={eligibleWalkthroughTurns.length}
               />
             {:else}
               <div class="empty walkthrough-empty">
--- a/projects/monolith/frontend/src/routes/private/agents/SessionWalkthrough.test.js
+++ b/projects/monolith/frontend/src/routes/private/agents/SessionWalkthrough.test.js
@@ -1,10 +1,62 @@
 // @vitest-environment happy-dom
 import { afterEach, describe, expect, test, vi } from "vitest";
 import { mount, tick, unmount } from "svelte";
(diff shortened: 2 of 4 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4968, opened 2026-08-16
moving: private family move planner at friends.jomcgi.dev/moving, gated to a new authentik group, ADR security/006


jomcgi:
## Status 2026-08-22

PRs 1 to 4 shipped and live: authentik lane and `family` group (#4971, #4990), domain and API (#4972), frontend (#4979, #4988, then polish in #5010, #5018, #5038, #5039, #5040), ingress on `friends.jomcgi.dev/moving` (#4983, `cfIngress.friends` enabled in `projects/monolith/deploy/values.yaml`). ADR security/006 in #4969. The invitation flow was cut from PR 1 (comment below).

Remaining: PR 5, one-way Google Calendar sync. `projects/monolith/moving/models.py` already carries `gcal_state IN ('queued', 'synced', 'held')` on milestones, but no sync job exists (no Google Calendar client under `projects/monolith/moving/`; the only calendar code in the monolith is `chat/whatsapp_calendar.py`). Files: `projects/monolith/moving/` (a `gcal.py` scheduled via `scheduler.api.register_job`), an `OnePasswordItem` for the Google credential in the monolith chart, and a values block to point at the shared "Crossing" calendar. Note all-day events in the Google API use an EXCLUSIVE end date.

Acceptance for PR 5: milestones push once and flip to `synced`; an event moved in Google flips the row to `held` and the UI shows it; nothing ever writes a Google change back into the row without a human choosing.

---

Serve a private move planner ("Crossing") at `https://friends.jomcgi.dev/moving`, gated to a new authentik group `family` (Joe and Anna), for the Vancouver to UK move.

Design reference (approved): https://claude.ai/code/artifact/da48c96d-c44b-44ac-90c7-f0ee0d53f790

Desktop is a shared-border neobrutalist grid. Mobile (<700px) is a tabbed Today / Plan / More app. The mockup is the spec for layout, copy tone, and interaction.

Naming: code, tables, module, and URL are all `moving`. "Crossing" is the masthead word only.

## Why this host

`friends.jomcgi.dev` is already the authentik-audience lane, serving Context Forge's preview app at `/preview/` with group `homelab-admin`. Every other path on that host matches no route and 404s, so `/moving/` is additive. A SecurityPolicy is route-scoped, so a second route with its own policy is how `family` gets a narrower grant than `homelab-admin` without touching the preview lane.

[...]

PR #4969, opened 2026-08-16
docs(adr): record the moving app's authorization lane on friends.jomcgi.dev

jomcgi:
Records ADR security/006, the decision behind #4968.

## Decision

Serve the family move planner (`moving`, displayed as "Crossing") at `friends.jomcgi.dev/moving`, as a second HTTPRoute on that hostname with its own authentik SecurityPolicy, gated to a new `family` group and to no other group.

## Why not `private.jomcgi.dev`

Two reasons, and the second is the load-bearing one.

The Cloudflare Access policy generated by `cf-ingress.security-policy` carries no `authorization` block at all. It authorizes anyone holding a valid `jomcgi` team session and projects `email` into `X-Auth-Email`. There is no group concept on that lane, so a narrower group for one path cannot be expressed there without moving that path's authorization into the Cloudflare dashboard and out of Git.

More importantly, `private.jomcgi.dev` is a shared hostname. Five charts independently render HTTPRoutes on it, each calling `cf-ingress.security-policy` on its own object: monolith, argocd, signoz, longhorn, kargo. A `family` binding anywhere on that host is one policy mistake away from tools that reach the cluster control plane. On `friends.jomcgi.dev` the only other route is `/preview/`, so the same mistake at worst leaks the move planner.

## What this ADR deliberately does not decide

Migrating all of `private.jomcgi.dev` to authentik groups is recorded as deferred, not rejected. It is independently worth doing: it would move private-tier authorization into Git and fix Cloudflare Access consuming the `Authorization` header so no MCP caller carries a real actor. But the ADR names the constraint any such migration has to answer first, which is documented in `httproute-preview.yaml` and is stronger than the migration cost: operator surfaces are on Access deliberately, so that authentik being down cannot lock the operator out of the tools used to fix authentik.

## Verification

Docs only, no chart or template changes. `ci regen` and `ci lint` green; regen output is one new manifest entry plus an `order` renumber of the four docs that follow it. No `Chart.yaml` `version:` or `targetRevision:` touched.

Refs #4968


jomcgi:
## Helm Manifest Diff

PR #4972, opened 2026-08-16
feat(moving): add the move planner domain, schema and API

jomcgi:
PR 2 of #4968. The backend for the family move planner, behind ADR `security/006`.

No frontend, no ingress, no Google Calendar. Those are later PRs. Nothing here is reachable from outside the cluster yet.

## What lands

- **Migration** `20260816000000_moving_schema.sql`: schema `moving` with `tasks`, `milestones`, `spans`, `roles`, `viewers`. TEXT plus CHECK rather than Postgres enum types, because adding an enum value needs an `ALTER TYPE` that does not roll back cleanly in a transaction and these vocabularies will move as the move does.
- **Models** mirroring the migration, with tests asserting they actually match it.
- **Viewer resolution**: a dependency that maps the `X-Auth-Email` header to `joe` or `anna` via `moving.viewers`. Missing header is 403, unknown email is 403, never a default viewer.
- **Router** at `/api/moving`: one `GET /state` for the whole dashboard, task CRUD, and idempotent done/undone.
- **Collisions** as a pure function over loaded rows, so it is testable without a database.

## Two design points worth review

**No email address is in Git.** This repository is public, so the viewer mapping lives in a `moving.viewers` table seeded out of band rather than in `deploy/values.yaml`. Tests use `a@example.test` only. That also makes "unknown email is a 403" a database fact rather than a config convention.

**`/api/moving` is deliberately absent from `httproute-private.yaml`.** Without a rule there, `/api/moving/*` on private.jomcgi.dev falls through to the SvelteKit catch-all and 404s, which is what keeps this data off the operator hostname. The only door will be the friends.jomcgi.dev route in PR 4.

## BUILD wiring, and why it is in this PR

`projects/monolith/BUILD` is hand-maintained, not gazelle-generated: it carries a `# gazelle:exclude` per domain precisely to keep gazelle away. A new domain has to be wired into six places, and adding it to `_BACKEND_SRCS` alone shipped the sources to the binaries while leaving them out of `monolith_backend`, so every test that composes the app failed on `ModuleNotFoundError: No module named 'moving'`.

Deliberately NOT added to `monolith_public_backend` or `main_public`: this domain is private tier only, and adding it there would compile family data into the public binary. That is the one place where doing the consistent-looking thing would be wrong.

## Why the tests live in `moving/tests/`

A domain-root `conftest.py` is not matched by `monolith_backend`'s excludes (`**/*_test.py`, `**/test_*.py`, `*/tests/**/*.py`), so it shipped into the library's runfiles. pytest auto-loaded it in unrelated targets, which imported `moving.models` and registered the schema-qualified tables into the process-global `SQLModel.metadata`. `ember_public_router_test` then failed creating `moving.tasks` on SQLite, and `import_boundaries_test` failed collecting the package without fastapi. Two unrelated-looking failures, one stray file.

`home`, `chat` and `knowledge` already keep fixtures at `<domain>/tests/conftest.py` for this reason. Following them makes the leak structurally impossible rather than merely fixed, at the cost of one `--ignore` line in `bdd_completeness_test`, which is the established pattern for every domain with a `tests/` directory.

## Verification

`ci test`: **Executed 3 out of 725 tests: 725 tests pass.** Reached after three red rounds (23 failures, then 2, then 1), each a single mechanism, all described above. The 26 domain tests also pass directly under pytest. No `Chart.yaml` `version:` or `targetRevision:` touched.

Refs #4968


jomcgi:
## Helm Manifest Diff

jomcgi:
## Review round applied

Every finding from the review is addressed in `47cef30b3`. Two needed a decision from Joe and got one; the rest were applied as found.

### The two blocking findings turned out to be one problem

`viewer` was resolved and then only echoed back. No endpoint filtered on `owner`, and the write endpoints bound it purely to force the 403, so both people could see and edit everything and the Mine/All control in the design had nothing behind it. Separately, `if task.track == span.kind: continue` compared `sell|admin|ship|people` against `visitor|work|move|trip`, disjoint vocabularies, so the guard could never fire.

Both wanted the same missing thing: an `owner` column on `spans` and `roles`.

**Decisions:** per-person scoping is wanted, and a task colliding with any span it is due inside is the rule. The broken guard is deleted rather than repaired, because the docstring described behaviour that never existed. That wording came from the task spec, not the implementation.

### What changed

- `owner` added to `spans` and `roles`. `NOT NULL DEFAULT 'both'` on all four tables: a nullable owner would fall out of BOTH viewers' "mine" scope, which is a silent disappearance rather than a visible error.
- `GET /state?scope=mine|all`, with progress and collisions computed over the same filtered set the response returns. A progress bar counting rows you cannot see is wrong.
- Writes go through one `_editable_task_or_403` helper, so there is a single point to audit. `owner = 'both'` is editable by either person.
- `spans.kind` is now `NOT NULL`. A null kind made two spans compare equal and skip the cross-kind rule.
- `milestones.gcal_event_id` gains `UNIQUE`, before the calendar sync needs it and while there is nothing to deduplicate.
- `models_test` now parses the migration and compares against it, instead of restating hand-transcribed literals that stayed green when the SQL changed.
- A test walks `router.routes` and asserts every one carries `get_viewer`, so a future route added without it fails rather than shipping unauthenticated.
- `X-Auth-Email` binds as `list[str]` and 403s unless exactly one value is present. Envoy appends its projected identity, so a forged inbound value would arrive first. Defence in depth behind the listener strip, not a substitute for it.
- A malformed task id now 404s instead of raising `invalid input syntax for type uuid` under Postgres and surfacing as a 500 with an aborted transaction. The old test pinned the SQLite behaviour and hid the difference.
- `moving` added to `FORBIDDEN_MODULES` and to `import_boundaries_test`'s `DOMAINS`. Both were wired only halfway, so the guards existed but enforced nothing for this domain.

All of this edits the existing migration in place rather than stacking a second one. That is only possible because it is unmerged and the tables are empty, and the window closes on merge.

### Verification

- 41 domain tests pass, up from 26.
- `ci test`: `Executed 62 out of 725 tests: 725 tests pass`, then a second run on the same tree returned `Executed 0 out of 725 tests: 725 tests pass` on cache. Two runs, one executing and one cache-hitting, rather than one ambiguous result.
- `atlas migrate validate` passes and `atlas.sum` was regenerated by the hook, which ran for the first time on this commit because it is the first to touch a migration file.

### Filed separately

#4974, the migrations ConfigMap is at roughly 239,870 bytes against the 256 KiB annotation cap. Nothing here violates the bulk-data rule, but the headroom is worth tracking before the next large migration.

Issue #4970, opened 2026-08-16
embervm: bank or relight optimization for measured workload delay


jomcgi:
Approved worth-review scope (2026-09-14): gate banking/relight optimization on measured delay for a named current workload. Replace needs-human with needs-thought.

Gate: identify a supported workload, measure its bank or relight delay, and explain the user or factory impact worth reducing. Revalidate current behavior and identify the dominant stage before selecting an improvement. Historical nested-virtualization benchmarks, dormant code, a demo workload, or infrastructure prerequisites alone do not satisfy this gate.

When the gate is met:

- Check the workload's current lifecycle and storage contract, pinned implementation, relevant flags and filesystem requirements. UK home metal alongside GKE remains part of the placement direction; do not assume a single-machine-type fleet or an obsolete transition plan.
- Choose one bounded improvement justified by the measured bottleneck. Reuse existing implementation where suitable rather than automatically activating every warmth optimization.
- Preserve snapshot/generation pairing, commit-versus-abort safety, durable workspace data and applicable full-snapshot fallback. Establish boot-time tracking and re-bank requirements if diff banking is selected.
- Measure before/after behavior for the named workload, including correctness and fallback evidence. Define a separately scoped rollout and rollback only if the selected change needs runtime activation.

The previous issue reported dormant diff-banking code from #4973 and a retired sandbox-session allowlist entry. Treat those as historical leads to revalidate, not authority to enable diff banking on demo-postgres or another workload.

Removed from the active deliverable: bundled balloon/cache-drop banking, prefetch, optional idle-signal channel and a broad warmth-pipeline roadmap. These techniques may be considered individually only when the named measurement supports them. No hugepage programme is revived.

This decision does not authorize diff-banking activation, dev/prod flag changes, fleet re-banking or a demo rollout. Earlier unconditional activation steps are superseded by this evidence gate.

Historical context: #4964, #4973 and #5699. Availability of those prerequisites alone is not a benefit case.

jomcgi:
Correction after fact-checking the checklist against the codebase and the #4964 benchmarks. The plan's ordering survives; its motivating arithmetic was built on the wrong workload quadrant.

**The quadrant split the checklist missed:**
[...]

Issue #4964, opened 2026-08-16
platform: UK home primary workloads alongside the always-on GKE hub


jomcgi:
## Operator direction, 2026-09-13

Retain the home-plus-GKE end state. Joe confirmed at least three already-owned home nodes have been shipped to the UK ahead of the move, each with 12+ CPU and 16 GB RAM. Shipping those nodes is complete; UK installation and readiness are not yet established by that confirmation.

Home metal will run primary workloads alongside GKE. GKE remains always on as the more reliable hub. Owned home compute avoids additional cloud compute rental; bare metal is a strong fit for EmberVM. This is not a sole-GKE retirement plan or a requirement to move every workload home.

## Next bounded planning slice

Prepare a concrete placement and UK bring-up plan against current deployed state:

- Record which primary workloads should run at home, which reliability-sensitive services remain on GKE, and the first home workload to prove. Treat EmberVM as a named candidate, not an unvalidated fleet migration.
- Verify usable home resources, disk and virtualization support, provisioning and network availability. Budget capacity after host/platform overhead rather than treating the hardware totals as schedulable capacity.
- Choose the minimum management/enrollment path that keeps GKE as hub and preserves outbound-only home connectivity. The old argocd-agent proposal needs current-state validation; it is not a mandate for a new management framework.
- Define what happens when home power or connectivity fails: what remains available on GKE, which work pauses, and which recovery is manual. Always-on GKE does not by itself establish automatic workload failover.
- Reconcile actual backup/restore and workload-data requirements before moving state. Revalidate snapshot portability for any EmberVM placement crossing hardware boundaries.
- Produce bounded implementation children with explicit acceptance, then select the first ready slice. This parent remains needs-thought rather than admitting the historical migration programme as one task.

## Scope boundaries and related decisions

#5485 concerns remaining Vancouver teardown/inference transition, not permanent retirement of the shipped UK fleet. Node-4/4090 disposition and inference placement remain with #5485/#5461 and require their own evidence and decisions.

The earlier closures of #3855/#4753 and PR #6069 remain in force. Two Kubernetes clusters do not automatically require EmberVM cell partitioning. Revisit only if the chosen ownership topology demonstrates that requirement.

No hardware wipe/disposal, production migration, new cloud provisioning, or automatic cross-cluster failover is authorized by this issue update.

## Historical plan

[...]

PR #4973, opened 2026-08-16
feat(embervm): diff snapshot banking with offline merge and full fallback

jomcgi:
Implements the main event from #4970: session re-banks write only the dirty page set instead of a Full memory image, shrinking the idle-gated bank pause from ~658ms/GiB (measured, #4964 benchmarks) to a dirty-set-sized write, with the merge running after resume so it never sits on the pause-critical path.

## What
- `track_dirty_pages` added to MachineConfig, armed at boot for banking classes and re-armed via `enable_diff_snapshots` on every restore
- Bank sequencing per lineage: Full first, then Diff; merges are strictly sequential (Firecracker's dirty bitmap resets at each create)
- `snapshot-editor` rebases each diff onto the previous full, atomically (merge temp + rename, previous full valid until replacement committed); the zstd CAS continues to store fulls only
- Hard fallback rule: any diff-path failure (create, missing base, merge error, editor absent) logs and takes a Full snapshot; diff banking cannot brick banking
- Rollback knob `noded.diffBanking` (default true), plumbed code + chart template + values together and verified by `helm template`
- `snapshot-editor` is vendored from the same upstream Firecracker v1.16.1 release as the hypervisor with a build-time version-affinity check (the kata bundle stays on its pinned v1.12.1)

## Verification
- `ci lint` clean; `ci test`: `Executed 2 out of 721 tests: 721 tests pass.` (mostly cache-hit on the implementer's earlier green run)
- New driver tests: full-then-diff sequencing, diff re-arm on load for banking classes only, fallback-to-Full on merge failure, atomic-rename crash safety
- No Chart.yaml version or targetRevision changes (post-merge write-back owns those)

Refs #4970, #4964 (benchmark and correction comments)


jomcgi:
## Helm Manifest Diff

PR #4979, opened 2026-08-16
feat(moving): add the desktop core of the move planner page

jomcgi:
PR 3a of #4968. The desktop core of the planner at `/moving` on friends.jomcgi.dev.

The gantt timeline, the dock modals (To sell / Calendar / Roles) and the mobile Today/Plan/More tabs are deliberately **not** here. They are PR 3b. Nothing routes to `/moving` until PR 4, so this merges safely while incomplete.

## The load-bearing change is in BUILD, not in the page

`:src_public` drops private pages, their SSR handlers and their client chunks from the public build entirely, per ADR 004 Phase 5d code-absence. It excludes `src/routes/private/**` and `src/lib/private/**`. A new `src/routes/friends/**` tree matches neither, so without an entry this family app would compile into the public jomcgi.dev bundle: unreachable, but sitting there readable in devtools.

`src/build-exclusions.test.js` enforces it. It parses the `src_public` block specifically rather than grepping the whole file, because a match in the wrong glob (`:src`, where the pattern must NOT appear) would pass a naive grep. The guard was verified by deleting the exclusion, watching the test fail, and restoring it.

## What is on the page

Masthead, a hero with the countdown to the earliest `move` span plus overall progress, a Do Today list with the Mine/All scope toggle, and the collisions panel. Cells sit on an ink ground with a consistent gap so the ground reads as the border.

Empty state got deliberate attention, because it is what production shows first: the tables are empty, so no NaN, no divide-by-zero in progress, and "no move date set" rather than a countdown to nothing.

## Three traps this repo has hit before, avoided on purpose

- **The scope toggle navigates with `goto` plus `keepFocus` and `noScroll`.** `pushState`/`replaceState` never reassign `page.url`, so URL-derived `$derived` state would not recompute. The signature of that bug is deep links working while clicks do nothing.
- **Theme rules live in `moving-theme.css`, not a component `<style>` block.** The Svelte compiler silently comments out unscoped `html[data-*]` selectors while `build_test` stays green, which would ship dark mode as a no-op.
- **A 403 is a rendered panel, not an error page.** `moving.viewers` starting empty means 403 is the expected first-run response, so it had to be a designed state rather than a crash.

## Verification

- `//projects/monolith/frontend:test` and `:build_test` forced with `--nocache_test_results`: **Executed 2 out of 2 tests: 2 tests pass.** Forced deliberately, because the earlier full run reported `Executed 0 out of 725` on a warm cache, and `:test` alone does not compile the app. Only `:build_test` catches a Svelte template error.
- 24 vitest cases across the helpers, the access panel and the server loader.
- `ci lint` green. No `Chart.yaml` `version:` or `targetRevision:` touched.

Refs #4968

### DIFF
--- a/projects/monolith/frontend/BUILD
+++ b/projects/monolith/frontend/BUILD
@@ -223,9 +223,9 @@ vitest_bin.vitest_test(
     args = ["run"],
     chdir = package_name(),
     data = [
+        "BUILD",
         ":src",
         ":test_lib",
-        "BUILD",
     ],
 )
 
--- a/projects/monolith/frontend/BUILD
+++ b/projects/monolith/frontend/BUILD
@@ -82,10 +82,11 @@ js_run_binary(
 )
 
 # Public-only source tree (ADR 004 Phase 5d code-absence): identical to :src but
-# additionally excludes the authenticated route tree (src/routes/private/**) and
-# private-only components (src/lib/private/**). SvelteKit is filesystem-routed,
-# so omitting these files from the source set drops those pages, their SSR
-# handlers, and their client chunks entirely from the build.
+# additionally excludes both authenticated route trees and private-only
+# components. SvelteKit is filesystem-routed, so omitting these files from the
+# source set drops those pages, their SSR handlers, and their client chunks
+# entirely from the public build. Friends routes have their own authorization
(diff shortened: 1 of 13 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4976, opened 2026-08-16
embervm: platform-managed KEK custody via derivation in the control-plane key service, ADR embervm/035


jomcgi:
Implements the custody decision in `docs/decisions/embervm/035-platform-kek-custody-derived-control-plane.md` (Accepted), which resolves ADR embervm/033's open question 1. Parent tracking for the wider encryption-at-rest work is #4691; the principal-model boundary rules this respects are #4940.

## Decision being implemented

Platform-managed principal KEKs are derived on demand as `HKDF(K_root, principal_id || epoch)` inside the EmberVM control plane's key service (the same component ADR 028 decision 8 names for Account chunk keys). 1Password holds exactly one root item regardless of principal count (the repo carries 33 `kind: OnePasswordItem` declarations today; this adds one). Nothing here touches the customer-managed mode.

## Work items

- **Root provisioning**: one `OnePasswordItem` for `K_root` delivered to the key service; held in memory for derivation, never written to the datastore or logs. Support a second concurrent generation during root rotation (item count 2 only during the transition window).
- **Derivation**: HKDF with hash choice and domain-separation labels to be fixed here (ADR 035 open question 1), including whether the per-CPU-vendor artifact keying axis folds into the info string or stays at the data-key layer.
- **Epoch facts**: per-principal current epoch and minimum-accepted epoch as control-plane facts under the op-log discipline (durable-before-observed, invariant 7). The floor raise is an op-log event so a lost floor is recoverable from the journal.
- **Wrapped data-key format**: record `(principal, epoch)` alongside each wrapped data key so unwrap needs only the root plus the recorded epoch and never depends on floor state.
- **Revocation path**: refuse derivation below the minimum-accepted epoch; lazy rewrap on access plus a background sweep so the floor can actually rise; revocation is complete when the floor rises, and that event is the auditable fact.
- **Mode switch**: rewrap a principal's data keys between the derived KEK and a customer KMS key (both directions); artifacts are not re-encrypted.
- **Observability**: counters for derivations, floor raises, rewrap progress, and refused below-floor derivations.
- **Integration**: capability issuance stays the 033 decision 3 tuple check; this issue supplies the KEK the unwrap consumes. Coordinate scope with #4691 rather than duplicating the restore-path work tracked there.

## Non-goals (ruled out by the ADR)

- No per-principal `OnePasswordItem` resources, ever; vault footprint stays O(1).
[...]

Issue #4940, opened 2026-08-15
monolith: auth domain, principal model, and delegated workload identity


jomcgi:
## Approved disposition: current auth documentation and gated expansion

This issue is a needs-thought tracker, not an implementation brief for a new identity/delegation programme. Existing authentication and authorization remain required.

## Current mechanisms

The historical implementation record names #4942 and later corrections as landing the auth domain, principal representation and verifier/middleware integration. #4941 now verifies and documents the current trust boundary; do not rely on the obsolete statement that no monolith authentication exists, or treat proposed delegation behavior as already shipped.

## Remaining work and gates

- #4941: agent-ready, concise documentation of current principal verification, resource authorization and actual anonymous/invalid-token behavior. Explicitly distinguish implementation from proposals.
- #4943: needs-thought, only a named caller with an authentication gap can justify a new standing workload identity.
- #4944 and #4946: needs-thought, first identify a named agent operation requiring per-issue permissions the existing factory path cannot enforce, then choose the smallest correction. Neither proposal is justification for the other.
- #4945: needs-thought, proceed only if the selected delegation design requires a new credential delivery path.

The retired merge-queue reconciler and hypothetical future tenants do not satisfy these gates. Executor role names do not themselves grant authority or require separate OAuth clients.

## Independent security work

#4569's caller/result scoping and #5789's existing MCP control enforcement retain their own scope and readiness. They do not wait for this optional expansion. Preserve the security-finding classification on this tracker; gating proposals is not a claim that outstanding security findings are resolved.

No blanket restriction on existing supported GitHub access, new credentials or runtime policy change is authorized by this triage decision.

The earlier unconditional rollout order and settled future-design brief are superseded. Historical design details remain in issue history; gated children retain the relevant constraints if their evidence gates are met.

jomcgi:
## Triage recommendation

**Recommendation:** Keep as tracking parent

[...]

PR #4980, opened 2026-08-16
docs(adr): resolve embervm/033 open question 1, platform KEK custody by derivation

jomcgi:
Closes ADR embervm/033's Open Question 1 (KEK custody for the platform-managed mode). Tracked by #4976.

## What this decides

ADR embervm/036: platform-managed principal KEKs are **derived**, `HKDF(K_root, principal_id || epoch)`, from a single root held by the EmberVM control plane's key service (the component ADR 028 decision 8 already names). 1Password holds exactly one item regardless of principal count.

Neither of the two candidates 033 named won:

- **Per-principal 1Password items**: fails the operator constraint. Secret count would grow with principal count against a personal vault; there are 33 `kind: OnePasswordItem` declarations today.
- **Class-3 swap-tier mirror (ADR 016)**: class mismatch. The KEK is class 1 under ADR 016's own taxonomy (we are the validator, the material is derivable and rotatable), and the tier's defining jobs (per-key token buckets against provider rate limits, fleet-shared fixed keys) have no analogue here. The oracle property is kept without the tier: the key never leaves custody, the operation travels to it.
- **authentik as key store** (asked for explicitly, evaluated, rejected): workload identities are declared in blueprints, which are GitOps files in this repo, so key material declared there lands in git. Authentik's only key primitive in-tree is an OIDC `signing_key`; there is no wrap/unwrap or data-key API. It would also invert #4940's boundary and place the IdP on the restore path.

## What it deliberately does not re-decide

Per-principal envelope encryption (033), the immutable-vs-mutable plane split and Account-scoped convergent encryption for rootfs chunks (028), and the existing prohibition on cross-principal content-addressed dedup (ARCHITECTURE.md section 5, invariant 3).

## Honest consequences, stated in the ADR

- Derivation does **not** improve the compromise bound: reading the root derives every KEK, exactly as reading a stored keyring holds every KEK. It removes the storage and vault-scaling problems only. This is the same bound 028 already accepted pre-EKS for the immutable plane, and both planes exit it together at the KMS swap.
- **Revocation becomes an epoch floor, not key deletion.** A derived key cannot be deleted, so a compromised principal KEK is retired by rewrapping under a new epoch and raising the custodian's minimum-accepted epoch. The floor is the fact that must be durable.
- Custody in the control plane is a real blast-radius expansion: the CP does not see guest memory today. Bounded by keys and ciphertext living in different components (the CP never touches the object store).

## Also in this PR

- The `workload-claude-runtime.yaml` comment said "deploy keeps that flip off" for `persistenceEnabled`, but `deploy/values.yaml` sets it **true**. It also described the class as memory-banking when idle, which is the branch the chart default renders, not the one production runs. `do_bank` (`session_manager.ex`) selects on that flag: persistence ON parks, OFF memory-banks, and both are reached through a knob named `idleBankSeconds`.
- Docs manifests regenerated by `ci regen` (036 inserts at order 138 and cascades `order` for later docs).

## Verification

`ci` green on the final tree: `Executed 6 out of 725 tests: 725 tests pass`.

Refs #4940, #4976, ADR embervm/028, ADR embervm/033.


jomcgi:
## Helm Manifest Diff

Issue #4773, opened 2026-08-13
embervm: resource measurements for a concrete sizing or diagnostic need


jomcgi:
Approved worth-review scope (2026-09-14): retain behind a concrete resource-measurement need. Keep needs-thought and remove agent-ready.

Gate: name a sizing or incident-diagnosis decision that existing measurements cannot answer, with evidence of the missing information. Examples include distinguishing guest pressure from daemon overhead for UK/factory nodes. Historical measurements in this issue are leads, not current fleet measurements or approved constants. Availability of a metrics receiver alone does not satisfy the gate.

When the gate is met:

- Revalidate current runtime, resource accounting, existing measurements and consumers. Start with bounded measurements through existing facilities sufficient to answer the named question.
- Distinguish current RSS, peak RSS, declared guest memory, cgroup charges and shared cache. Do not report subtraction of guest RSS and cache from cgroup usage as exact daemon memory without establishing that the quantities are compatible and account for sharing and overlap.
- Add recurring telemetry only if the decision requires it and existing facilities cannot supply it. Keep collection overhead, cardinality and failures bounded; state attribution limits and unknown values explicitly.
- Report the evidence and smallest proposed correction, if any. No arbitrary two-week collection requirement: justify the sample's representativeness for the decision.

Measurement does not authorize reserve, capacity or admission-policy changes. Separately scope and validate any later tuning against current node classes and all accounting consumers; do not reuse historical constants or silently change safety behavior. #5851/#5859 can supply a measured factory bottleneck, but their completion alone does not satisfy this gate.

Out of scope without a separately justified decision: a new exporter or telemetry platform, per-VM cgroup changes, a reserved-admission trial, and automatic daemonReserveMib splitting. Preserve existing useful measurements and admission safeguards.

Historical context: #4762, #4772 and ADR embervm/001, 016, 021. Earlier mandatory sampler and downstream tuning sequences are superseded by this gate. Any proposed dependency on this sampler must establish that existing measurements are insufficient for its own named decision.

jomcgi:
## Agent brief

### Outcome
agent-ready: phase 1 (the per-VM sampler ticker plus OTLP gauges) is fully specified with no human decision left, and phase 2 (the reserve split) is explicitly gated on about two weeks of sampler data, so an agent can deliver phase 1 now without resolving phase 2 constants.

### Acceptance
- Generalise Stats() from once-at-release to a 15-30s ticker over the live VM map, fail-soft on /proc read errors.
[...]

PR #4982, opened 2026-08-16
perf(embervm): cut the primed-pool floors and the brick daemon reserve

jomcgi:
## What

Two independent memory changes to EmberVM, both values-only.

**Primed-pool floors.** The floor is applied PER BRICK INSTANCE (`PoolManager.refill_node/2` runs once per capacity fact), so every nonzero floor is a multiplier over the fleet, and the fleet is now six bricks.

| Workload | floor | memMib | change |
|---|---|---|---|
| sandbox | 2 -> **0** | 512 | 1 GiB/brick of primed VMs for a workload whose demand is the `*/5` probe plus visitor traffic |
| sandbox-session | 1 -> **0** | 2048 | 0 live / 0 banked in the 31 days since it was created |
| semgrep | 4 -> **1** | 1536 | a 6 GiB ask per brick that a 2gi brick can never satisfy |
| bazel-query | 1 (unchanged) | 1024 | its warm base *is* the Skyframe exhibit |
| claude-runtime | 0 (unchanged) | 4096 | already deliberately 0 |

**Brick daemon reserve**, 512 -> 256 MiB.

## Measurement

Live fleet, 2026-08-16, from the CP's `CapacityObserver` and `kubectl top`:

- 8 primed VMs holding **8192 MiB of reservation against ~510 MiB of actual working set**
- node-1/2/3 at **74% / 77% / 87%** node memory; node-4 at 29%
- idle bricks' `noded` container RSS: **9 MiB** (node-1), **11 MiB** (node-2)

## Why no memMib was reduced

Each is defensible against its own recorded rationale (bazel-query's 1024 is a measurement from #4062; semgrep's 1536 is sized for the rule engine's peak). There is no per-VM memory gauge to size them against, which is #4773, so cutting one blind trades reservation for an OOM inside a guest that surfaces as a red `*/5` probe rather than as a saving. The reservation problem was the floor multiplier, not the sizes.

## The daemon reserve is not a footprint cut

The brick pod requests its full class limit from the node either way. This is a capacity *unlock* inside a reservation already paid for: +256 MiB guest-schedulable per brick, moving 1gi from 512 to 768 MiB usable.

It stops at 256 rather than lower on purpose. The field is documented as covering the daemon's own RSS (9-11 MiB measured), but the admission predicate reads `memory.max - memory.current` and **cgroup v2 counts page cache**, so a brick that has baked a rootfs carries hundreds of MiB of reclaimable file cache charged to its own cgroup (node-3's brick: **1391 MiB with zero live VMs**). The 512 has silently been paying for that and for per-VM VMM overhead; `VMOverheadMib` is the field meant to carry the VMM half and defaults to 0. Splitting it into measured parts is #4773.

Because the observed admission model re-checks live headroom at boot, undersizing shows up as `pressure:mem` boot rejections and evicted snapshot cache (slower restores), not as an OOM kill.

## Rendered effect

`EMBERVM_BRICK_CLASSES` usable_mib: 1gi 512 -> 768, 2gi 1536 -> 1792, 4gi 3584 -> 3840, 8gi 7680 -> 7936, 16gi 15872 -> 16128.

## Verification

- `ci test`: `Executed 3 out of 725 tests: 725 tests pass`
- `helm template` against `deploy/values.yaml` confirms every rendered floor, memMib and the class table above

## Rollback

Each value independently: floors back to 2/1/4, `daemonReserveMib` back to 512. No code change, no schema change.

### DIFF
--- a/projects/embervm/chart/values.yaml
+++ b/projects/embervm/chart/values.yaml
@@ -450,10 +450,23 @@ sandboxWorkload:
   # it worse, not better: because the floor is per instance, each new brick
   # arrived wanting its own 4.
   #
-  # 2 leaves at least one free slot on the smallest brick for real demand. This
-  # is a STOPGAP for the per-instance semantics; once floor is a fleet-wide pool
-  # size it should go back up (issue #4077 follow-up).
-  floor: 2
+  # 0, down from 2, because the per-instance semantics above make any nonzero
+  # floor a MULTIPLIER over the fleet, and the fleet is now six bricks: floor 2
+  # at 512 MiB was 6 GiB of primed VMs standing by for a workload whose entire
+  # demand is the */5 synthetic probe plus occasional visitor traffic on the
+  # public demo. Measured 2026-08-16 across the live fleet: 8 primed VMs holding
+  # 8192 MiB of reservation against ~510 MiB of actual working set.
+  #
+  # A floor of 0 does NOT mean a cold boot per request. The pool primes on demand
+  # from the workload's base snapshot, which is the same demand-paged restore a
+  # primed VM was created by, so a miss costs a restore (~250ms cold spawn worst
+  # case) rather than a build. That is the same trade claudeRuntimeWorkload
+  # already takes deliberately at floor 0, and the same one bazelQueryWorkload's
+  # floor comment describes as cheap.
+  #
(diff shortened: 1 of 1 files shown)


jomcgi:
## Helm Manifest Diff

PR #4983, opened 2026-08-16
feat(moving): route friends.jomcgi.dev/moving behind the family group

jomcgi:
PR 4 of #4968, implementing ADR `security/006`. **This is the PR that opens the door.**

Adds one HTTPRoute, one SecurityPolicy targeting it, and the client-secret OnePasswordItem.

## Routes

| Path | Backend | Why |
| --- | --- | --- |
| `/moving` | SvelteKit `:3000` | The page. The SvelteKit reroute hook maps it to `/friends/moving` internally, so no gateway rewrite. |
| `/api/moving` | FastAPI `:8000` | Falling through to `:3000` returns the SPA shell, which the browser reports as `Unexpected token '<'`. Nothing in CI exercises a browser-originated path against real ingress, so that class of gap ships green. |
| `/_app/` | SvelteKit `:3000` | Inside this route deliberately, so the policy covers it. Serving the bundle unauthenticated would leak the page to anyone who guessed the path. |

All three in one route so one SecurityPolicy covers them. The OAuth callback and logout paths fall inside `/moving`, because Envoy serves them itself but the request must select the route first or it 404s before the policy is consulted.

## The gate

OIDC redirect against authentik, JWT validated against its JWKS, `authorization.defaultAction: Deny`, and a single Allow rule matching the `groups` **claim** against `family`. A claim inside a signature Envoy just verified cannot be forged; the projected `X-Auth-Email` header answers only "whose view", and is trustworthy because the gateway-wide ClientTrafficPolicy strips inbound copies at the listener.

Cookies are `moving-access-token` / `moving-id-token`, distinct from the preview lane's `preview-*` on the same hostname. Shared names would shadow each other's sessions and present as an infinite login loop rather than a clean rejection.

## Why a second route rather than rules on an existing one

A SecurityPolicy targets a whole HTTPRoute, so two authorization models on one hostname require two routes. `friends.jomcgi.dev` already carries `/preview/` gated to `homelab-admin` from the **mcp** namespace; this carries `/moving` gated to `family` from **monolith**. The listener sets `allowedRoutes.namespaces.from: All` and the prefixes are disjoint.

## The dev collision this avoids

Dev's Application loads `projects/monolith/deploy/values.yaml` **and then** `dev/deploy/values.yaml`, so production's `enabled: true` arrives in dev inherited. Left alone, the dev release would render a second HTTPRoute claiming this hostname on the shared Gateway, a second SecurityPolicy, and a `moving-oidc-client` OnePasswordItem in `monolith-dev`. Explicitly disabled in dev's overlay, the same gating `shipsRedirect` and `githubWebhook` already carry.

Verified both ways: the dev release renders nothing from this template, production renders all three objects.

## Verification

`ci test`: **Executed 2 out of 725 tests: 725 tests pass.** `helm template` checked for the production release and the dev release separately. No `Chart.yaml` `version:` or `targetRevision:` touched.

**Post-deploy checks, against the live URL rather than the manifest.** A SecurityPolicy is a separate object from the route it targets, so "the policy denies" and "the policy is attached" are different facts, and `defaultAction: Deny` cannot protect a route the policy never bound to:

1. Logged out, `https://friends.jomcgi.dev/moving` redirects to authentik. Never a 200.
2. Authenticated but not in `family`, 403.
3. In `family`, the page renders and `/api/moving/state` returns JSON rather than the SPA shell.
4. `https://friends.jomcgi.dev/preview/` still works, and every other path on the hostname still 404s.

Refs #4968

### DIFF
--- a/projects/monolith/chart/templates/httproute-friends.yaml
+++ b/projects/monolith/chart/templates/httproute-friends.yaml
@@ -1,0 +1,200 @@
+{{- if .Values.cfIngress.friends.enabled }}
+{{- /*
+The moving planner's lane on friends.jomcgi.dev (ADR security/006).
+
+WHY THIS IS A SEPARATE ROUTE rather than rules added to the preview lane's
+route, or to httproute-private.yaml.
+
+A SecurityPolicy targets a WHOLE HTTPRoute, so two authorization models on one
+hostname require two routes. friends.jomcgi.dev already carries /preview/,
+gated to homelab-admin by context-forge-gateway's own route and policy. This
+app answers to a different group entirely, `family`, which contains one person
+who is not an operator. Path isolation is how the two coexist: /preview/ keeps
+its policy untouched, /moving gets its own, and neither can swallow the other.
+
+That route comes from the mcp namespace while this one comes from monolith.
+The shared Gateway listener sets allowedRoutes.namespaces.from: All and the
+two path prefixes are disjoint, so they coexist without conflict.
+
+WHY NOT ON private.jomcgi.dev. The Cloudflare Access policy that guards that
+hostname carries no `authorization` block at all: it authorizes anyone holding
+a valid team session. There is no group concept on that lane, so a narrower
+grant cannot be expressed there. It is also a hostname shared by argocd,
+signoz, longhorn and kargo, so a mistaken binding there sits one claim away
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4977, opened 2026-08-16
agents: voice companion surface and voice_ui_* MCP contract (design handoff)


jomcgi:
**Interactive mockup for review**: https://claude.ai/code/artifact/4f80b1c5-a96c-4ddf-8b6f-49d5f2b72800 (rev 2, 2026-08-15). Press "play the conversation" or step with the arrows. The decay engine and the pin control are live in the demo rather than staged frames, so the lifecycle rules below can be watched, not just read. The artifact is account-private; this issue body carries the full contract so an implementing session does not depend on fetching it.

## Concept

A parallel screen that a voice session grabs over MCP. The voice model holds the remote: it conjures console surfaces (run DAG, walkthrough, decision gates) onto a stage keyed to what the conversation is about right now, runs session tools out loud, and the stage holds only what is currently relevant. Same material as `/private/agents`, one new idea: the screen follows the talk. Companion closed means every call no-ops; nothing in the voice flow may depend on the screen being up.

## Proposed contract

Four tools alongside the `agent_session_*` family in `projects/monolith/agent_sessions/mcp.py`. All accept-and-return in the session-lane style, all no-ops with no companion open:

| tool | shape | behavior |
|------|-------|----------|
| `voice_ui_attach` | `{session_id}` | Binds any open companion to the session. Populates identity: model, vm state (from the CP listing, not session columns), repo, branch. Caller passes the session id explicitly because MCP calls carry no real actor identity today. |
| `voice_ui_show` | `{surface, ref, focus?}` | Conjures a surface: `run`, `walkthrough`, `transcript`, `vm`. Whatever it displaces starts decaying rather than being cleared. `focus` targets a node, a turn, or a file in the walkthrough. |
| `voice_ui_ask` | `{question, options, ref}` | Raises a decision card. Returns accepted immediately; the answer (spoken or clicked) is delivered back through the existing `agent_session_send` path as the next message, so the voice turn never blocks on the screen. |
| `voice_ui_dismiss` | `{surface?}` | Clears one surface or the stage. The user's own pin and dismiss controls always outrank the model's. |

Two behaviors that are deliberately not tools:

- **Auto-surfacing.** Session-lane calls the voice model makes (`semgrep_scan`, `agent_session_send`, `agent_trigger_job`) render as ephemeral tool cards when a companion is open. The wire ledger is the source, the card is its view; no new tool is minted for this.
[...]

PR #4984, opened 2026-08-16
docs(adr): record the voice companion contract as ADR 058

jomcgi:
Settles the five decisions the #4977 design handoff left open, so the voice companion work can start. Docs only: the ADR, its index row, and the two regenerated manifests the docs freshness checks require.

## The decisions

| Decision | Call | Why it went that way |
|---|---|---|
| Transport and ledger persistence | Durable table, polled | They are one choice, not two. A durable ledger makes "no surface without a ledger row" a SQL-checkable invariant instead of a review convention, and the audit trail costs nothing because the thing being polled is the thing being stored |
| Attach | Binds a named session, or mints one and returns its id | "Put it on the screen" often precedes "now work on this". A bare re-attach returns the existing binding, so a reconnecting tab cannot spawn sessions |
| Contention | Latest explicit attach wins | The only rule that keeps every `voice_ui_*` call accept-and-return. A refusal would need an error shape, a UI, and a voice line to explain it |
| Voice authorization | Two tiers, clicked and spoken answers as peers | A third click-only tier depends on a hand-maintained list of scary tool names, and such a list is wrong the first time someone forgets it |
| Principal scoping | Not in v1, gap stated | Better than a check that reads as enforced and enforces nothing |

## Two findings that shaped it

**An in-process ledger with SSE has a live tripwire under it.** It works today only because `backend.replicas: 1`. `projects/monolith/chart/values.yaml:26` already carries a disabled `backend.autoscaling` block and `templates/deployment.yaml:8` omits `spec.replicas` when it is enabled. The day that flips, a `voice_ui_show` landing on replica B while the browser holds its stream on replica A drops the surface with no error and no log. A durable ledger does not care which replica served the call.

**ADR 049's conclusion transfers but its reasoning does not.** 049 argues poll because the session pipeline is buffered at four hops (shim accumulates, `node.proto` is unary, the control plane parks on one `GenServer` call, the transport awaits the full body). A `voice_ui_*` event originates inside the monolith process that serves the companion, so none of that applies. The ADR says so explicitly, so a later reader does not think we inherited an untested argument. Push transport stays a legitimate later change, additive because the ledger is the source.

## On decision 5

The forwarded authentik token is already on the wire and already carries the right audience (`aud: [client_id, "https://private.jomcgi.dev"]`, so no RFC 8693 needed). The monolith ignores it. Making attach principal-scoped needs #4569 item 1, which is a shared prerequisite for the auth domain and deserves a standalone auth PR rather than arriving as a subsection of a UI feature.

What actually controls access meanwhile is real: the companion is on the private tier behind Cloudflare Access, so the viewer is already entitled to every session the console lists, and attach only changes which one they are shown. The ADR names the exact call that changes when #4569 lands, and that a call with no valid principal must refuse rather than degrade, since a mint is a write and there is no least-privilege version of creating a session.

`projects/mcp/ARCHITECTURE.md:143` is stale on the audience question; #4569 has the corrected version. Not fixed here to keep this docs-only, worth a follow-up.

## Verification

- `ci lint`: passed, 4 changed files, no reformats
- `ci test`: `Executed 4 out of 725 tests: 725 tests pass` ([invocation](https://app.buildbuddy.io/invocation/f9ea8be2-8e79-4528-bfdd-a7f764d607ae))
- Both manifests regenerated with the scripts the skill names, so the docs-site and repo-docs freshness checks have current input
- No `Chart.yaml` `version:` or `application.yaml` `targetRevision:` touched

## Not closing #4977

The ADR settles that issue's decision checklist. Its implementation checklist stays open, and the `needs-thought` label should come off in favour of `agent-ready` once this merges.

Status is `Accepted` rather than `Draft` because the five questions were put and answered rather than left as direction. Say the word if you would rather it sat as Draft until the first implementation PR.


jomcgi:
## Helm Manifest Diff

PR #4985, opened 2026-08-16
feat(embervm): add a pi-only guest image and a 256 MiB pi-runtime workload

jomcgi:
Second of three. Follows #4982 (footprint trim); a monolith-side routing PR follows this one.

## The waste this targets

`agent_sessions/__init__.py:16` maps model `qwen` to family `pi`, and every family goes through the single `claude-runtime` workload at **memMib 4096**. `ember_public/synthetic_probe.py:277` runs `probe_qwen()` on the `*/5` CronWorkflow. So every five minutes, forever, one trivial turn cold-boots a 4 GiB guest carrying a 262 MB Claude ELF and a 114 MB codex binary that a pi turn can never invoke.

Layer sizes on the published `runtimes/claude` image (300 MB compressed):

| layer | MB |
|---|---|
| codex CLI | 114.2 |
| claude CLI | 85.7 |
| wolfi base | 59.7 |
| **pi CLI** | **39.0** |
| guest-init | 2.3 |
| shim + guest config | 0.16 |

## What this adds

**`projects/embervm/runtimes/pi`** — the same guest with the two unusable CLI tars dropped, about 200 MB of the 300.

The shim, trust-record seed and guest-init are the **same build artifacts**, taken from the claude runtime package rather than rebuilt, so the two images cannot drift. No shim fork was needed: `ProcessManager` already constructs its claude/codex/pi adapters lazily and only spawns the CLI a turn selects.

The wolfi package set is **deliberately identical**. It is only 59.7 MB of the total, so trimming it buys little, while keeping it identical makes the image a behavioural drop-in for a pi turn. `package_parity_test` holds that claim instead of leaving it a comment, with an `EXPECTED_DIVERGENCE` map so a future divergence has to be deliberate and reviewable.

**`piRuntimeWorkload`** — session class, **memMib 256**, floor 0, cap 2, persistence off.

## Why 256 holds, and where it stops

What lives in guest RAM is the Bun-compiled pi binary, the stdlib-python shim, and the tmpfs over `/tmp` and `/workspace`. pi's code pages are mmapped from the rootfs **block device**, so they are reclaimable page cache rather than resident, and `PI_CONTEXT_WINDOW` caps context at 30000 tokens against Qwen's 32768, so the JS heap has a ceiling rather than a frontier model's appetite. The CRD floors `memMib` at 128 (`crds/workload-crd.yaml:215`), so this is legal, not a special case.

It stops holding the moment a session hydrates a repository: with persistence off, `/workspace` **is** the tmpfs, so a checkout spends these same 256 MiB with no disk to fall back on. **This is the small-task lane** (probes, short conversational turns). A repo-attached qwen session belongs on claude-runtime. The ladder if it proves tight is values-only: 256 -> 512 -> a `persistenceEnabled` flip with a `workspaceSizeBytes`.

Persistence off also means ADR 027's `memory:true` quadrant: a bank is 256 MiB of snapshot, cheap enough that the simple path is right, and it keeps this workload clear of the placeholder-volume-drive coupling that claude-runtime carries.

## The egress trap this closes

`_noded-pod.tpl` derived `EMBERVM_NODED_EGRESS_WORKLOADS` from `claudeRuntimeWorkload.name` **alone**. A second runtime workload would have been outside the allowlist, and that does not deny loudly, it dial-times-out inside the guest, while pi reaches the in-cluster vLLM endpoint over exactly that lane. The list is now derived from all enabled runtime workloads, keeping the chart's "derived, never hand-copied" property. The `__no_workload__` sentinel path is unchanged.

Rendered: `EMBERVM_NODED_EGRESS_WORKLOADS = "claude-runtime,pi-runtime"`.

## Nothing routes here yet, on purpose

The monolith still sends every family to `claude-runtime`. embervm and monolith are separate ArgoCD apps with independent rollouts, so shipping the routing flip in this PR would let whichever synced first decide whether qwen sessions 404 on an unknown workload for a few minutes. It also means the 256 MiB sizing is proven by a real session create **before** the `*/5` probe depends on it.

## Dev

Dev enables this workload, its one session workload, because dev is where "does a guest boot and report ready in 256 MiB" gets answered on a brick with room (1gi = 768 MiB guest-schedulable after #4982). Dev has no egress lane, so it proves base build, cold boot, readiness and create/destroy, **not** a turn.

`workloads.runtimePi.rootfsPath` is set for dev regardless of the CR, because base BUILDERS are gated separately from Workload CRs; left inherited it would be unbacked in the container and mkfs would write a multi-GB image to node-4's root filesystem.

## Verification

- `ci test`: `Executed 6 out of 727 tests: 727 tests pass` (727 vs 725 on main: the two new guards registered and ran)
- The apko image built on the remote runner, 55 packages, `arch=x86_64` only
- `helm template` against prod and dev values confirms the CR, the `build-runtime-pi-rootfs` initContainer, the dev-scoped rootfs path, and the egress list above

## Rollback

`piRuntimeWorkload.enabled: false` in both deploy values files. The image and chart plumbing are inert without it.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -529,7 +529,11 @@ apko.translate_lock(
     name = "embervm_runtime_claude_lock",
     lock = "//projects/embervm/runtimes/claude:apko.lock.json",
 )
-use_repo(apko, "embervm_lock", "embervm_noded_lock", "embervm_runtime_bazel_lock", "embervm_runtime_claude_lock", "embervm_runtime_k3s_agent_lock", "embervm_runtime_k3s_server_lock", "embervm_runtime_postgres_lock", "embervm_runtime_python_lock", "embervm_scratch_prep_lock", "embervm_tokenbroker_lock", "embervm_xds_lock", "monolith_frontend_lock", "rootfs_builder_lock", "sandbox_guest_lock", "semgrep_guest_lock", "whatsapp_lock")
+apko.translate_lock(
+    name = "embervm_runtime_pi_lock",
+    lock = "//projects/embervm/runtimes/pi:apko.lock.json",
+)
+use_repo(apko, "embervm_lock", "embervm_noded_lock", "embervm_runtime_bazel_lock", "embervm_runtime_claude_lock", "embervm_runtime_k3s_agent_lock", "embervm_runtime_k3s_server_lock", "embervm_runtime_pi_lock", "embervm_runtime_postgres_lock", "embervm_runtime_python_lock", "embervm_scratch_prep_lock", "embervm_tokenbroker_lock", "embervm_xds_lock", "monolith_frontend_lock", "rootfs_builder_lock", "sandbox_guest_lock", "semgrep_guest_lock", "whatsapp_lock")
 
 # Download multiarch archives (binaries from zip files)
 multiarch_http_archive = use_repo_rule("//bazel/tools/http:multiarch_http_archive.bzl", "multiarch_http_archive")
--- a/bazel/images/BUILD
+++ b/bazel/images/BUILD
@@ -36,6 +36,7 @@ multirun(
         "//projects/embervm/runtimes/claude:image.push",
         "//projects/embervm/runtimes/k3s/k3s-agent:image.push",
         "//projects/embervm/runtimes/k3s/k3s-server:image.push",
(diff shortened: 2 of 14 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4981, opened 2026-08-16
sandbox: six per-language code-executor guests, retire sessioned run_python


jomcgi:
## Problem statement

The agent code-executor sandbox runs exactly one language. `run_python` is the
only execution tool an agent has, so any question best answered by Go, Rust,
Elixir, OCaml, or JavaScript either gets answered from memory (badly) or gets
translated into Python first.

Separately, the sessioned variant (`sandbox-session`, EmberVM R2) has not earned
its keep: it holds a 2 GiB workload, a persistent-kernel code path in the guest,
a Postgres table, and a `session=` parameter on two tools, all to serve a warm
namespace across turns. It goes.

## Decisions taken

- **Six sibling guest images, one per language**, not one polyglot image. Each
  gets its own apko lock, its own pinned digest, its own Workload CR, and its own
  failure domain, so a Rust package regression cannot break Python.
- **One `guest-init` binary**, table-driven off a baked `/etc/sandbox-language`
  file. The language is fixed per image, so the guest never sees a language
  field: the monolith selects a language by choosing which workload URL to POST.
- **`run_python` is replaced by `run_code(language, code, files)`**, with
  `python` as the default value. Every caller, prompt, and WhatsApp copy
  reference updates with it.
- **`floor: 0` on every sandbox workload, including python.** Verified safe:
  `dispatcher.ex:637` has an explicit MISS tier that places a brand-new VM from
  the base snapshot when nothing warm is parked. Python trades 2 parked VMs for
  a restore on first invoke and gives the bricks their RAM back.
- **`rootfsReclaim.enabled: "1"`.** Verified on 2026-08-15 that no brick
  currently holds a reclaimable orphan (`*.tmp.*` or `*.building`), so arming it
  is arming an empty path, which is the safe moment to do it.

## Sizing

| Workload | memMib | floor | cap | timeoutSeconds |
|---|---|---|---|---|
| `sandbox-python` | 512 | 0 | 16 | 30 |
| `sandbox-go` | 1024 | 0 | 4 | 45 |
| `sandbox-rust` | 1024 | 0 | 4 | 45 |
| `sandbox-elixir` | 512 | 0 | 4 | 30 |
| `sandbox-ocaml` | 512 | 0 | 4 | 30 |
| `sandbox-javascript` | 512 | 0 | 4 | 30 |

Go and Rust carry more memory (tmpfs `GOCACHE`, rustc plus libLLVM page cache)
[...]

PR #4986, opened 2026-08-16
sandbox: six per-language guest images, retire sessioned execution from the guest

jomcgi:
First of two PRs for #4981. This one is **images only**, deliberately: CI pushes
images on merge to main, so a Workload CR referencing a not-yet-pushed image
sends `BaseBuilder` into a tight `BuildBase` retry loop flooding logs with
`FAILED_PRECONDITION`. That is the same failure that keeps `sampleWorkload.enabled`
at `false` today (`values.yaml:412`). The chart, the `run_code` tool, the drop
migration, and ADR 057 follow in PR 2 once these images exist in GHCR.

## What changes

**guest-init is parameterized by language.** A `Spec` table carries, per
language, the source filename, the optional compile argv, the run argv, the
language-specific environment, the warm-up argv, the writable cache dirs, and
the paths excluded from collected output. `sandbox-guest-init` resolves one Spec
at startup from `/etc/sandbox-language`, a one-line file baked per image.

The wire contract therefore keeps **no** language field: `{code, files?,
timeout_seconds?}` is unchanged. A caller selects a language by choosing which
workload to POST to, so language selection is a routing decision upstream and an
isolation boundary in EmberVM, rather than a branch inside the guest.

**Six sibling images** under `projects/firecracker/sandbox/`, replacing the
single `guest/` directory. They share one guest-init binary, layered from the
parent package so all six tar the same built artifact.

**Session mode is deleted** from the guest: `kernel.go`, its tests, and the
`Mode` / `SessionReset` wire fields. The EmberVM session class is untouched,
`claude-runtime` still uses it.

## Resolved toolchains

| Language | Packages |
|---|---|
| go | `go-1.26` 1.26.6-r0 |
| rust | `rust-1.97` 1.97.1-r0, `libLLVM-22`, `build-base` (gcc, binutils) |
| elixir | `erlang-28` 28.5.0.5-r0, `erl28-elixir-1.19` 1.19.5-r1 |
| ocaml | `ocaml` 5.5.0-r4, `ocamlfind` |
| javascript | `nodejs-24` 24.19.0-r0 |

Rust carries `build-base` because `rustc` does not link its own binaries, it
shells out to `cc`; without a C toolchain every compile fails at link time.

Go is pinned offline with `GOTOOLCHAIN=local` and `GOPROXY=off`. Without both, a
snippet importing something absent blocks on a network that is not there and
spends the whole wall clock before reporting a network error instead of the
compile error.

## Notes for review

- **Compile and run share one wall clock.** A compiled language does not get two
  budgets, and a failing compile short-circuits with the compiler's own output.
- **`ExcludeOutputs` is load-bearing for rust**, which would otherwise return its
  linked `main` binary as a multi-megabyte base64 attachment on every call.
- **`CacheDirs` is a field, not a name check.** Warm-up runs with
  `Spec.Environment`, so a language pointed at a cache dir nothing creates still
  exits 0 and still logs `language warm-up done` while writing somewhere the
  request path never reads. `TestCacheDirsCoverTmpEnvPaths` fails if that drifts.
- **The python rootfs re-bakes.** The rename alone would have hardlinked (the
  bake is keyed on content digest), but the lock refresh bumped `ncurses` by a
  week, so content moved.

## Pre-existing problems found, NOT fixed here

1. `projects/embervm/image/apko.yaml` pins `erlang-27=27.3.4.2-r0`, which Wolfi
   no longer ships (only `27.3.4.6-r1`), so that lock cannot be re-resolved.
2. `update-apko-locks.sh` sweeps `projects/platform/renovate/templates/apko-lock-*.yaml`
   into its `find`; those are a ConfigMap and a CronWorkflow, not apko configs.

Together these make the `update-apko-locks` pre-commit hook fail for anyone
touching any `apko.yaml`. This commit used `SKIP=update-apko-locks` after that
same hook had generated the six locks needed here. A repo-wide refresh also
rewrote 10 unrelated locks; all 10 were reverted, since a fleet-wide guest
re-bake does not belong in this PR.

## Verification

`ci test`: `Executed 11 out of 730 tests: 730 tests pass`, 1292 actions, with
apko genuinely building the new images in-run.

Refs #4981


jomcgi:
## Helm Manifest Diff

PR #4987, opened 2026-08-16
fix(apko): upgrade erlang to 27.3.4.16 and scope the lock sweep

jomcgi:
The `update-apko-locks` pre-commit hook currently fails for **anyone touching any
`apko.yaml`** in the repo, for two independent reasons. Found while adding the
per-language sandbox guests (#4981), fixed separately because neither is related
to that work.

## 1. The erlang pin could no longer be resolved

Wolfi drops old versions from its APKINDEX while leaving the apk itself
resolvable on the CDN. So `projects/embervm/image` kept **building** fine (the
pinned `erlang-27-27.3.4.2-r0.apk` URL still returns HTTP 200) while `apko lock`
could no longer satisfy the `erlang-27=27.3.4.2-r0` constraint.

That is the nasty shape of this failure: a stale exact pin breaks only at
*refresh* time, silently, and stays invisible right up until the blob is finally
purged, at which point it becomes a build outage with no code change behind it.

Both sides move together to **27.3.4.16**, as that file's own comment demands: an
`include_erts: false` release pins exact OTP app versions in its `.boot` script,
so a build/runtime patch mismatch crash-loops the pod at boot. 27.3.4.16 was
chosen because it is the newest version present in **both** the Wolfi index and
`builds.hex.pm/builds/otp/ubuntu-22.04`, not merely the newest of either.

## 2. The sweep matched files that are not apko configs

`find -name "apko-*.yaml"` also matches
`projects/platform/renovate/templates/apko-lock-{script,cronworkflow}.yaml`,
which are a ConfigMap and a CronWorkflow. apko can never resolve them, and since
the script exits non-zero when ANY config fails, those two alone were enough to
reject every commit.

They are now filtered by **shape** (a top-level `contents:` block) rather than by
path, because a path exclusion holds only until the next misnamed file lands
somewhere else. Skips are logged, so a real config that somehow fails the test
shows up as a visible line instead of quietly never being locked again.

## 3. Scoping, so the fix does not trade a loud failure for a silent one

Fixing the sweep alone would make the script *succeed* and rewrite all twelve
locks. Each rewritten lock re-bakes that guest's multi-GB rootfs on every brick,
so committing one `apko.yaml` would quietly drag a fleet-wide re-bake along with
it.

The hook now passes filenames and the script locks only the configs being
committed. Repo-wide is still available by running it with no arguments, which is
the `renovate-apko-lock-maintenance` CronWorkflow's job, on its own schedule,
where the resulting diff is reviewed as its own change.

After this, the hook reports **Passed** instead of Failed.

## Verification

`ci test`: `Executed 4 out of 725 tests: 725 tests pass`.

The OTP change invalidated and re-ran the whole Elixir pipeline rather than
cache-hitting it, which is the point:

```
From Executing genrule //projects/embervm/control:release_boot_smoke:
RELEASE BOOT OK: release booted and /healthz returned 200 on the executor
healthz_status=200
```

`release_boot_smoke` is exactly the check a build/runtime OTP mismatch would
fail. `//projects/embervm/control:release_tar`, the protoc-gen-elixir genrule,
and `//bazel/erlang:mix_test_smoke` all rebuilt against the new OTP too.

Refs #4981

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -26,12 +26,17 @@ repos:
         files: 'pyproject\.toml$'
         pass_filenames: false
 
+      # pass_filenames is true so a commit only re-locks the configs it actually
+      # touches. Repo-wide is still available by running the script with no
+      # arguments; it is the renovate-apko-lock-maintenance CronWorkflow's job,
+      # not something a single commit should drag along (each rewritten lock
+      # re-bakes that guest's rootfs on every brick).
       - id: update-apko-locks
         name: Update apko locks
         entry: bazel/tools/format/update-apko-locks.sh
         language: script
         files: 'apko\.yaml$'
-        pass_filenames: false
+        pass_filenames: true
 
       - id: update-atlas-sum
         name: Update Atlas migration checksums
--- a/bazel/erlang/repositories.bzl
+++ b/bazel/erlang/repositories.bzl
@@ -183,14 +183,21 @@ exports_files(["bin/protoc"], visibility = ["//visibility:public"])
 
 def _erlang_impl(_ctx):
     http_archive(
-        # OTP version MUST match the apko image's Wolfi erlang-27 (27.3.4.2)
+        # OTP version MUST match the apko image's Wolfi erlang-27 (27.3.4.16)
         # exactly: an include_erts:false release pins exact OTP app versions in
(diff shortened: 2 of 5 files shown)


jomcgi:
## Helm Manifest Diff

PR #4988, opened 2026-08-16
feat(moving): rebuild the planner page against the approved design

jomcgi:
Replaces the PR 3a page, and folds in what was deferred to PR 3b. Refs #4968.

## Why a rebuild rather than a fix

The first build was specced in prose, because the implementer could not open the design. It drifted badly: display type everywhere, invented navy and dark-green cell grounds, a full-bleed yellow masthead, and a vertical stack that scrolled instead of a grid that fits one screen. With an empty database the hero rendered its fallback string at 92px, so an empty plan read as a broken page.

This time the approved mockup's own HTML and CSS was placed in the worktree and ported directly, with the instruction that where the prose and the file disagree, the file wins.

## Corrections, all from the artifact's own CSS

- **Cells are paper on an ink ground.** `gap: 2px; background: var(--ink); border: 2px solid var(--ink)`, so the ground shows through as shared borders. No coloured cell backgrounds; navy and dark green are gone.
- **Yellow is a signal, not a ground.** It appears on the Do Today header band and pressed segmented buttons only. The artifact's comment: "A full yellow field under 8 rows of bold ink was fatiguing to read."
- **Coral is a mark**, on overdue chips and collision icons, never a ground under text.
- **The head strip is a 10.5px mono line** at `padding: 9px 18px`, not a masthead.
- **Type scale**: hero `clamp(54px, 6.2vw, 92px)` display, cell headers 11px mono, body 13.5 to 14.5px. Nothing else is display-scale.
- **Empty states are quiet**: body-size secondary text, cells keeping their shape and headers.

## What is now built

The gantt over real spans and milestones (held ones get the coral diamond), the dock with To sell, Calendar and Roles modals over real data, and the tabbed phone layout switching on the real viewport under 700px.

The artifact's mockbar is dropped. Its Joe/Anna switch existed because a mockup has no identity; the server load already resolves the viewer from the verified header. Its Desktop/Mobile switch existed because a mockup has no viewport; `matchMedia` does it, with no bezel, since a real phone is the bezel. The artifact's "Build" modal is also dropped: its schema and prompt panels document how the mockup was made and are not part of the app.

## Notes

Fonts come from a new `src/routes/friends/+layout.svelte` mirroring the public tier's Google Fonts link, so the display and mono faces load without touching the public or private layouts.

Theme tokens are scoped under `.moving`, not `:root`, so they cannot collide with the other tiers' identically-named tokens. Dark mode lives in the CSS file rather than a component `<style>` block, because the Svelte compiler silently comments out unscoped `html[data-*]` selectors while the build stays green.

## Verification

`//projects/monolith/frontend:test` and `:build_test` forced with `--nocache_test_results`: **Executed 2 out of 2 tests: 2 tests pass.** Forced deliberately: `:test` does not compile Svelte templates, and a 1,077-line page rewrite is where a template error would hide behind green unit tests.

9 vitest cases including new coverage for gantt positioning, agenda merging and sell totals. `ci lint` green. No `Chart.yaml` `version:` or `targetRevision:` touched, and `hooks.js`, the `:src_public` exclusion and `build-exclusions.test.js` are untouched.

### DIFF
--- a/projects/monolith/frontend/src/routes/friends/+layout.svelte
+++ b/projects/monolith/frontend/src/routes/friends/+layout.svelte
@@ -1,0 +1,18 @@
+<script>
+  let { children } = $props();
+</script>
+
+<svelte:head>
+  <link rel="preconnect" href="https://fonts.googleapis.com" />
+  <link
+    rel="preconnect"
+    href="https://fonts.gstatic.com"
+    crossorigin="anonymous"
+  />
+  <link
+    href="https://fonts.googleapis.com/css2?family=Hanken+Grotesk:ital,wght@0,300;0,400;0,500;0,600;1,300;1,400&family=Instrument+Serif:ital@0;1&family=JetBrains+Mono:wght@400;500;600;700&display=swap"
+    rel="stylesheet"
+  />
+</svelte:head>
+
+{@render children()}
--- a/projects/monolith/frontend/src/routes/friends/moving/+page.svelte
+++ b/projects/monolith/frontend/src/routes/friends/moving/+page.svelte
@@ -1,13 +1,22 @@
 <script>
   import { goto } from "$app/navigation";
   import { page } from "$app/stores";
+  import { onMount } from "svelte";
   import AccessPanel from "./AccessPanel.svelte";
   import {
     collisionWording,
+    formatCad,
+    formatDateRange,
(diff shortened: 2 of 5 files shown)


jomcgi:
## Helm Manifest Diff

PR #4992, opened 2026-08-16
sandbox: run_code language routing, six workloads, retire sessioned execution

### LINKED ISSUE #4981: sandbox: six per-language code-executor guests, retire sessioned run_python

jomcgi:
Second of two PRs for #4981. #4986 shipped the six guest images; all six are now
live in GHCR at `2026.08.16.06.51.52-1c2fecda3`, verified by `crane export`
rather than by tag:

```
etc/sandbox-language      <- "go" in the go image, "rust" in the rust image
etc/environment.md
usr/local/bin/sandbox-guest-init
usr/bin/go
```

This PR wires them up.

## Tool surface

`run_python(code, files, session)` becomes `run_code(language, code, files)`,
python by default. The language selects the WORKLOAD
(`SANDBOX_WORKLOAD_PREFIX` + language), because the guest carries no language
field: a language IS a workload.

This reverses ADR 044 decision 2, which chose a concrete `run_python` over a
generic `run_code(language=...)`. One of its two reasons is live rather than
theoretical: the Context Forge ACL is tool-granular
(`projects/mcp/ARCHITECTURE.md:41`), so a caller entitled to `run_code` is
entitled to all six languages. Accepted deliberately, and recorded in ADR 057:
the six are peers inside ONE security posture (same zero-egress microVM, same
uid 65532, same wall clock, same caps), so "may run python but not go" is not a
boundary anyone would draw. 044's argument was really about a hypothetical
`run_shell`, which WOULD widen the posture, and that stays its own tool.

## Two silent-failure classes fixed

**`_with_scratch_env` is now python-only.** It prepends PYTHON source
(`import os as _os; ...`) to inject the scratch DSN, which is a syntax error in
the other five. Unguarded it would have broken every non-python run while every
existing test stayed green. There is now a test asserting rust code is
byte-identical after the call with the DSN set.

**Renaming a tool has a prompt surface, not just a call surface.** Two leftovers
were behavioural, neither covered by a test:
- `chat/attention.py` builds the routing prompt listing the chat agent's tools,
  naming `run_python`. It would have advertised a tool that no longer exists.
- `chat/bot.py` branched on `event.part.tool_name == "run_python"` to render the
  live status line. Renamed, that branch never fires and it silently falls
  through to dumping the raw args dict.

The idempotency key now hashes language alongside code, or the same source text
submitted as two languages would dedupe onto one result.

## Chart

Six Workload CRs render from a range over `sandboxWorkloads` rather than six
near-identical files. The image pins CANNOT be collapsed the same way:
`helm_images_values` emits one YAML block per dotted key, so keys sharing a
top-level prefix collapse to the last on parse, hence flat
`sandbox<Lang>.guestImage`.

| Workload | memMib | floor | cap | timeout |
|---|---|---|---|---|
| `sandbox-python` | 512 | 0 | 16 | 30 |
| `sandbox-go` | 1024 | 0 | 4 | 45 |
| `sandbox-rust` | 1024 | 0 | 4 | 45 |
| `sandbox-elixir` | 512 | 0 | 4 | 30 |
| `sandbox-ocaml` | 512 | 0 | 4 | 30 |
| `sandbox-javascript` | 512 | 0 | 4 | 30 |

`floor: 0` everywhere including python, verified against `dispatcher.ex`
`pick_node/3`: with no warm VM parked it falls through to a MISS tier that
places a new VM from the base snapshot, so a zero floor serves and pays a
restore on first use. Python trades its two parked VMs for that, which is what
makes five mostly-cold languages affordable on the bricks.

Go and Rust carry more memory and a longer budget: Go's warm `GOCACHE` is
WRITTEN to the tmpfs so it is charged against `memMib`, Rust pages in rustc plus
libLLVM, and both run a compiler inside the same wall clock as the program.

Also corrects the semgrep `floor` comment, which still claimed floor applies PER
BRICK INSTANCE. `plan_fleet_allocation/4` computes deficits against
`fleet_free_totals/1`, which sums free slots across every instance, so a floor
has been ONE fleet-wide number since that fix landed.

## Removals

`sandbox-session`, the monolith session client half, `models.py`,
`repository.py`, `session_test.py`, `SANDBOX_SESSION_WORKLOAD`, and a drop
migration for `sandbox.session` (RESTRICT, never CASCADE, so an unexpected
object in that schema fails loudly instead of being silently destroyed).

The EmberVM session CLASS is untouched: `claude-runtime` still uses it. This
retires one consumer, not the substrate. `agent_sessions/transport.py` has its
own unrelated `create_session` and is not touched.

## Rollout note

`initContainers` run serially, so every brick now bakes six sandbox rootfs
before `noded` starts. On a cold digest cache that is roughly 4 to 5 minutes
each and bricks roll one at a time, so expect a long first roll with a capacity
dip. Subsequent rolls hardlink from the content-addressed cache.

`rootfsReclaim` is armed in the PROD DEPLOY VALUES, not the chart default: the
chart is published as an OCI artifact, so an armed default would delete base
bundles on any cluster installing it without an override. Verified before arming
that no brick held a reclaimable orphan, so it arms an empty path.

## Verification

`ci test`: `Executed 4 out of 731 tests: 731 tests pass`.

An earlier run caught two real failures worth naming, both from guards that
exist because the symptom is otherwise silent:

1. Every sandbox workload needed its own DEV `rootfsPath`. `rootfsPath` is a
   separate key from `noded.firecracker.nvmeRoot`, and dev mounts only its own
   scratch, so an inherited production path is unbacked in the container and
   mkfs writes multi-GB images to node-4's ROOT filesystem, rebuilt on every
   restart (#4837).
2. `rootfsReclaim` was armed in the chart default rather than deploy values
   (`rootfs_reclaim_guard_test`).

Chart verified by rendering: six CRs with correct names, `memMib` and resolved
image refs; six `build-sandbox-<lang>-rootfs` initContainers as valid RFC 1123
labels; six image-identity entries; zero `sandbox-session` references.

Closes #4981


jomcgi:
## Helm Manifest Diff

PR #4993, opened 2026-08-16
feat(embervm): track base builds per CPU vendor and report coverage

jomcgi:
Fixes the bookkeeping and visibility half of #4937. **Changes no build behaviour** — per-vendor enqueue is a deliberate follow-up, see Rollout below.

## The defect

A guest base is a **per-vendor artifact**: noded keys it `sha256(image_ref, workload_revision, cpu_vendor)` (base_builder.ex:1470), so one workload has one base per vendor with different refs. The CP tracked it with a **scalar** `snapshot_ref` / `built_signature`, set in `apply_result/4` from whichever build finished last.

Three predicates read that scalar:

| line | code | effect |
|---|---|---|
| 731 | `w.built_signature == signature(w) and w.snapshot_ref != nil` | reconcile becomes a no-op |
| 2748 | the same guard in `retry_workload/2` | retry no-ops |
| 2903 | `if w.snapshot_ref` | Ready = True |

So once one vendor builds, the workload is "done" for the whole fleet. Live: `pi-runtime` has an `amd` base, no `intel` base, `Ready: True`, and **zero** further build attempts over a stable 24-minute observation window. Three of four Firecracker nodes cannot serve it.

That also explains the hand-edited `EMBER_BASE_EPOCH` strings in `deploy/values.yaml`. They work by changing `signature(w)` so `built_signature == signature(w)` stops matching — manual cache-busters for a guard that should not have matched.

## What this PR does

- **`vendor_built`**: `vendor -> {signature, ref, digest}`, recorded from the building instance's reported `cpu_vendor`. An instance with **no capacity fact, or a blank vendor, records nothing** — a reporting gap is not evidence that a vendor lacks a base. This is also why the existing 1330-test suite passes untouched: most of it runs with no vendor facts.
- **Vendor-scoped turnover**, and this one is a latent bug fix. `turned_over?` compared against the scalar, so a build on a second vendor would push the *other* vendor's **live current ref** onto `superseded_refs` and into `base_refs`, and `maybe_evict_base/3` could later evict a base that is actively serving. Unreachable today only because nothing calls `report_base_refs/3` in production. Fixed before per-vendor builds make it reachable.
- **`BaseVendorCoverage`** condition, reporting missing and stale vendors distinctly (different problems). Written only from `write_base_status/3`, because this module owns the `conditions` array and a JSON merge-patch replaces it wholesale.
- CRD `conditions[].type` description updated; it still said `Ready | BaseBuilt`.

**`Ready` is deliberately unchanged.** A partially built workload is genuinely servable on the vendor that has a base, so failing Ready fleet-wide would convert a degraded state into an outage.

## Why coverage cannot be read off `snapshotRefs` today

`snapshot_refs_by_vendor/2` reports which vendors hold **a** ref, with no signature attached. So a vendor holding a week-stale base renders identically to a current one. The live fleet shows six workloads as `amd,intel` and only `pi-runtime` as `amd`-only, but the field log in `deploy/values.yaml` records that `claude-runtime`'s intel base is stale from Aug 8/10. `BaseVendorCoverage` compares each vendor's recorded signature against the desired one, which is the distinction that view structurally cannot make.

## Verification

- `ci test`: `Executed 1 out of 732 tests: 732 tests pass`, ExUnit `1330 tests, 0 failures, 6 skipped`
- **The test file diff is 304 additions, 0 deletions.** No existing test was edited, which is the actual proof this is a no-op.
- CRD `conditions[].type` is a free-form string (only `status` carries an enum), so the new condition type cannot 422 the status write.
- `apply_result/4` has exactly one caller; no consumer outside this module indexes into `conditions`.

8 new tests. The turnover one moves the pin cross-vendor (the same unpin/re-place churn a chart roll causes), then asserts `superseded_refs == []` — an assertion that **fails against the old code**. The coverage test asserts Ready is True in every status write, guarded by `assert built_and_later != []` so an empty list cannot pass it vacuously.

## Rollout

Inert. No enqueue or placement path changed, so no base rebuilds. On CP restart the workloads map is rebuilt from the watcher LIST re-cast exactly as today.

The follow-up that adds per-vendor enqueue **will** backfill every stale or missing (workload, vendor) pair on first deploy, onto intel nodes that are the masters carrying the 35 GiB capped scratch. Landing this first makes that backfill list readable from the CRs before arming it.

## Known carry-forward

One clause from the reviewed plan is not here: advancing the scalar only for a result from the primary pin. It is a no-op while every build runs on that single pin, and must land with per-vendor enqueue.

### DIFF
--- a/projects/embervm/chart/crds/workload-crd.yaml
+++ b/projects/embervm/chart/crds/workload-crd.yaml
@@ -1125,7 +1125,13 @@ spec:
                     properties:
                       type:
                         type: string
-                        description: Ready | BaseBuilt.
+                        description: >-
+                          Ready | BaseBuilt | BaseVendorCoverage. Ready means a
+                          restorable base exists SOMEWHERE, which on a
+                          vendor-split fleet is not the same as everywhere:
+                          BaseVendorCoverage is the per-CPU-vendor view and goes
+                          False while any known fleet vendor lacks a base at the
+                          desired signature.
                       status:
                         type: string
                         enum:
--- a/projects/embervm/control/lib/embervm/base_builder.ex
+++ b/projects/embervm/control/lib/embervm/base_builder.ex
@@ -3,7 +3,7 @@ defmodule Embervm.BaseBuilder do
   Turns a `Workload`'s OCI image into a pristine base snapshot by driving the
   node daemon's `BuildBase` RPC, then reports the result back into the
   Workload's `status` subresource (`snapshotRef`, `snapshotDigest`, and the
-  `Ready`/`BaseBuilt` conditions).
+  `Ready`/`BaseBuilt`/`BaseVendorCoverage` conditions).
 
   This is the first control-plane component that DRIVES the node daemon (issues
   a mutating RPC and reconciles its result into CR status), where
@@ -102,8 +102,9 @@ defmodule Embervm.BaseBuilder do
 
   Status is patched with a JSON merge-patch, which REPLACES arrays wholesale, so
   two writers touching `conditions` would clobber each other. Ownership is split
-  by key: this module owns `conditions` (`Ready` + `BaseBuilt`), `snapshotRef`,
-  and `snapshotDigest` for valid task Workloads; the watcher owns
+  by key: this module owns `conditions` (`Ready` + `BaseBuilt` +
+  `BaseVendorCoverage`), `snapshotRef`, and `snapshotDigest` for valid task
+  Workloads; the watcher owns
   `observedGeneration` and `primedFloorSatisfied` (disjoint keys, no lost
   update) and keeps `conditions` only for the invalid-CR validation lane, which
   never reaches this module.
@@ -454,8 +455,9 @@ defmodule Embervm.BaseBuilder do
       nodes: node_runtime,
       workloads: %{},
       workload_sync_done: false,
-      # pid -> %{node_id, name, signature} for the CURRENT build worker, so a
-      # result from a superseded worker (or for a since-changed spec) is dropped.
+      # pid -> %{node_id, name, signature, cpu_vendor} for the CURRENT build
+      # worker, so a result from a superseded worker (or for a since-changed spec)
+      # is dropped and an accepted result is attributed to its build vendor.
       workers: %{},
       # In-flight base exports, a MapSet of {workload, ref} (fast-durability-export
       # fix). A base export now returns a fast ack and completes asynchronously on
@@ -601,6 +603,7 @@ defmodule Embervm.BaseBuilder do
            snapshot_ref: w.snapshot_ref,
            snapshot_digest: w.snapshot_digest,
            built: w.built_signature != nil and w.built_signature == signature(w),
+           vendor_built: w.vendor_built,
            superseded_refs: w.superseded_refs,
            base_refs: w.base_refs,
            backoff_ms: w.backoff_ms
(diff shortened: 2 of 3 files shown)


jomcgi:
## Helm Manifest Diff

<details>
<summary><code>authentik</code></summary>

```diff

spec.template.spec  (apps/v1/Deployment/authentik/authentik-server)
  - one map entry removed:
    imagePullSecrets:
    - name: ghcr-imagepull-secret

spec.template.spec.containers.server.image  (apps/v1/Deployment/authentik/authentik-server)
  ± value change
    - ghcr.io/jomcgi/authentik-fork:2026.8.0-rc7
    + ghcr.io/goauthentik/server:2026.8.0-rc7

spec.template.spec  (apps/v1/Deployment/authentik/authentik-worker)
  - one map entry removed:
    imagePullSecrets:
    - name: ghcr-imagepull-secret

spec.template.spec.containers.worker.image  (apps/v1/Deployment/authentik/authentik-worker)
  ± value change
    - ghcr.io/jomcgi/authentik-fork:2026.8.0-rc7
    + ghcr.io/goauthentik/server:2026.8.0-rc7

(root level)  (onepassword.com/v1/OnePasswordItem/authentik/ghcr-imagepull-secret)
- one document removed:
  ---
  type: kubernetes.io/dockerconfigjson
  # Source: authentik-homelab/charts/authentik/templates/additional-objects.yaml
  apiVersion: onepassword.com/v1
  kind: OnePasswordItem
  spec:
    itemPath: vaults/k8s-homelab/items/ghcr-read-permissions
  metadata:
    name: ghcr-imagepull-secret
    namespace: authentik
    labels:
      app.kubernetes.io/name: authentik
      app.kubernetes.io/part-of: authentik
```

</details>

jomcgi:
## Review round 1 (Sol, adversarial) → four fixes in e68fcd5ac

Findings and disposition:

| # | severity | finding | disposition |
|---|---|---|---|
| 2 | **behavioural regression** | blank-vendor build then known-vendor rebuild leaves the previous ref never superseded, so it leaks on capped scratch | **fixed** via `scalar_vendor` |
| 1 | major | `vendor_built` is in-memory, so a CP restart reports live vendors as missing | **fixed** by splitting MISSING from UNVERIFIED |
| 3 | major | coverage computed only in `write_base_status/3`, which the already-built no-op branch never reaches, so a false True persists | **fixed** by republishing on change from the periodic tick |
| 4 | major | force rebuild clears the scalars but leaves the vendor record complete | **fixed**, drops the anchor vendor's entry |
| 5 | major | old in-flight workers crash after a hot code upgrade | **refuted, no action** |

### Why finding 2 blocked the merge

The parent commit *would* have superseded that ref (`A != B` on the scalar). This PR made it not, so it was a regression I introduced, in the direction that fills the masters' 35 GiB scratch. The fix records which vendor owns the current scalar ref and adopts it only when that vendor is blank or the same; a different vendor still yields nil, so the cross-vendor superseding this work exists to prevent stays prevented.

### Why finding 5 is refuted

It assumes an in-place BEAM upgrade with old-shape worker meta still in flight. The control plane is a Kubernetes Deployment: a new version is a new pod with a fresh BEAM. Adding `code_change/3` and legacy-shape clauses would be dead defensive code for a state that cannot occur.

### Coverage is now three-state, and that matters for the follow-up

| record | observed ref | classification |
|---|---|---|
| no | no | **missing** (genuinely absent) |
| no | yes | **unverified** (cannot vouch for signature) |
| yes | stale signature | **stale** |
| yes | current signature | covered |

Disjoint and exhaustive over `fleet_vendors`. The distinction is load-bearing for the next PR, where **missing** becomes real multi-GB bakes: without it, every CP restart would reclassify healthy vendors as missing and backfill the whole fleet onto the masters.

Verification: `ci lint` clean; the ExUnit suite runs inside the `//bazel/erlang:mix_test_smoke` genrule (not a test target, so it never appears in the `Executed N out of M tests` line) and passes at `1335 tests, 0 failures, 6 skipped`, up 5 from 1330. Test file diff remains additions-only across both commits.

PR #4996, opened 2026-08-16
feat(monolith): voice_ui ledger, tools, and companion routes

jomcgi:
Slice 1 of the voice companion, implementing ADR agents/058. Backend only, no frontend: a companion registers over HTTP, the voice model drives it over MCP, and every accepted call writes exactly one ledger row. Refs #4977.

## What lands

- **Two tables** under the `agent_sessions` schema. `voice_ui_companions` holds the binding, `voice_ui_ledger` holds one row per accepted call. Polling the ledger is the transport and doubles as the companion heartbeat, so there is no second liveness mechanism.
- **Four MCP tools**: `monolith_voice_ui_attach`, `_show`, `_ask`, `_dismiss`, beside the existing `agent_session_*` family.
- **Two routes**: `POST /api/agents/companion` (mint or heartbeat) and `GET /api/agents/companion/{id}/ledger?since=` (poll).

## The two invariants, as tests

- **Degrade to voice.** Every tool with no companion open accepts, writes nothing, and raises nothing. Nothing in the voice path can depend on a screen.
- **No surface without a ledger row.** Every accepted call writes exactly one row before returning, carrying the caller principal.

## Decisions worth a reviewer's attention

**The principal is recorded and gated on nowhere**, per ADR 058 decision 5. This is the first consumer of the auth domain from #4955: `current_principal()` had no callers outside `projects/monolith/auth/`. Rows will read `authority: anonymous` until tokens actually arrive on this path, and that visibility is the point rather than a gap. There is deliberately no disabled flag.

**Arguments are validated before the companion lookup.** An earlier revision checked the companion first, so `show(surface="typo")` returned `accepted: True` with no companion and `accepted: False` with one: the same malformed call answered two ways depending on unrelated UI state, and a typo was invisible in the common case. Validating first makes the answer constant regardless of the screen, which is what degrade-to-voice actually asks for. `companion_open` is omitted from the rejection path because at that point it is undetermined, and asserting it would be a guess.

**A bare `attach()` is idempotent.** It returns the companion's existing binding and mints only when unbound, so a reconnecting tab cannot spawn sessions. An explicit `attach(id)` always rebinds, which is the ADR's latest-attach-wins.

**A minted session is zero-turn and keeps `DEFAULT_AGENT_REPO`.** `_persist_session` is DB-only and the mint never calls `_schedule_next_message`, so no guest is created and nothing is checked out. Recording the repo costs nothing now and avoids permanently limiting a session whose repo is fixed at creation. A test asserts the mint does not schedule a message.

## One trap this PR had to fix

`projects/monolith/BUILD` is hand-maintained and declares every `py_test` explicitly. The `agent_sessions/**/*.py` glob feeding `pkg_agent_sessions` **excludes** `**/*_test.py`, so a new test file gets no target and never runs. With the 13 tests present but unwired, `ci test` reported `Executed 0 out of 732 tests: 732 tests pass`, exit 0, lint green. Adding `agent_sessions_voice_ui_test` took it to `Executed 25 out of 733`. `pkg_agent_sessions` also gained `:pkg_auth`, since `mcp.py` and `router.py` now read the principal. Production was never affected, because `auth/**/*.py` is already in `_BACKEND_SRCS`; the gap was confined to the per-package library that test targets use.

## Verification

- `ci lint`: clean
- `ci test`: `Executed 80 out of 732 tests: 732 tests pass` ([invocation](https://app.buildbuddy.io/invocation/b4c95e16-45ec-4e27-89d0-a1fdcdc28e02)), with the suite reporting `1322 tests, 0 failures, 6 skipped`
- No `Chart.yaml` `version:` or `application.yaml` `targetRevision:` touched
- Migration is 24 lines, well inside the migrations ConfigMap limit, and sorts after `20260816010000_drop_sandbox_session.sql`

## Not in this slice

The companion UI, `ask` answer resolution through `agent_session_send`, and surface lifecycle and decay. `voice_ui_ask` records the question and returns immediately; it never blocks a voice turn.

### DIFF
--- a/projects/monolith/BUILD
+++ b/projects/monolith/BUILD
@@ -1243,6 +1243,8 @@ py_library(
     visibility = ["//:__subpackages__"],
     deps = [
         ":agent_sessions_rationale",
+        # mcp.py and router.py read the request principal via auth.api.
+        ":pkg_auth",
         ":pkg_core",
         ":pkg_cyclegroup",
         ":pkg_faas",
@@ -6910,6 +6912,23 @@ py_test(
         ":pkg_agent_sessions",
         "@pip//httpx",
         "@pip//pytest",
+    ],
+)
+
+py_test(
+    name = "agent_sessions_voice_ui_test",
+    srcs = ["agent_sessions/voice_ui_test.py"],
+    imports = ["."],
+    deps = [
+        ":pkg_agent_sessions",
+        # The companion routes and the voice_ui_* tools read the request
+        # principal, so the test drives auth.dependencies directly.
+        ":pkg_auth",
+        ":pkg_core",
+        "@pip//fastapi",
+        "@pip//httpx",
+        "@pip//pytest",
+        "@pip//sqlmodel",
     ],
 )
 
--- a/projects/monolith/agent_sessions/mcp.py
+++ b/projects/monolith/agent_sessions/mcp.py
@@ -720,7 +720,7 @@ def _mint_voice_ui_session() -> AgentSession:
 
 def _voice_ui_principal() -> tuple[str, str]:
     principal = current_principal()
-    # Principal facts are recorded, never used as a gate. ADR 057 deliberately
+    # Principal facts are recorded, never used as a gate. ADR 058 deliberately
     # keeps the voice path available when MCP carries an anonymous principal.
     return principal.subject, str(principal.authority)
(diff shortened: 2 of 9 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4994, opened 2026-08-16
feat(embervm): website snapshotter task guest, headless Chromium screenshots over MCP


jomcgi:
## Summary

Give agents a way to see what is actually deployed: an EmberVM task-class `shotter` workload that renders a `jomcgi.dev` or `private.jomcgi.dev` page in headless Chromium and returns a PNG over MCP.

## Why this shape

EmberVM task guests restore a post-warm snapshot at 2.5ms load-to-resume (`ARCHITECTURE.md:850`), and the base snapshot is cut once `/shim/ready` first returns 200, which the guest flips only after the `Warm` hook (`sandbox/guest-init/cmd/main.go:51-59`, `noded/server/server.go:844`). So a browser launched in `Warm` is resident in every restored clone. `runtimes/bazel` already proves this pattern by snapshotting a live warm JVM as a task-class base. Net effect: a fresh browser per screenshot at restore cost rather than launch cost, which is faster than a long-lived pod could offer and keeps per-shot isolation.

Chromium is the engine because the target is a hydrating SvelteKit app with token-driven CSS. Lighter renderers (wkhtmltoimage, WeasyPrint, satori) would produce a plausible but wrong picture, which is worse than none when the point is fidelity. Playwright is deliberately not used: it is a Node driver over CDP, and driving CDP directly from the existing Go guest-init keeps Node out of the rootfs (bases fill the masters' 35GiB scratch).

## Routing decision

Public hostnames are mapped to internal services rather than dialled through Cloudflare:

| Requested host | Dialled |
|---|---|
| `jomcgi.dev` | `monolith-public-frontend.monolith-public.svc.cluster.local:3000` |
| `private.jomcgi.dev` | `monolith.monolith.svc.cluster.local:3000` |

Rationale: `private.jomcgi.dev` sits behind Cloudflare Access, so a credential-less browser going out through the CDN would faithfully screenshot the login wall and it would look like a successful capture. Mapping internally also avoids putting a CF Access service token behind a page renderer.

The egress sidecar has no host-rewriting capability (`egress-proxy/cmd/swap.go:474-475` states it never touches the URL), so the mapping lives in the in-guest proxy we write. That proxy reads the absolute-form request line to build the vsock preamble and replays the request head verbatim, so `Host: jomcgi.dev` survives to the origin and host-keyed routes such as `/agents` still resolve correctly.

## Security constraint (load-bearing)

[...]

PR #4997, opened 2026-08-16
test(monolith): failing BDD specs for the shotter website snapshotter

jomcgi:
Phase 3 of #4994. Specs define done before implementation lands.

29 tests, tagged `future = True` so the gating Test action (which excludes `-future`) stays green while they are red. They fail with `ModuleNotFoundError` for `shotter.mcp`, `shotter.client`, and `shotter.hosts`, which is the intended state until Phase 4.

## What is specified

- **Host validation**, the security-critical part. Accepts only `jomcgi.dev` and `private.jomcgi.dev`. Rejects `api.github.com` (named explicitly, it is the credential-injection risk in ADR embervm/035 section 4), suffix lookalikes (`evil-jomcgi.dev`, `jomcgi.dev.evil.com`), unmapped subdomains, non-https schemes, embedded credentials, and non-standard ports. Matching is exact, mirroring the egress sidecar's own exact matcher.
- **Host to internal service mapping**, asserted on exact values. Substring matching is not sufficient here: "monolith" is a substring of "monolith-public-frontend", so a loose check passes even when the private tier is wired to the public service.
- **Viewport bounds**, rejecting absurd dimensions rather than clamping.
- **Return shape**: one `mcp.types.ImageContent` block with `mimeType == "image/png"` plus structured metadata carrying the stored URL, per ADR embervm/035 section 5.
- **Timeout conversion**, with a stubbed transport so the test fails only when a timeout does not surface as `ToolTimeout`.

## What is deliberately NOT specified here

The in-guest proxy allowlist. It is Go running inside the Firecracker guest, so a Python stand-in would prove the shape of the seam while proving nothing about what the guest enforces. A Go test in that package covers it in Phase 4, and the requirement is recorded on #4994.

## Note for reviewers: the `bdd_completeness_test` ignore

`bdd_completeness_test` takes `data = glob(["*/tests/*_test.py"])`, so every domain's tests enter its runfiles and each is suppressed by a hand-maintained `--ignore` in its `PYTEST_ADDOPTS`. Without adding `shotter`, pytest collected and ran these red specs inside that target. `future = True` does not help, since that tags the dedicated `bdd_test` target while this is a separate `py_test` picking the file up by glob. A comment now explains why the list has to be maintained.

Verified by running the target on clean `origin/main` (passes), on this branch without the ignore (fails), and on this branch with it (passes).

Refs #4994

### DIFF
--- a/projects/monolith/BUILD
+++ b/projects/monolith/BUILD
@@ -6666,12 +6666,12 @@ bdd_test(
 # lands the implementation.
 bdd_test(
     name = "shotter_bdd_mcp_tool_test",
-    future = True,
     srcs = [
         "shotter/tests/__init__.py",
         "shotter/tests/bdd_mcp_tool_test.py",
         "shotter/tests/conftest.py",
     ],
+    future = True,
 )
 
 py_test(
--- a/projects/monolith/BUILD
+++ b/projects/monolith/BUILD
@@ -17,6 +17,7 @@
 # gazelle:exclude scripts
 # gazelle:exclude home
 # gazelle:exclude sandbox
+# gazelle:exclude shotter
 # gazelle:exclude scheduler
 # gazelle:exclude semgrep_scan
 # gazelle:exclude goosecracker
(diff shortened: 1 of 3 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4429, opened 2026-08-06
embervm: guest egress lane runs ~9x slower than the mirror, 11.5s vs 1.3s for the same clone


jomcgi:
Hydration is now the dominant cost of an initial session, and the bottleneck is the LANE, not the git mirror and not network capacity.

## Measurement

Same repo, same `--depth=1 --single-branch` clone, same 11.24 MiB pack:

| path | time |
|------|------|
| inside the guest, through the vsock lane | **11,451 ms** (`phase=hydration_clone`) |
| at the mirror over loopback, no lane | **1,309 ms** (includes kubectl exec overhead, so real server time is under a second) |

The mirror serves this fine. The lane costs roughly 10 extra seconds, about a 9x penalty.

For context on why this now matters: with warm restore armed and the shallow clone shipped, an initial repo-attached session is ~15.2s end to end, and 11.5s of that is this clone. The relit turn is ~4s and is almost entirely model latency. So this single number is the remaining first-message cost.

## It is not bandwidth, and not the router

The datapath is guest -> vsock lane -> noded forwarder -> egress proxy -> git-mirror pod. All of it is host-internal or in-cluster, so physical network capacity is irrelevant.

It is also not an obvious small-buffer bug: the shim's copy loop already reads 64 KiB at a time (`source.recv(64 * 1024)`, shim.py:775).

11.24 MiB in ~10.2s of lane time at 64 KiB per chunk is roughly 180 chunks, so about **55ms per 64 KiB chunk**. That is a per-round-trip latency signature, not a throughput ceiling. Something in the chain appears to be serialising on a round trip per chunk rather than streaming.

## Where to look

- the `core.gitProxy` helper written by `_write_git_proxy_helper` (shim.py:239) and how it bridges to the vsock lane
- the noded-side forwarder: whether it reads and writes in lockstep per chunk instead of pipelining
- whether any hop is doing a synchronous request/response per buffer, or setting TCP_NODELAY / disabling Nagle in a way that costs a round trip per write
- vsock queue sizing on the FC device

## Suggested first experiment

Measure raw lane throughput independent of git: push a known payload of a few MiB through the lane and time it, then vary the buffer size. That separates "the lane is slow for all traffic" from "git's request/response pattern interacts badly with the lane". The answer determines whether this is a copy-loop fix or a protocol-shape fix.

## Related

- #4417 (second lane connection wedges) is a different defect on the same lane, and is why hydration uses a single-connection full-history-free clone at all
[...]

PR #4998, opened 2026-08-16
feat(embervm): website snapshotter, headless Chromium screenshots over MCP

### LINKED ISSUE #4994: feat(embervm): website snapshotter task guest, headless Chromium screenshots over MCP

jomcgi:
Phase 4 of #4994. Gives agents a way to see what is actually deployed, rather than inferring it from source.

A new task-class EmberVM workload, `shotter`, runs headless Chromium, renders a `jomcgi.dev` or `private.jomcgi.dev` page, and returns a PNG through a monolith MCP tool as a real `ImageContent` block.

Design and rationale: ADR [embervm/035](docs/decisions/embervm/035-website-snapshotter-task-guest.md). The 29 BDD specs that define the monolith contract merged separately in #4997 and are unedited here; they move out of the `future` lane into the gating suite now that they pass. STPA refreshes for both touched systems are in #5000 and #5001, held until this lands so the safety model does not describe a workload that is not on main.

## One decision I did not make

**On deploy, any authenticated MCP caller can screenshot `private.jomcgi.dev`.** Context Forge federates a newly registered tool at `visibility=public`, monolith does no per-tool authorization, and the forwarded authentik token stays inert until #4569 validates it. The reconcile pass that would fix this does not exist.

I do not read this as a blocker, because every existing monolith MCP tool already has the same property and `monolith-k8s-get-pod-logs` arguably exposes more. But private-tier *pages* are a new kind of content behind that door, so it is your call whether to ship both tiers now or restrict to the public tier until #4569. Restricting would be a small change to `HOST_SERVICE_MAP` plus the baked guest policy.

## Why a task-class guest

`noded` cuts the base snapshot the first time `/shim/ready` returns 200, and a guest flips readiness only after its warm hook. So Chromium launched during warm is resident in every restored clone, at 2.5ms load-to-resume. Each screenshot gets a **fresh** browser at restore cost rather than launch cost, which a long-lived pod cannot offer without sharing browser state between callers. `runtimes/bazel` already proves the pattern by snapshotting a live warm JVM.

Chromium comes from the Wolfi apk, so apko resolves its shared-library closure and no hand-mirrored Debian package list is needed. No Node and no Playwright: driving CDP directly from the existing Go guest-init keeps a second language runtime out of a rootfs that competes for the masters' 35GiB scratch.

## The security model

Egress policy is per-sidecar, not per-workload. Any workload granted egress inherits `external: allow` plus credential injection toward `api.github.com`, and the sidecar cannot tell a browser's incidental subresource fetch from a deliberate one.

The in-guest proxy is therefore the **primary** control, and monolith-side URL validation is a second, independent layer rather than the thing relied on. The proxy checks its allowlist against the destination **actually being dialled** (after mapping, so a redirect or subresource elsewhere is refused identically), refuses **before any vsock dial**, and matches exactly on parsed host and port.

Its policy is **baked into the rootfs** at `/etc/shotter-egress.json`, not delivered as config. A task-class guest has no env delivery path at all (`initEnv` is inert past the base-cache hash per #4429, `ember.env.*` only fires for `ClaimStateful`/`ClaimGroupMember`), and more importantly config a caller or a mutable control-plane field could set would not be a control at all. A test now asserts the baked policy is a subset of the chart allowlist, so drift cannot land silently.

## Two things that look wrong and are not

**Navigation uses `http://` internally** even though the tool accepts `https://`. The proxy maps on host only and replaces the port with the internal plaintext 3000, so an `https://` URL would open the tunnel and then die in a TLS handshake against an HTTP server; the sidecar cannot terminate TLS. Safe because the proxy replays the request head byte for byte, so the origin still sees `Host: jomcgi.dev` and host-keyed routes such as `/agents` resolve. A `CONNECT` to a mapped host is now refused explicitly rather than tunnelled into an opaque TLS error.

**PNG upload is best-effort but not silent.** The URL is derived from the content hash before the upload is attempted, so `put_screenshot` returns `(url, stored)` and the tool surfaces `stored`. A SeaweedFS blip should not discard a screenshot the caller can already see inline, but a URL that 404s must never be indistinguishable from one that resolves.

## Review round

An Opus review of the full diff plus two STPA passes produced 15 findings, all fixed in `fix(embervm): address review findings`. The ones worth knowing about:

- **The feature could not have worked.** Both destinations were Cilium ingress default-deny: the public frontend admitted only `envoy-gateway-system`, and monolith admitted `embervm` on **8091 only**, which does not extend to 3000 because a `toPorts` entry is per port. Every capture would have failed as a silent dial timeout (PR #4294). Both policies now grant `embervm` to the SSR port. This also falsified a justification I had written into `deploy/values.yaml` and into ADR 035 section 4, that these endpoints were already reachable in-cluster. They were not; the accepted-cost conclusion survives only because the widening is *smaller* than claimed. The values comment is corrected here and the ADR needs a follow-up amendment.
- **The Idempotency-Key served a stale screenshot** for the whole result TTL. It was copied from `sandbox/client.py`, where the premise holds because code-to-output is pure. A URL is not. An agent could screenshot a page, deploy a fix, screenshot again, and get the byte-identical pre-fix image with no indication it was cached. Now a per-call nonce.
- **Readiness flipped on two paths where the guest was broken**: when the proxy config failed to load, and when Chromium answered CDP without being able to render. Both baked a permanently failing guest into the base snapshot while every layer looked healthy. Warm now does a real trial capture through the same path an invocation uses, and readiness stays false on config failure.
- **The timeout chain did not nest** (handler cap reached 55s against a 50s workload budget) and `timeout_ms` was never validated. **The PNG cap** was checked against decoded bytes while base64 crosses the transport, leaving about 90 bytes of headroom rather than 2 MiB.

The reviewer probed the proxy allowlist with trailing dots, case folding, IDN, IPv6 bracketing, port substitution, header smuggling and pipelined bytes: all failed closed, with the check running against the resolved destination before any dial.

## Nesting

Timeouts: Context Forge 60s > monolith read 55s > workload 50s > guest handler 48s > CDP navigate.
Sizes: encoded PNG 7 MiB < workload `resultMaxBytes` 8 MiB < the 32 MiB noded reads back.

Memory is coupled to the guest tmpfs by hand, since nothing in the build ties a Go string literal to a YAML integer: `memMib` 2048 holds the 512m `/tmp` tmpfs plus Chromium's working set, commented on both sides.

## Verification

- `ci`: `Executed 51 out of 735 tests: 735 tests pass`
- The apko image builds remotely (`Build completed successfully`)
- Chart rendered for **both** environments: prod emits the Workload, the `build-shotter-rootfs` initContainer, `EMBERVM_NODED_EGRESS_WORKLOADS=claude-runtime,pi-runtime,shotter`, and both allowlist entries; dev emits none of them
- Both Cilium policies rendered and confirmed to grant `embervm` to port 3000
- `chart_env_identity_test` now lists `shotter` so the dev disable is enforced rather than merely written down

## Open, not blocking

- ADR 035 is status **Draft**, and its section 4 needs the correction described above.
- `egress.internal.allowlist` is global to the sidecar, so the two frontend entries widen reach for the claude runtime too. Carried in the ADR as an open question since it sets precedent.
- Stored screenshots have no retention policy, including of the private tier (STPA #5001, `shotter.store.wrong-duration`).
- `client.py` assigns `client.timeout` after constructing `httpx.AsyncClient()` instead of passing the kwarg, because the merged BDD fakes have bare zero-argument constructors. Production code bent around a test double; the fix belongs in the specs.
- The tool uses an explicit `mcp.tool(name=...)`, the first in the repo.
- `full_page` and `wait_until` are implemented in the guest and plumbed through the client but not exposed on the tool.

Closes #4994


jomcgi:
## Helm Manifest Diff

Issue #4999, opened 2026-08-16
embervm: shotter readiness check does not verify Chromium can render — STPA shotter-readiness.corrupted


jomcgi:
## Finding

STPA safety model for `projects/embervm` (`projects/embervm/STPA.md`, finding `shotter-readiness.corrupted`), surfaced while refreshing the model for the shotter website snapshotter (ADR embervm/035, PR #4998, issue #4994).

`ember-shotter-init` gates `/shim/ready` on Chromium's CDP HTTP endpoint answering `GET /json/version` with a non-empty `Browser` and `webSocketDebuggerUrl`, plus a fixed settle delay:

`projects/embervm/runtimes/shotter/guest-init/cmd/main.go:180` (`waitForCDP`), flip at `main.go:105` (`ready.Store(true)`).

That proves the DevTools control protocol is alive, not that the renderer can actually navigate and capture a page (e.g. a stalled or crashed renderer process behind an otherwise-responsive CDP HTTP server). noded snapshots the base the instant `/shim/ready` first returns 200 (`projects/embervm/noded/server/server.go:838-844`), so a control-plane-reachable but functionally broken browser gets baked into every clone restored from that base. Each restored clone still returns a legible per-request error (`errNavigationFailed` / `errNavigationTimeout`, `guest-init/cmd/screenshot.go`), so no individual caller sees a silent failure, but the *system* degrades quietly: every future invocation fails the same way until someone notices the pattern and rebuilds the base, instead of the intended single, loud failure at BuildBase time that the package doc for `ember-shotter-init` describes as the goal ("A warm failure never flips readiness... BuildBase then fails on its outer boot timeout instead of silently snapshotting a cold or half-started browser").

## Suggested fix

Extend the warm-readiness check with an actual functional self-test over CDP before flipping `/shim/ready`: open a target, navigate to a trivial page (e.g. `about:blank` or a `data:` URL), and confirm the navigation lifecycle event fires, not just that the HTTP control endpoint answers.

## References

- `projects/embervm/STPA.md`, finding `shotter-readiness.corrupted` / hazard `shotter-broken-base-snapshot`
- ADR `docs/decisions/embervm/035-website-snapshotter-task-guest.md`
- Tracking issue #4994, implementation PR #4998

PR #5000, opened 2026-08-16
docs(stpa): add shotter task-guest safety model for embervm

jomcgi:
## Summary

Refreshes `projects/embervm/STPA.md` for the website snapshotter task guest (ADR `docs/decisions/embervm/035-website-snapshotter-task-guest.md`, issue #4994, implementation PR #4998, not yet on main).

- Adds `shotter-chromium` and `shotter-proxy` nodes/control actions to the physical view: the browser's content-driven egress and the in-guest hard-allowlist proxy that is the primary control on it.
- Adds unsafe-feedback `shotter-readiness.corrupted`: the base snapshot is cut the instant Chromium's CDP HTTP endpoint answers `/json/version`, which proves the control protocol is alive but not that the renderer can actually navigate and capture a page, so a functionally broken browser can get baked into every restored clone. Tracked as #4999.
- Adds hazards: `shared-egress-allowlist-widening` (the two new `egress.internal.allowlist` entries are global to the sidecar, reachable by every other egress-enabled workload, not only shotter), `shotter-policy-drift` (the guest's baked `/etc/shotter-egress.json` and the chart's allowlist are two hand-maintained sources with no build-time check that they agree), `shotter-memmib-tmpfs-coupling` (memMib and the guest's tmpfs size are unenforced across a YAML value and a Go constant), and `egress-workload-derivation-gap` (the per-workload egress forwarder list is a hand-written if-chain; a future consumer left out gets a silent dial-timeout, not a loud deny).
- Examined and rejected as non-UCAs, with reasoning: allowlist bypass via redirect/subresource, requested-vs-dialled host mismatch, CDP-target leaks across invocations, and config-load failure mode. All are mitigated in the existing control structure.

## Test plan

- [x] Rendered via the STPA skill's deterministic jq renderer (BLOCK A); diff reviewed against `origin/main`'s prior version.
- [x] `ci lint` and `ci test` ran clean on push (pre-push hooks).
- [ ] Human review of the new hazards, in particular `shotter-policy-drift` and `shared-egress-allowlist-widening`, which are unmitigated open items.

Refs #4994.


jomcgi:
## Helm Manifest Diff

PR #5003, opened 2026-08-16
feat(monolith): route qwen sessions to the pi-runtime workload

jomcgi:
Routes the pi family (`model=qwen`) off the shared `claude-runtime` workload onto the 256 MiB `pi-runtime` lane. Closes the follow-up named in `projects/embervm/deploy/values.yaml`, which stood the workload up but deliberately left the monolith-side routing for a separate change.

## Why

`claude-runtime` is sized at 4096 MiB to hold a 262 MB Node ELF with a V8 heap for the claude CLI. A qwen turn can never invoke it. `pi-runtime` runs the *same* shim on a pi-only image (200 MB of unusable CLIs dropped, roughly 3x smaller) at 256 MiB, so a banked qwen session snapshot is 256 MiB rather than 4 GiB.

## The seam

The workload name appears in exactly two URLs, both `/v1/workloads/{name}/sessions` (create and list). Invoke and destroy are session-scoped, so **existing qwen sessions on `claude-runtime` keep working untouched**: only new creates move. No migration.

All four `create_session` call sites in `deliver` thread the model. The fourth is the `403/410` mid-conversation recovery arm, and it is the one that fails silently if missed: the turn still succeeds, it just runs on the wrong lane, so nothing surfaces the escape. There is a test that drives the real `create_session` through that arm.

`list_sessions` takes an optional `workload` rather than aggregating the two lanes, because the session cap is per workload (`pi-runtime` has `cap: 2`) and an operator managing a cap wants one lane at a time.

## The risk, and why it is acceptable

`ember_public/synthetic_probe.py` `probe_qwen` calls `run_synthetic_session(model="qwen")` through this same transport every five minutes and latches `ember_synthetic_probe`, which `/api/health` 503s off. **This change moves a public health signal onto the new lane.**

So the lane was proven in production first, driven from the monolith pod with its own ServiceAccount token so it exercised the real principal:

| step | result |
|---|---|
| `POST /v1/workloads/pi-runtime/sessions` | `201` in 0.1s, `runtimes/pi@sha256:dbd2a45b...` |
| `POST /v1/sessions/{id}/invoke` | `200` in 1.8s, `result "qwen synthetic ok"`, `terminal_reason "stop"`, usage 522 in / 27 out (real llama.cpp round trip) |
| `DELETE /v1/sessions/{id}` | `202` |

`probe_qwen` accepts `terminal_reason` in `{"completed", "stop"}`, so its assertion is satisfied by the observed behaviour, and the 256 MiB sizing holds under a real session.

Egress needed nothing: the live `EMBERVM_NODED_EGRESS_WORKLOADS` is already `claude-runtime,pi-runtime` (the chart derives it from enabled workload names) and `inference.inference.svc.cluster.local:8080` is already in the sidecar allowlist.

`agentSessions.piWorkload` is the **revert lever**: set it to `claude-runtime` to put qwen back on the old lane by a values edit, no code deploy. Unset in `deploy/values.yaml` on purpose, so the code default is what ships. Verified by rendering both ways.

## Known soft edge, noted not fixed

`deliver(ember=None, restore_from=<lineage>)` for a qwen lineage created on `claude-runtime` now attempts a cross-workload restore, which the CP denies (workload/principal mismatch is one of its documented `create_denial` reasons). The existing arm already degrades to a blank session with `degraded: "restore_denied"`, so it fails soft and loses only the guest workspace of pre-existing qwen lineages, which are few. That fallback is preserved and tested.

## Verification

`ci test`: `Executed 78 out of 735 tests: 735 tests pass`, plus `1335 tests, 0 failures, 6 skipped` from the genrule suites.

Note for reviewers: an earlier run of this branch **exited 0 while reporting `734 tests pass and 1 fails remotely`**. The failure was a test-isolation defect (`importlib.reload` rebinding `EmberSessionGone` to a fresh class, silently breaking every later `pytest.raises` in the file). It is fixed by asserting on a `_resolve_pi_workload()` helper instead of reloading the module.

### DIFF
--- a/projects/monolith/agent_sessions/__init__.py
+++ b/projects/monolith/agent_sessions/__init__.py
@@ -22,3 +22,22 @@ def model_family(model: str | None) -> str:
     raise ValueError(
         f"Unknown model {model!r}; valid models: opus, sonnet, fable, luna, terra, sol, qwen"
     )
+
+
+# The EmberVM workload names a session lands on, keyed by family. These are a
+# CROSS-SERVICE CONTRACT with the embervm chart's workload CRs
+# (projects/embervm/chart/templates/workload-claude-runtime.yaml,
+# workload-pi-runtime.yaml), not a local label: the noded egress allowlist
+# (EMBERVM_NODED_EGRESS_WORKLOADS) is derived from those same workload names.
+# A name that drifts here does not fail loudly. It dial-timeouts inside the
+# guest, because the egress lane for an unrecognised workload is simply never
+# armed.
+DEFAULT_WORKLOAD = "claude-runtime"
+PI_WORKLOAD = "pi-runtime"
+
+
+def workload_for_family(family: str) -> str:
+    """Return the EmberVM workload name a session of this family creates on."""
+    if family == "pi":
+        return PI_WORKLOAD
+    return DEFAULT_WORKLOAD
--- a/projects/monolith/agent_sessions/__init__.py
+++ b/projects/monolith/agent_sessions/__init__.py
@@ -32,12 +32,17 @@ def model_family(model: str | None) -> str:
 # A name that drifts here does not fail loudly. It dial-timeouts inside the
 # guest, because the egress lane for an unrecognised workload is simply never
 # armed.
+#
+# The other half of the same contract: BOTH workload CRs must stay enabled.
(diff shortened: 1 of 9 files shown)


jomcgi:
## Helm Manifest Diff

Issue #5002, opened 2026-08-16
mcp: Context Forge forwards the caller's Authorization on gateway refresh despite an empty passthrough allowlist


jomcgi:
## Status 2026-08-22

The concrete trigger is gone. The tool-refresh CronJob, and with it the `X-Upstream-Authorization: Discovery anonymous` workaround, was deleted in commit 54df89ae9 (closes #5035): Context Forge's built-in refresh (`AUTO_REFRESH_SERVERS=true`, 600s, #5048) now owns the catalogue, and that path runs from the health loop with no caller request to forward. Nothing in `projects/mcp/context-forge-gateway/chart/` references the header any more.

What remains is the open question in "Done when": which Context Forge code path copies the caller's `Authorization` into an upstream call despite `passthrough_headers = []`, and whether `X-Upstream-Authorization` wins when both are present. That still matters for #4943 (standing workload identities): if the forwarding is intentional, a workload identity presented as the gateway's own auth can be overridden by whatever the caller sent. Any manual `POST /gateways/{id}/tools/refresh` from an authenticated client would also hit it again.

Acceptance: name the file and function in mcp-context-forge 1.0.7 that forwards the header, and record in `projects/mcp/ARCHITECTURE.md` which header wins. If #4943 lands without needing the answer, close this as obsolete.

---

The tool-refresh CronJob's own Context Forge admin JWT reaches the monolith, which rejects it (401 `authentication failed: unrecognized`) and fails discovery. The catalogue was frozen from 2026-08-09 until 2026-08-16 because of it.

## Evidence

- `POST /gateways/{id}/tools/refresh` passes `request_headers=dict(request.headers)` into `refresh_gateway_manually` (`main.py`).
- One manual refresh produced exactly one new `authentication failed: unrecognized` in the live monolith pod. One call in, one rejection out.
- Yet by configuration it should not happen: `gateways.passthrough_headers = []`, global `default_passthrough_headers = ['X-Tenant-Id','X-Trace-Id']`, `enable_overwrite_base_headers = False`, and the gateway's own auth decodes to `{}` (`auth_type='none'`). `get_passthrough_headers` skips its copy loop entirely when `allowed_headers` is empty.
- `remove_headers` is not a lever: it is consumed only by `dataplane_publisher.py`, not by the in-process HTTP path.

So the forwarding is real but its code path is unidentified. Reading the config said it could not happen; measuring the datapath said it does.

## Why it matters

[...]

PR #5004, opened 2026-08-16
fix(mcp): stop tool discovery failing silently and dropping run_code

jomcgi:
The Context Forge tool catalogue was frozen from 2026-08-09 to 2026-08-16, so `run_code` (and `monolith_voice_ui_*`, `screenshot_jomcgi_dev`) were invisible to every MCP caller while the refresh CronJob reported `Complete` every 10 minutes.

## Two independent failures, both of which read green

**1. Discovery 401s, and the Job exits 0 anyway.** Context Forge forwards the caller's `Authorization` to the upstream MCP server, so the admin JWT the CronJob mints for Context Forge's own API also arrived at the monolith, which rejects any bearer it cannot verify. Context Forge answers HTTP 200 with `success: false` in the body, and the Job's only check was `raise_for_status()`.

```
monolith 200 {"success":false,"error":"...Client error '401 Unauthorized'...","toolsAdded":0,...}
```

The monolith logged exactly one `authentication failed: unrecognized` per cron tick, and one more for each manual refresh, which is how the two ends were tied together.

**2. `run_code` is rejected by the gateway's XSS validator.** Its description lists languages as `- javascript: a CommonJS script`, and `(?:^|\s|[\"'\`<>=])(javascript:|...)` reads whitespace followed by `javascript:` as a URI scheme. The tool is dropped while the refresh still reports `success: true`, naming the casualty only in `validationErrors`.

## Changes

- The refresh sends `X-Upstream-Authorization` with a deliberately non-bearer scheme. The monolith ignores a non-bearer scheme and serves discovery anonymously, which it already permits, so the forwarded credential becomes a non-event. #4943 replaces this with a real workload token.
- The CronJob fails on `success: false` and on a non-empty `validationErrors`. Both were true today; both read green.
- `run_code`'s description wraps the language name in backticks, with a comment recording why.
- `projects/mcp/ARCHITECTURE.md` documents both, since neither is recoverable from the manifest.

## Verification

- The edited docstring passes Context Forge's real `SecurityValidator.sanitize_display_text`.
- `X-Upstream-Authorization: Discovery anonymous` was tested against the live gateway before being written into the manifest: `success: true`, 60 tools updated.
- The rendered chart's embedded Python compiles, checked via `helm template` rather than by eye.
- `ci`: `Executed 223 out of 736 tests: 736 tests pass`.

Not fixed here, tracked in #5002: the code path that forwards the caller's `Authorization` is still unidentified. By configuration it should not happen (`passthrough_headers = []`, allowlist without `Authorization`, gateway auth decoding to `{}`), yet it measurably does.

Note the deploy asymmetry: `context-forge-gateway` is a git app on `targetRevision: HEAD`, so the CronJob fix is live at merge, while the monolith description needs the image rebuild and the chart write-back. Refreshes between those two points will fail loudly on `validationErrors`, which is the new check working.


jomcgi:
## Helm Manifest Diff

<details>
<summary><code>context-forge-gateway</code></summary>

```diff

spec.jobTemplate.spec.template.spec.containers.tool-refresh.command  (batch/v1/CronJob/context-forge-gateway-tool-refresh)
  - one list entry removed:
    - |
      set -e
      TOKEN=$(python3 -m mcpgateway.utils.create_jwt_token -u "$PLATFORM_ADMIN_EMAIL" --admin -e 600 | tail -1)
      export TOKEN
      python3 - <<'PY'
      import os, httpx
      base = os.environ["CF_GATEWAY_URL"]
      name = os.environ["CF_GATEWAY_NAME"]
      headers = {"Authorization": "Bearer " + os.environ["TOKEN"]}
      resp = httpx.get(f"{base}/gateways", headers=headers, timeout=30)
      resp.raise_for_status()
      body = resp.json()
      # The API may return a bare list or wrap it; handle both.
      items = body if isinstance(body, list) else body.get("data", body.get("gateways", []))
      gid = next((g["id"] for g in items if g.get("name") == name), None)
      if not gid:
          raise SystemExit(f"gateway {name!r} not found among {len(items)} gateways")
      r = httpx.post(f"{base}/gateways/{gid}/tools/refresh", headers=headers, timeout=90)
      print(name, r.status_code, r.text[:400])
      r.raise_for_status()
      PY
      
    
  
  + one list entry added:
    - |
      set -e
      TOKEN=$(python3 -m mcpgateway.utils.create_jwt_token -u "$PLATFORM_ADMIN_EMAIL" --admin -e 600 | tail -1)
      export TOKEN
      python3 - <<'PY'
      import os, httpx
      base = os.environ["CF_GATEWAY_URL"]
      name = os.environ["CF_GATEWAY_NAME"]
      headers = {"Authorization": "Bearer " + os.environ["TOKEN"]}
      resp = httpx.get(f"{base}/gateways", headers=headers, timeout=30)
      resp.raise_for_status()
      body = resp.json()
      # The API may return a bare list or wrap it; handle both.
      items = body if isinstance(body, list) else body.get("data", body.get("gateways", []))
      gid = next((g["id"] for g in items if g.get("name") == name), None)
      if not gid:
          raise SystemExit(f"gateway {name!r} not found among {len(items)} gateways")
      # CF forwards this request's Authorization to the upstream MCP server,
      # and the monolith 401s any bearer it cannot verify, so discovery fails
      # with the CF-minted admin JWT. X-Upstream-Authorization overrides what
      # the upstream sees and is honoured independently of the passthrough
      # allowlist. The scheme is deliberately NOT Bearer: the monolith ignores
      # a non-bearer scheme and treats discovery as anonymous, which is what it
      # already permits. Swap in a real workload token once authentik service
      # accounts land (#4943).
      refresh_headers = dict(headers, **{"X-Upstream-Authorization": "Discovery anonymous"})
      r = httpx.post(f"{base}/gateways/{gid}/tools/refresh", headers=refresh_headers, timeout=90)
      print(name, r.status_code, r.text[:2000])
      r.raise_for_status()
      # A failed refresh answers 200 with success=false in the BODY, so checking
      # the status alone reports green while the catalogue silently goes stale.
      result = r.json()
      if not result.get("success"):
          raise SystemExit(f"refresh failed: {result.get('error')}")
      # A tool whose description trips CF's XSS validator is dropped from the
      # catalogue while the refresh still reports success overall.
      invalid = result.get("validationErrors") or []
      if invalid:
          raise SystemExit(f"refresh dropped {len(invalid)} tool(s): {invalid}")
      PY
      
    
  
```

</details>

PR #5005, opened 2026-08-16
feat(embervm): enqueue base builds per CPU vendor so coverage converges

jomcgi:
The placement half of #4937. PR #4993 landed the per-vendor bookkeeping and the `BaseVendorCoverage` condition but deliberately changed no build decisions; this arms them.

## The defect

noded keys a base as `sha256(image_ref, workload_revision, cpu_vendor)`, so vendor is part of a base's identity. But `reconcile_desc/2` and `retry_workload/2` both asked "is this built?" against the scalar `snapshot_ref`, which means "some vendor, somewhere, has a base". The moment one vendor succeeded, both short-circuited permanently for a fleet where the other vendor had nothing. Every `EMBER_BASE_EPOCH` string in `deploy/values.yaml` is a manual cache-buster for that guard.

Both branches now repair-enqueue each fleet vendor lacking a current-signature base, through a shared `enqueue_vendor_repairs/3` so the two sites cannot drift.

## What is deliberately NOT changed

**`Ready` stays untouched.** A partially covered workload genuinely serves on the vendor that has a base, so failing `Ready` fleet-wide would turn a degraded state into an outage. `BaseVendorCoverage` carries the gap. There is an explicit assertion for this.

**No status write on the repair path.** Flipping `BaseBuilt` to building for a repair build the caller cannot observe would be a regression, since the workload already serves.

## The open design question from #4937, decided

`vendor_needs_build?/2` does not distinguish *unverified* (observed ref, no CP record) from *missing* (nothing at all). **This enqueues for both**, and that is the correct call rather than a shortcut:

- `vendor_built` is in-memory GenServer state with no rehydration from CRD status, so **every control plane restart wipes attribution**. Verified live after the 0.19.1 roll: **10 of 11** prod workloads read `unverified: amd`, up from 5 earlier the same day. Unverified is the normal post-restart state, not an edge case.
- Skipping it would leave `BaseVendorCoverage` False forever after every restart and repair nothing, which defeats the point of enqueuing at all.
- A redundant enqueue is an RPC, not a bake: `noded/server/server.go` returns `AlreadyBuilt: true` off the existing snapshotRef *before* both `beginBuild` and `driveBuild`.
- It converges: `start_worker/3` stamps `cpu_vendor` from the building node's capacity fact, and an `AlreadyBuilt` response is still `{:ok, %BuildBaseResponse{}}`, so `apply_result/4` writes `vendor_built[vendor]` and the next reconcile stops. There is a test asserting exactly this (a second reconcile issues no further build).

So arming this bakes nothing on today's fleet and fires a handful of already-built RPCs that repair the records and flip coverage True.

## Two things found while reviewing the implementation

**The enqueue target must come from `eligible_build_instances/2`, not `node_of_vendor/3`.** The latter is the *retention* primitive: it scans `representative_facts_by_node` with no eligibility filter and never consults `state.node_ids`, because addressing a vendor's S3 prefix does not need an instance that could host a build. Enqueuing does, twice over: `enqueue/3` does `update_in(state.nodes[id].queue, ...)` which **raises for an instance outside the placement set** and would take the GenServer down, and an instance failing `build_eligible?/2` cannot run the build it is handed, so it fails and the backoff retry re-enqueues it forever. Deriving from the same source `fleet_vendors/2` uses also guarantees every vendor it returns resolves.

**`should_hydrate?/3` needed a vendor guard.** `apply_result/4` sets the scalar ref from whichever vendor built last, and this change makes cross-vendor builds routine rather than rare, so the pinned instance's vendor can now differ from the scalar's. Hydrating an intel-built ref onto an amd instance would be wrong. It now requires the two to match, mirroring what `hydrate_for_anchor/3` already does, and fails closed on a blank vendor.

## Verification

`ci test`: `1340 tests, 0 failures, 6 skipped` from the embervm control plane ExUnit suite (which runs in a genrule and so never appears in the Bazel summary), plus `Executed 2 out of 735 tests: 735 tests pass`.

New tests cover: repair-enqueue on a missing vendor without a status flip, `Ready` still True on partial coverage, full coverage staying a no-op, the unverified-converges-on-AlreadyBuilt property, `retry_workload` enqueuing instead of logging "nothing to do", and the `should_hydrate?` vendor-mismatch guard.

## Follow-up not in scope

`vendor_built` being wiped on every CP restart is worth fixing on its own terms (rehydrating attribution from `status.snapshotRefs` plus a signature check). This PR makes that survivable rather than fixing it.

### DIFF
--- a/projects/embervm/control/lib/embervm/base_builder.ex
+++ b/projects/embervm/control/lib/embervm/base_builder.ex
@@ -755,9 +755,14 @@ defmodule Embervm.BaseBuilder do
         # under N=1 S3 retention (Joe, 2026-07-21) no predecessor ref persists in
         # the store to hydrate from, so a signature->ref history would be dead
         # weight. See docs/plans/2026-07-21-embervm-base-durability-and-cross-vendor.md.
+        #
+        # Hydrate the VENDOR-CORRECT ref (pinned_vendor_ref/3), not the scalar
+        # w.snapshot_ref: should_hydrate?/3's guard 0 already proved this instance's
+        # own vendor has a trustworthy record, so restore that ref rather than
+        # whichever vendor built last.
         state
         |> mark_hydrating(name)
-        |> spawn_hydrate(w)
+        |> spawn_hydrate(%{w | snapshot_ref: pinned_vendor_ref(state, w, node_id)})
 
       w.built_signature == signature(w) and w.snapshot_ref != nil ->
         # The SCALAR base is built, but "some vendor, somewhere" is not the same
@@ -789,11 +794,21 @@ defmodule Embervm.BaseBuilder do
   # rebuilt? True only when ALL five guards hold (fail-safe: any unmet guard falls
   # through to the ordinary build/no-op branches):
   #
-  #   0. the PINNED instance's own reported CPU vendor matches `scalar_vendor`
-  #      (the vendor apply_result/4 last stamped the scalar snapshot_ref with).
-  #      Per-vendor enqueue makes cross-vendor builds routine, so the scalar ref
-  #      can now belong to a vendor other than the one this instance runs; a
-  #      blank/unknown vendor also fails this guard (fail closed);
+  #   0. the PINNED instance resolves to a ref this control plane can vouch for,
+  #      via `pinned_vendor_ref/3`. This anchors on the DURABLE per-vendor CP
+  #      record (`w.vendor_built`), not on comparing the pinned instance's vendor
+  #      against `scalar_vendor` (the vendor `apply_result/4` last stamped the
+  #      SCALAR snapshot_ref with). Per-vendor enqueue makes cross-vendor builds
+  #      the STEADY STATE, not an edge case: `placement/3` pins a workload to ONE
+  #      instance, a repair build by definition runs on a DIFFERENT vendor from
+  #      most workloads' pin, and `apply_result/4` stamps the scalar
+  #      unconditionally while never touching `w.node_id`. A scalar comparison is
+  #      therefore false for nearly every workload in steady state, which would
+  #      disable restore-first fleet-wide rather than guard one cross-vendor
+  #      case. The per-vendor record survives exactly the failure this guard
+  #      exists to recover from (the node losing its local base): it is keyed by
+  #      vendor and does not move just because some OTHER vendor built more
+  #      recently;
   #   1. a base is recorded (snapshot_ref set) AND the spec is unchanged
   #      (built_signature == signature): we are restoring the SAME ref we would
   #      otherwise rebuild, never papering over a spec change (whose target ref is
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

jomcgi:
## Correction round: both blockers fixed

### Finding 1, the hydrate guard, was right and the fix was wrong in an important way

Confirmed the premise against the live fleet: node-1/2/3 are Intel, node-4 is AMD. And the failure does not depend on brick sizing at all. A repair build is *by definition* on the other vendor, so it is always the most recent build, so `apply_result/4` leaves `scalar_vendor` differing from the pinned instance's vendor as the **steady state**. The guard as written would then have been false forever, disabling restore-first fleet-wide, with `vendor_needs_build?/2` reading the record as current so nothing rebuilt either.

The guard's intent was right and its anchor was wrong: it keyed on the scalar, which is precisely what this PR destabilises. `pinned_vendor_ref/3` now resolves the ref from the durable per-vendor CP record (`w.vendor_built[vendor]`, signature-checked), adopting the scalar only when it demonstrably belongs to that vendor, and `reconcile_desc/2` hydrates *that* ref rather than the scalar. That is the mechanism `hydrate_for_anchor/3` already used; the original change mirrored its intent but not its mechanism.

### Finding 2, the cost claim, no longer depends on registration order

`build_instance_of_vendor/3` now ranks by `{base_presence_rank(...), build_rank(...)}`, preferring an instance whose node already reports a base for the workload (READY above present-but-not-ready above absent) and falling back to the previous rank only to break ties. So the enqueue lands where noded's `AlreadyBuilt` short-circuit actually fires, which is what makes "an RPC, not a bake" true rather than a coin flip on dial-home order.

### The rest

- **3:** a failed *repair* build no longer writes `{:failed, ...}` when the workload still serves at the current signature. The backoff retry is still armed; only the status write is suppressed. A genuine first-build or rebuild failure with no current base is unchanged.
- **4:** `base_refs[previous_ref]` now records the **building** instance's node from `worker_meta`, not the pin, so `spawn_evict` addresses the node that actually holds the superseded base.
- **5:** `enqueue_vendor_repairs/3` returns unchanged while the workload is in `state.hydrating`, so a repair build cannot race an in-flight restore onto the same base key.
- **8:** vendor comparison normalised through `to_string |> String.trim`, and `should_hydrate?/3` gained a defensive clause for a nil or non-binary `node_id`. Issue #4105 was exactly that shape.
- **6 and 7:** comments only, no behaviour change, as recommended.

### Tests

The two you flagged as not proving what they read as proving are fixed:

- The "Ready stays True" assertion was a tautology, re-reading a status captured before the repair started. It now forces a real status write during partial coverage (intel joining flips `BaseVendorCoverage` to False, which republishes), asserts the published status actually changed, and only then asserts `Ready` is still True.
- The unverified-converges fixture used `put_vendor_fact`, which produced a wildcard-ranked instance that stole the pin, so the cross-instance path was never taken. It now uses a properly classed brick.

Five tests added: the finding-1 regression (amd-pinned with the scalar moved to intel, an absent report must still hydrate from amd's own record), the finding-2 tie-break with **two instances of the same vendor**, `build_instance_of_vendor/3` returning nil for a vendor whose only instance fails `build_eligible?/2`, the finding-3 failure case, and the finding-5 hydrate race.

`ci test`: `1345 tests, 0 failures, 6 skipped` (1340 baseline plus the 5 new), `Executed 0 out of 736 tests: 736 tests pass`, no `FAILED:` and no `fails remotely` anywhere in the log. Read from the saved log rather than the exit code, since `ci test` exited 0 on a genuinely red run twice today.

PR #5008, opened 2026-08-16
feat(monolith): the voice companion screen, polled off the ledger

jomcgi:
Slice 2 of #4977, following ADR agents/058. Slice 1 (#4996) landed the durable
ledger, the four `voice_ui_*` tools and the companion routes. Nothing rendered
them. This is the screen.

## What is here

- `/private/agents/companion`: attach bar, spoken strip, stage, wire.
- Two SvelteKit proxies for register and ledger poll.
- `stage.js`, a pure reducer over ledger rows, plus 11 unit tests.
- The design mockup from #4977 ported into the repo as `reference-mockup.html`,
  so the visual design is reviewable in a PR rather than living only in an
  account-private artifact.

## Three decisions worth reading

**The reducer is pure and lives outside the component.** ADR 058's argument for
a durable ledger was that it turns "no surface may appear without a ledger row"
from a code review convention into a query. Keeping the fold pure extends that
to the client: a card that cannot be derived from a row cannot be constructed,
and the invariant is unit tested rather than reviewed.

**The browser never calls `/api/agents/*`.** There is no HTTPRoute rule for
that prefix. Such a call falls through the private route's catch-all onto the
SvelteKit port and returns the SPA shell, failing at `JSON.parse` and shipping
green, because nothing in CI exercises a browser-originated path against the
real ingress. Calls go to `/agents/companion`, which the `hooks.js` reroute maps
under `/private`, exactly as the rest of the console already does. No chart
change needed.

**The 2s poll is the liveness heartbeat, not a tuning knob.**
`get_open_voice_ui_companion` treats a companion as open only while
`last_seen_at` is within 90s, and polling is what refreshes it. Stop polling for
90s and every `voice_ui_*` call silently starts no-opping with no error
anywhere. So polling continues while the tab is hidden, re-fires on
`visibilitychange`, and the attach bar reports staleness after 60s rather than
continuing to look attached.

## Deliberately absent

No pin control. Pin means exempt from decay, decay is slice 3, so the button
would be a control wired to nothing. Ask option buttons render inert for the
same reason, with a comment naming the slice that resolves them through
`agent_session_send`.

## Verification

- `ci test`: `Executed 3 out of 736 tests: 736 tests pass`.
- Forced single-target run to prove the new tests actually execute rather than
  serving a cache hit: `ci test //projects/monolith/frontend:test --
  --nocache_test_results` gives `Executed 1 out of 1 test: 1 test passes`, with
  `companion/stage.test.js (11 tests)` in the vitest output.

Both were run before the final rebase onto `d41f5e032`; that rebase brought in
MCP and chart-version changes with no frontend overlap.

Two of the 11 tests are regressions I added during review. Asks were keyed by
`ref`, so a second gate about the same subject silently replaced the first, and
since a user dismissal is remembered per key and no ask key is ever cleared,
dismissing one decision card permanently swallowed every later gate on that ref.
They are now keyed by ledger row id. The pre-existing "multiple asks may
coexist" test used two different refs, so it passed while that case went
untested.

Refs #4977, ADR agents/058

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,6 +89,10 @@ bb-out/
 # for triage, never committed
 .codex-dispatch/
 
+# Same, for grok CLI dispatches. There is no wrapper for those, so the log is
+# whatever the caller tees; it still never belongs in a commit.
+.grok-dispatch/
+
 # Chart versions published by main's push, handed to write-back-versions.sh
 # (ADR platform/009 decision 1). Written by bazel/helm/push.sh.tpl, consumed and
 # deleted in the same CI action, never committed.
--- a/projects/monolith/frontend/src/routes/private/agents/companion/+page.svelte
+++ b/projects/monolith/frontend/src/routes/private/agents/companion/+page.svelte
@@ -1,0 +1,698 @@
+<script>
+  import { onMount } from "svelte";
+  import { periodForHour } from "$lib/private/period.js";
+  import { vmState } from "../status.js";
+  import { applyLedgerRows, dismissCard, emptyStage } from "./stage.js";
+  import "../agents-theme.css";
+
+  const POLL_MS = 2000;
+  const STALE_MS = 60_000;
+  const STORAGE_ID = "voice-companion-id";
+  const STORAGE_SINCE = "voice-companion-since";
+
+  let companionId = $state(null);
+  let since = $state(0);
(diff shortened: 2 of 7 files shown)


jomcgi:
## Helm Manifest Diff

Issue #5007, opened 2026-08-16
moving: spans, milestones and roles are read-only, and gcal_state has no writer


jomcgi:
Follow-up to #4968. The planner shipped with four content tables but write
paths for only one of them, so the dashboard cannot be maintained from the UI.

## 1. Only tasks are writable

`moving/router.py` exposes POST, PATCH, DELETE and done/undone for `/tasks`.
`spans`, `milestones` and `roles` are readable through `GET /api/moving/state`
and writable by nothing. Adding a visitor, shifting the move date or logging a
job application all require a `psql` session against the primary.

The initial content had to be loaded out of band for exactly this reason.

Wanted: create, patch and delete for spans, milestones and roles, following the
ownership rule already established by `_editable_task_or_403` so a viewer
cannot edit rows owned by the other person, plus the UI affordances to reach
them.

## 2. There is no `leave` span kind

`spans.kind` is constrained to `visitor`, `work`, `move`, `trip`, where `work`
renders as "Work trips". Annual leave has no home. Filing time off under `work`
is both mislabelled and actively harmful, because `find_collisions` compares
across kinds and would raise a false collision against the visit the leave was
booked for.

Wanted: a `leave` kind, a `SPAN_KIND_LABELS` entry, and a migration relaxing the
CHECK constraint.

## 3. `gcal_state` is rendered but never written

`+page.svelte` styles milestone markers off `gcal_state` in four places, and
`mergeAgendaItems` surfaces it in every agenda row, but no job in the repo
writes `gcal_state`, `gcal_synced_at` or `gcal_event_id`. Absent a sync job
every milestone displays "queued" forever, so the indicator advertises a
mechanism that does not exist.

Current rows carry hand-set values: `synced` where the milestone genuinely has
a matching Google Calendar event, `queued` where it still needs creating. That
is a stopgap and will drift the moment anything changes on either side.

Wanted: either a real sync job, or drop the columns and the UI that reads them.

## 4. Collisions cannot be acknowledged

`find_collisions` is pure and recomputed per request, so a known and accepted
overlap warns forever. Packing during a visit is deliberate, not a mistake, and
there is no way to say so.

Wanted: a way to dismiss a collision, or a note field on it.

jomcgi:
Direction settled in the 2026-08-29 planning session, work starting now:

1. Write endpoints for spans, milestones and roles: in flight as the next PR, mirroring the task routes and the \`_editable_task_or_403\` ownership rule.
[...]

PR #5010, opened 2026-08-16
fix(monolith): de-collide the moving planner plot labels

jomcgi:
The moving planner timeline positioned every element by date with no
account of the pixel width of its text, so labels collided whenever two
dates sat closer together than their words were wide.

Three symptoms, one cause.

**Bar labels were buried.** `.blbl` was pinned at `left: calc(100% + 9px)`,
always to the right of its bar. An adjacent bar, painting later in DOM
order and carrying a `box-shadow: 0 0 0 2px var(--paper)`, covered the
previous bar's label outright, so only the last bar in a cluster kept its
name. Labels now render beneath their bar, and lanes pack into sub-rows so
a bar that would overlap drops to its own line.

**Leg tags crushed.** They were absolutely positioned at their span's start
date with `white-space: nowrap`, so two spans starting ten days apart on a
four month axis overlaid into mush. Consecutive tags now stagger
vertically. The old `.legtag:first-of-type` padding reset was a partial
hack at the same problem and is replaced rather than built on.

**Milestones were unreadable.** Markers sharing a date rendered
pixel-identical, collapsing three into one, and carried no visible label at
all, only an `aria-label` and a hover tooltip. Milestones now group by date
into one marker per day with a count badge and an always-visible date
label. A group reads as synced only when every milestone on the date is
synced, and as held if any one is held, so a held milestone is never buried
under a marker that looks all clear.

## Countdown

Re-anchored from the move span's `starts_on` to its `ends_on`, so it
measures the day you leave rather than the day packing begins. The hero
drops its trailing copy for the bare number, with the sentence moved to
hover and to an `aria-describedby` target so it is not mouse-only.

## Two things noticed while reading

- The calendar dock rendered `milestones.length` while calling them dates,
  which read wrong as soon as three milestones shared 7 October.
- The scope toggle was labelled `aria-label="Task scope"` when it
  re-scopes milestones, spans and roles as well as tasks.

## Notes

Layout maths lands in `moving.js` as pure exported functions
(`packLaneBars`, `packLegTags`, `groupMilestonesByDate`) with unit tests
covering the real data plus empty, single and tie-breaking cases, rather
than inline in the component where nothing could assert it.

Verified with `ci lint` clean and `ci test` on both
`//projects/monolith/frontend:test` and `:build_test`, run with
`--nocache_test_results` so the result is real execution rather than a
cache hit: `Executed 2 out of 2 tests: 2 tests pass`. The `build_test`
target matters here because `:test` alone does not compile Svelte
templates.

Follow-up gaps in this domain are tracked in #5007.

### DIFF
--- a/projects/monolith/frontend/src/routes/friends/moving/+page.svelte
+++ b/projects/monolith/frontend/src/routes/friends/moving/+page.svelte
@@ -11,8 +11,11 @@
     formatTaskDueDate,
     ganttDatePosition,
     ganttTimeline,
+    groupMilestonesByDate,
     mergeAgendaItems,
     moveCountdown,
+    packLaneBars,
+    packLegTags,
     progressSummary,
     sortOpenTasks,
     sumSellValues,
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

PR #5011, opened 2026-08-16
fix(sandbox): resolve the language binary against a PATH the guest has

### LINKED ISSUE #5009: sandbox: every language fails with "executable file not found in $PATH" because guest-init resolves argv[0] against its own empty PATH
`run_code` is reachable end to end and every language fails at exec:

```
go         -> exec: "go": executable file not found in $PATH
rust       -> exec: "rustc": executable file not found in $PATH
python     -> exec: "python3": executable file not found in $PATH
javascript -> exec: "node": executable file not found in $PATH
```

Four images, four different binaries, one error shape. A per-image packaging fault would not produce that.

## Cause

`guest-init` builds the command like this (`internal/handler/handler.go:285`, and identically in `cmd/main.go:120` for warm-up):

```go
cmd := exec.CommandContext(ctx, argv[0], argv[1:]...)
cmd.Env = spec.Environment(workdir)   // contains PATH=/usr/bin:/bin:/usr/local/bin
```

`exec.CommandContext` resolves `argv[0]` via `LookPath` against the **parent's** `$PATH`, at construction time. Assigning `cmd.Env` afterwards never affects that lookup. Task-class guests get no environment delivery at all (#4429), so guest-init boots with no `PATH` and every lookup fails, while `cmd.Env` correctly carries a `PATH` for a child that is never started.

The apko `environment:` block is already documented as inert for a raw Firecracker boot, with the comment saying guest-init sets these at runtime. It does, but on the wrong process.

## Why nothing caught it

Warm-up runs through the same code and fails the same way, but by design that is non-fatal:

```go
if out, err := cmd.CombinedOutput(); err != nil {
    logger.Warn("language warm-up failed; guest still serves cold", ...)
    return
}
```

Readiness flips regardless, `/shim/ready` returns 200, and BuildBase snapshots. **So a snapshotRef proves the guest booted and served, not that its toolchain ran.** An earlier review of the six-language rollout read the six snapshots as evidence that the Go build, the rustc compile, the BEAM boot and the rest had each executed. That inference was wrong.

`lang_test.go` asserts `Environment()` contains `GOTOOLCHAIN=local` and `GOPROXY=off`. That is a pure data assertion about a slice: it passes whether or not the exec ever resolves.

## Fix

Set the parent's PATH once in guest-init before any exec, or resolve `argv[0]` explicitly against the spec environment instead of the ambient one.

## Done when

One snippet per language runs and returns real stdout, asserted somewhere that fails when the exec does. A test that only inspects the env slice does not qualify.

jomcgi:
Closes #5009.

`run_code` is reachable end to end and every language fails at exec:

```
go         -> exec: "go": executable file not found in $PATH
rust       -> exec: "rustc": executable file not found in $PATH
python     -> exec: "python3": executable file not found in $PATH
javascript -> exec: "node": executable file not found in $PATH
```

Four images, four different binaries, one error shape.

## Cause

`exec.CommandContext` resolves `argv[0]` with `LookPath` against the **calling** process's PATH, at construction time. Assigning `cmd.Env` afterwards never affects that lookup. Firecracker hands PID 1 no environment (#4429), so the PATH guest-init carefully assembled reached only a child that was never started.

The apko `environment:` block already documents itself as inert for a raw Firecracker boot, noting that guest-init sets these at runtime. It did, on the wrong process.

## Fix

`EnsureSearchPath()` puts `SearchPath` into guest-init's own environment, called immediately above the exec in `runCommand` rather than once in `main`, where a refactor could leave it behind. `Environment()` now derives its PATH from the same constant so parent and child cannot drift.

It is **conditional** on PATH being empty, for two reasons: an inherited PATH is deliberate and belongs to whoever set it, and unconditionally overwriting it would point this package's existing exec-dependent tests at an in-guest image layout that does not exist on a test runner.

## Why nothing caught it

A failed warm-up is deliberately non-fatal, so readiness flipped, `/shim/ready` returned 200, and BuildBase snapshotted anyway. Six base snapshots were then read as evidence that the Go build, the rustc compile, the BEAM boot and the rest had each run. They proved only that six guests booted. That log line moves from `Warn` to `Error`, since it is the only signal a guest cannot run its own toolchain.

`lang_test.go` asserted `Environment()` contains `GOTOOLCHAIN=local` and `GOPROXY=off`. That is a claim about a `[]string`, and it passed throughout.

## Verification

The new test drives `runCommand` with PATH cleared and requires real stdout back. Falsified by deleting the fix and re-running:

```
without: FAIL  exec: "sh": executable file not found in $PATH
with:    PASS  stdout "resolved"
```

Identical error shape to production. A second test pins the conditional so it is not "simplified" into an unconditional overwrite.

`ci`: `Executed 1 out of 736 tests: 736 tests pass`.

## Deploy

No epoch bump. The workload pins its guest by digest (`ref: repository@digest`) and the base signature includes that ref, so six rebuilt images re-key and rebuild six warm bases on their own. `EMBER_BASE_EPOCH` exists for changes that do NOT move the digest, such as a new Firecracker binary under an unchanged guest.

Expect the six base rebuilds to land on the masters' scratch, the known pressure point for base rebuilds.

Not done at merge: I will run one snippet per language and paste the real stdout.

### DIFF
--- a/projects/firecracker/sandbox/guest-init/cmd/main.go
+++ b/projects/firecracker/sandbox/guest-init/cmd/main.go
@@ -41,6 +41,15 @@ func run(logger *slog.Logger) error {
 	// across disposable microVMs restored from the one warm-base snapshot.
 	mountTmpfsTmp(logger)
 
+	// Before ANY exec, including the warm-up below. Firecracker hands PID 1 no
+	// environment, and exec.Command resolves argv[0] against this process's
+	// PATH rather than the one in cmd.Env, so without this every language
+	// reports "executable file not found in $PATH" and the warm-up fails
+	// silently because a failed warm-up is deliberately non-fatal.
+	if err := handler.EnsureSearchPath(); err != nil {
+		return fmt.Errorf("ensure sandbox search path: %w", err)
+	}
+
 	spec, err := handler.SelectSpec()
 	if err != nil {
 		return fmt.Errorf("select sandbox language: %w", err)
@@ -121,7 +130,10 @@ func warmup(logger *slog.Logger, spec handler.Spec) {
 	cmd.Dir = "/tmp"
 	cmd.Env = spec.Environment("/tmp")
 	if out, err := cmd.CombinedOutput(); err != nil {
-		logger.Warn("language warm-up failed; guest still serves cold", "language", spec.Name, "err", err, "out", string(out))
+		// Error, not Warn: readiness still flips and the warm base still
+		// snapshots, so this log line is the ONLY signal that a guest cannot
+		// run its own toolchain. It was a Warn while all six languages failed.
+		logger.Error("language warm-up failed; guest still serves cold", "language", spec.Name, "err", err, "out", string(out))
 		return
 	}
 	logger.Info("language warm-up done", "language", spec.Name)
--- a/projects/firecracker/sandbox/guest-init/internal/handler/handler.go
+++ b/projects/firecracker/sandbox/guest-init/internal/handler/handler.go
@@ -282,6 +282,14 @@ func runCommand(ctx context.Context, spec Spec, workdir string, argv []string, s
 	if len(argv) == 0 {
 		return errors.New("language spec has no command")
 	}
+	// Immediately before the exec that depends on it, not once at startup far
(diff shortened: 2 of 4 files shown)


jomcgi:
## Helm Manifest Diff

PR #5012, opened 2026-08-16
fix(sandbox): set GOROOT so go can compile, and put the BEAM in utf8

jomcgi:
Follow-on to #5011, which fixed the PATH fault that hid both of these.

With exec working, five of six languages ran real code:

```
rust    primes below 100000: 9592     last five: [99929, 99961, 99971, 99989, 99991]
python  3.12.14                       mean: 47.521 stdev: 30.3953
node    v24.19.0                      fib(25): 75025
elixir  1.19.5 otp: 28                multiples of 3 or 5 below 1001: 234168
ocaml   5.5.0                         gcd 462 1071 = 21    sum 1..100 = 5050
```

Go failed differently, having found and executed its binary:

```
go: cannot find GOROOT directory: 'go' binary is trimmed and GOROOT is not set
```

## GOROOT

Wolfi builds go with `-trimpath`, so the binary cannot infer its own root, and `GOROOT` was set nowhere: not in the spec `Env`, not in the apko block. It never surfaced because the exec never reached `go` at all.

`/usr/lib/go` is **read from the shipped image**, not assumed:

```
usr/lib/go/bin/go
usr/lib/go/pkg/tool/linux_amd64/compile
usr/lib/go/src/runtime/proc.go
```

Binary, tool chain and stdlib source, so a complete root rather than just where the binary happens to sit.

## Elixir encoding

Every run printed to stderr:

> the VM is running with native name encoding of latin1 which may cause Elixir to malfunction as it expects utf8

The image ships no locale data, so the usual LANG fix is unavailable and `+fnu` is the right lever. Output was correct today, but that warning precedes real misbehaviour the moment a snippet touches a non-ASCII filename.

## What is deliberately NOT here

The apko `environment:` blocks are not updated to match. They describe themselves as documentation only, and the apko lock pins a **checksum of the whole file**, so mirroring two variables into them fails the build until both locks are regenerated, which re-resolves Wolfi and drags package drift into a two-variable fix. I tried it, CI rejected it:

```
Error: locking config: checksum in the lock file does not match the original config:
'projects/firecracker/sandbox/go/apko.yaml' (maybe regenerate the lock file)
```

The rationale lives in `lang.go` comments, where the values are read. `go/apko.yaml` documenting an environment without `GOROOT` is a known gap for whatever PR next has legitimate cause to regenerate that lock.

## Verification

`ci`: `Executed 1 out of 736 tests: 736 tests pass`.

`TestLanguageSpecs` pins the complete env slice with DeepEqual and rejected both additions until its expectations were updated, which is the strict-test tradeoff working as intended. Two focused tests were added alongside so a failure reads as "does not contain GOROOT" rather than a twelve-element slice diff.

Not done at merge: I will run go and elixir after the two bases rebuild and paste real stdout, plus an empty elixir stderr.

### DIFF
--- a/projects/firecracker/sandbox/guest-init/internal/handler/lang.go
+++ b/projects/firecracker/sandbox/guest-init/internal/handler/lang.go
@@ -94,6 +94,12 @@ var languageSpecs = map[string]Spec{
 		Prepare:    prepareGo,
 		Run:        []string{"go", "run", "."},
 		Env: []string{
+			// Wolfi ships the go binary built with -trimpath, so it cannot infer
+			// its own root and exits 2 with "cannot find GOROOT directory: 'go'
+			// binary is trimmed and GOROOT is not set". Verified against the
+			// shipped image, which carries a complete root at this path:
+			// bin/go, pkg/tool/linux_amd64/compile, and src/runtime.
+			"GOROOT=/usr/lib/go",
 			"GOCACHE=/tmp/gocache",
 			"GOPATH=/tmp/gopath",
 			"GOMODCACHE=/tmp/gopath/pkg/mod",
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

PR #5013, opened 2026-08-16
feat(model-bench): add in-cluster qwen3.8-27b xhigh leaderboard row

jomcgi:
## Summary

Adds Qwen3.8-27B to the public LLM leaderboard by grading the model we actually serve (llama.cpp Unsloth Q4_K_M on node-4), not OpenRouter.

- `bench run --base-url` / `--timeout`, plus `api_model` and `extra_body` on the registry, so a filtered run can hit the port-forwarded alias `qwen3.6-27b` with `reasoning_effort: xhigh`.
- `qwen/qwen3.8-27b` is experimental so a bare `bench run` does not send it to OpenRouter. `--model qwen3.8-27b` still sees it.
- Gold tests for `campsites` and `stars` imported `core.db` after the app.db extraction, but both snapshots are pre-extraction. Restored `app.db` so those cells grade.
- Client retries ConnectError / disconnects as well as 429/5xx. The kube API on the LAN dropped the first port-forward mid-pack.

## Result

Qualified. **6/7 hard**, floor 4/4, tool-use ok, $0 (self-hosted). Mean 114k tokens / 10.2 turns / 229s wall. Missed only `go-vsock-frame-01` (ran `go test ./...` without a module). Beats the previous in-cluster row `qwen3.6-27b` (5/7). Old OpenRouter cells were not rerun.

## How to reproduce

```bash
kubectl -n inference port-forward svc/inference 18080:8080
cd projects/model-bench
python3 -m bench snapshot
python3 -m bench run --model qwen3.8-27b --base-url http://127.0.0.1:18080/v1 --timeout 900 --concurrency 1
python3 -m bench report --json-out ../monolith/frontend/src/lib/public/llm-leaderboard/leaderboard.json
```

## Test plan

- [x] Targeted pytest: schema, openrouter, cli, registry (21 passed)
- [x] Live smoke: xhigh thinking + native tool_calls against the port-forward
- [x] Full 14-cell pack; infrastructure fails discarded and rerun
- [ ] PR `ci` / Workflows Test


jomcgi:
## Helm Manifest Diff

PR #5016, opened 2026-08-16
feat(monolith): arm the qwen session synthetic and wire it into health

jomcgi:
Arms `ember-qwen-session-synthetic` (Joe's call, made this session) and closes the gap that would have made arming hollow.

## The gate is satisfied

The prober carried an explicit hold: *"suspended until Joe watches a live run return a completed turn with a non-empty result and confirms that the session is destroyed afterward. Landing a prober you have not watched succeed is how #4074 paged on its own gaps."*

A live production run, driven from the monolith pod with its own ServiceAccount so it exercised the real principal:

| step | result |
|---|---|
| `POST /v1/workloads/pi-runtime/sessions` | `201` in 0.1s |
| `POST /v1/sessions/{id}/invoke` | `200` in 1.8s, `result "qwen synthetic ok"`, `terminal_reason "stop"`, usage 522 in / 27 out (real llama.cpp round trip) |
| `DELETE /v1/sessions/{id}` | `202` |

Completed turn, non-empty result, destroyed session. Routing `qwen` to `pi-runtime` landed in the same session (#5003), so the lane the prober will exercise is the one it is meant to test.

## Arming alone would have been hollow

`probe_qwen` has been writing an `ember_synthetic_probe` latch that **nothing read**. `register_health` carried only `ember_bazel`, `ember_semgrep`, `ember_pages`, `ember_postgres`. So a failing qwen turn was invisible to `/api/health` whether or not the prober ran. This registers `ember_qwen`.

## Why it needs its own staleness window

`EMBER_SYNTHETIC_STALENESS_S = 750.0` is documented in-file as 2.5x the **five minute** cadence of the `ember-synthetic` CronWorkflow. The qwen synthetic is a **separate hourly** CronWorkflow.

Registering it against 750s would mark a perfectly healthy latch stale 12.5 minutes into every hour, so the component would report `prober may be dead` for roughly 80% of the time it is working correctly, and `/api/health` would 503 for most of each hour. `EMBER_QWEN_STALENESS_S = 9000.0` is the same 2.5x rule applied to the cadence this probe actually has.

Two tests cover it: one asserting the constant exceeds an hour and is not the shared value, and an end-to-end one proving a 50-minute-old latch reads OK rather than "prober may be dead".

Note the first-run behaviour is already safe: `synthetic_probe_health` fails **open** on a missing row (`"no probe recorded yet"`), so health stays green between deploy and the first top-of-hour run.

## Comment drift

`piRuntimeWorkload` in the embervm chart claimed `probe_qwen` runs "one trivial turn every five minutes forever". It is hourly, and not part of the `*/5` run. Corrected.

The identical claim in `projects/embervm/runtimes/pi/apko.yaml` is **deliberately left alone**. I tried to fix it and the build failed:

```
Error: locking config: checksum in the lock file ... does not matches
the original config: 'projects/embervm/runtimes/pi/apko.yaml'
```

The apko lock checksums the whole file including comments, so a comment-only edit breaks the image build, and the fix would be a lock regen that refreshes Wolfi packages, moves the image digest and forces a base rebuild. Not a trade worth making for prose.

## Verification

`ci test`: `Executed 66 out of 736 tests: 736 tests pass`, no skips, no `ERROR:`, no `fails remotely`.

Rendered the CronWorkflow to confirm the flag actually lands (the template is gated on the build-time `jobs.image` pin, so a plain `helm template` renders nothing):

```
schedules:
  - "0 * * * *"
concurrencyPolicy: Forbid
suspend: false
```

Reviewer note: the first `ci test` on this branch **exited 0** while reporting `539 tests pass and 197 were skipped`. The skips were downstream of the failed apko action above. An unusual skip count is the tell; judge by comparing against a known good run, not by the exit code.

### DIFF
--- a/projects/embervm/chart/values.yaml
+++ b/projects/embervm/chart/values.yaml
@@ -1624,7 +1624,10 @@ claudeRuntimeWorkload:
 #
 # THE NAMED CONSUMER IS THE SYNTHETIC PROBE. The monolith maps model `qwen` to
 # family `pi` (agent_sessions/__init__.py), and ember_public's probe_qwen runs
-# one trivial turn every five minutes forever. Served on claude-runtime that is
+# one trivial turn HOURLY, from its own ember-qwen-session-synthetic
+# CronWorkflow. It is NOT part of the */5 ember-synthetic run: that one covers
+# bazel, semgrep, pages and postgres only. Read the CronWorkflow rather than
+# this comment if the cadence matters to you. Served on claude-runtime that is
 # a 4096 MiB cold boot of a guest holding a 262 MB Node ELF and a 114 MB codex
 # binary the turn can never invoke. Here it is 256 MiB.
 #
--- a/projects/monolith/chart/values.yaml
+++ b/projects/monolith/chart/values.yaml
@@ -808,18 +808,31 @@ jobs:
 
     # The qwen session synthetic is deliberately separate from the original
     # demo probes so its hourly cadence does not turn a local-model VM test
-    # into a five-minute workload. It is suspended until Joe watches a live
-    # run return a completed turn with a non-empty result and confirms that
-    # the session is destroyed afterward. Landing a prober you have not
-    # watched succeed is how #4074 paged on its own gaps.
+    # into a five-minute workload.
+    #
+    # ARMED 2026-08-16, once the gate below was satisfied by a live run against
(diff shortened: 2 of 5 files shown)


jomcgi:
## Helm Manifest Diff

Issue #5017, opened 2026-08-17
moving planner: leg tag stagger threshold is blind to label width


jomcgi:
`packLegTags()` in `projects/monolith/frontend/src/routes/friends/moving/moving.js` staggers overlapping leg tags using a fixed `LEG_TAG_MIN_GAP_PERCENT = 24`. The function's own comment admits the approximation: tags carry no known text width, so overlap is inferred from how close two tags *start*, never from how wide they are.

That means clearance is luck. "Pack out and leave Vancouver" renders about 185px, roughly 23% of the desktop track, so it clears the 24% threshold with 1% to spare. A slightly longer span label, or the same label at a larger type size, overlaps the next tag on the same stagger row with no warning and no test failure.

Not urgent, and deliberately left alone in the PR that fixed the milestone marker collisions: widening the shell to 1720px makes this strictly *safer*, because tag positions are percentages of the track while label widths are fixed pixels, so a wider track buys more clearance per percentage point.

The fix is to estimate width from label length against a track width in pixels, the way a caller that has measured the plot element could supply it, rather than from position alone. `packLaneBars()` does not have this problem because bars are sized by their date span, so their width is genuinely known.

Found while fixing the near-date milestone marker overprint on the same chart.

PR #5018, opened 2026-08-17
fix(moving): stop milestone markers overprinting and scale up on wide displays

jomcgi:
Two defects on the private moving planner Gantt, both reported from a screenshot of the live page.

## Milestone markers overprinted

`groupMilestonesByDate()` keys its map on the `occurs_on` string, so it merged only milestones on the identical day. Milestones on 7 Oct and 9 Oct stayed two separate markers. At the deployed width the plot track is roughly 800px across 121 days, so two days is about 13px of separation while a `.dia-date` label ("7 Oct", 9.5px monospace, centred) is about 29px wide. The labels overlapped by ~16px and rendered as the unreadable `7 O8t9 Oct`. The `.dia-count` badge compounded it, reaching ~19px right of its own diamond and landing on the neighbour.

Grouping by date equality cannot solve a problem that is about distance on the track, so `clusterMilestoneGroups()` runs as a second, purely visual pass in track-percentage space after positions are known. `groupMilestonesByDate()` keeps its semantic job and its tests unchanged.

Two geometry details are load-bearing:

- **The marker sits on the cluster's earliest date, not the midpoint of its range.** The break test compares first-member positions, so drawing the label there is what makes the `minGapPercent` separation it enforces also true of the labels. A midpoint drifts right by up to half a cluster's span, and in the worst case (a wide cluster followed by a lone marker) closes back to `minGap/2` of its neighbour, which is where this started. It also keeps the diamond on a real date that lines up with the axis. There is a test that pins this property directly and fails under the midpoint version.
- **The gap threshold covers the widest label a cluster can produce**, which is a range like `7-9 Oct` (~45px), not a single date like `16 Nov` (~34px). Clustering introduced a label wider than the threshold was originally sized against.

Nothing is lost: the count badge, the aria-label, the tooltip, and the Calendar dialog all still enumerate every milestone individually, and multi-date tooltips now lead each line with that milestone's own date.

## Page rendered as a small island on wide displays

`.app` capped at `max-width: 1380px` with fixed pixel type throughout, so on a ~3000px viewport it used under half the screen at 9.5px chart labels. One `@media (min-width: 1600px)` raises the cap to 1720px, widens the side columns so the extra width does not all pour into the middle, and steps the chart type up one notch.

**Row heights are deliberately untouched.** The `34px` row constant is written independently in `+page.svelte` (`MILESTONE_ROW_H`, `BAR_ROW_H`, feeding the computed `plotHeight`) and in `moving-theme.css` (`calc(var(--sub-rows, 1) * 34px)`). Scaling one without the other desyncs the plot's total height from the rows drawn inside it, so the wide-viewport pass changes type sizes and the shell width only. Verified: the `34px` count in the CSS is 4 before and 4 after.

## Not in scope

`packLegTags()` still uses a width-blind 24 percent stagger gap, tracked as #5017. Widening the shell makes it strictly safer, since tag positions are percentages while label widths are fixed pixels, so a wider track buys more clearance.

## Verification

- `ci lint` clean.
- `ci test`: `Executed 5 out of 736 tests: 736 tests pass`. The frontend target was necessarily executed rather than cache-hit, since the diff changes `moving.test.js`.
- New tests cover the real 7/9 Oct overlap, the chain-merge guard, held-state precedence across a cluster, determinism, empty input, and the emitted-gap property.

Deploy note: no chart version or `targetRevision` touched, per ADR platform/009.

### DIFF
--- a/projects/monolith/frontend/src/routes/friends/moving/+page.svelte
+++ b/projects/monolith/frontend/src/routes/friends/moving/+page.svelte
@@ -4,6 +4,7 @@
   import { onMount } from "svelte";
   import AccessPanel from "./AccessPanel.svelte";
   import {
+    clusterMilestoneGroups,
     collisionWording,
     formatCad,
     formatDateRange,
@@ -86,16 +87,18 @@
   );
   const milestoneGroups = $derived(groupMilestonesByDate(milestones));
   const milestoneMarkers = $derived(
-    milestoneGroups
-      .map((group) => ({
-        ...group,
-        position: ganttDatePosition(
-          group.occursOn,
-          timeline.startsOn,
-          timeline.endsOn,
-        ),
-      }))
-      .filter((group) => group.position != null),
+    clusterMilestoneGroups(
+      milestoneGroups
+        .map((group) => ({
+          ...group,
+          position: ganttDatePosition(
+            group.occursOn,
+            timeline.startsOn,
+            timeline.endsOn,
+          ),
+        }))
+        .filter((group) => group.position != null),
+    ),
   );
   const legSpans = $derived(
     spans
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

PR #5019, opened 2026-08-17
fix(embervm): navigate shotter to the mapped in-cluster host

jomcgi:
## Problem

`screenshot_jomcgi_dev` fails for **every** input it accepts, returning a guest 502. Confirmed live, not theoretical.

The entire `.dev` TLD is on Chromium's built-in HSTS preload list. So handing Chromium `http://jomcgi.dev` makes it rewrite the URL back to `https://` internally, before any network request, and dial `CONNECT jomcgi.dev:443`. `proxy.go` then correctly refuses CONNECT for a mapped host (the destination is a plaintext `:3000` service and cannot be tunnelled as TLS), the top-level navigation fails as `errNavigationFailed`, and that maps to 502.

Prod evidence, on both reproduction attempts:

```
WARN ember-shotter-init: refusing CONNECT for a mapped host destination=jomcgi.dev:443
```

A `run_code` control probe through the identical CP endpoint returned in 15ms, which rules out the control plane, the endpoint and the network path.

## Why the previous approach could not work

`rewriteToPlaintextHTTP` rewrote the **scheme only** and left the public host in place, deferring host mapping to the proxy (ADR embervm/035 section 3). The host is precisely what triggers the HSTS upgrade, and Chromium decides before the proxy sees a byte. Both hosts this tool accepts are `.dev`, so the strategy had no working input.

## Fix

Resolve the host through the same baked `shotter-egress.json` mapping **inside the guest**, rewriting scheme, host and port together:

```
https://jomcgi.dev/foo
  -> http://monolith-public-frontend.monolith-public.svc.cluster.local:3000/foo
```

Not a `.dev` name, so no preload entry applies, so plaintext, so the existing lane carries it exactly as designed. The in-cluster plaintext hop was never broken and is already allowlisted; only the hostname needed to change.

`final_url` is mapped back to the public host so the response never leaks internal service DNS.

## What is deliberately unchanged

`proxy.go` is untouched. The mapped-host CONNECT guard **stays**: it was written for absolute-`https://` subresources and remains correct for them. It simply stops firing on the top-level navigation.

No sidecar change, no CA, no MITM, no cluster change. An earlier draft went down the TLS-termination route before it became clear the only real constraint was the hostname.

## Verification

- `ci` green: `Executed 1 out of 736 tests: 736 tests pass`.
- Tests cover both mapped hosts, mixed-case hosts, unmapped-host rejection, non-http schemes, and fail-closed behaviour on an empty config.
- The baked-config test now pins `HostMapping` to the constants the rewrite tests assert. The navigate URL is now the mapped destination, so drift would otherwise stay green in CI and surface only as a prod navigation to a stale service.

## Not proven by CI, by construction

Nothing here exercises a real navigation. `trialRender` uses `about:blank`, dev has the workload disabled (`dev/deploy/values.yaml:117`) and could not run it anyway at ~768 MiB guest-schedulable against shotter's 2048 MiB, and no target depends on the launch path. **Prod invocation is the only verifier.** I will call the tool after rollout rather than infer from readiness.

One known residual risk: any subresource hardcoded as an absolute `https://jomcgi.dev/...` URL in the markup will still hit the CONNECT guard and fail to load. Host-relative and Host-derived URLs are fine. That is a render-time question the first real capture answers.

### DIFF
--- a/projects/embervm/runtimes/shotter/guest-init/cmd/main.go
+++ b/projects/embervm/runtimes/shotter/guest-init/cmd/main.go
@@ -97,7 +97,7 @@ func run(logger *slog.Logger) error {
 	}
 
 	var ready atomic.Bool
-	serveErr := startVsockServer(ctx, logger, ready.Load)
+	serveErr := startVsockServer(ctx, logger, ready.Load, proxyConfig)
 
 	if proxyConfigErr != nil {
 		// A warm failure never flips readiness (see the package doc above).
@@ -303,7 +303,7 @@ func probeCDP(ctx context.Context, client *http.Client, url string) (cdpVersion,
 	return version, nil
 }
 
-func newMux(ready func() bool, logger *slog.Logger) *http.ServeMux {
+func newMux(ready func() bool, logger *slog.Logger, config ProxyConfig) *http.ServeMux {
 	mux := http.NewServeMux()
 	mux.HandleFunc("GET /shim/healthz", func(w http.ResponseWriter, _ *http.Request) {
 		w.WriteHeader(http.StatusOK)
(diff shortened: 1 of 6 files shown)


jomcgi:
## Helm Manifest Diff

PR #5025, opened 2026-08-17
feat(inference): expose the OpenAI API at private.jomcgi.dev/llm/v1

jomcgi:
## What

A private-tier HTTPRoute plus the standard Cloudflare Access SecurityPolicy, so an OpenAI-compatible client outside the cluster can call the llama.cpp server:

```
base_url = https://private.jomcgi.dev/llm/v1
model    = qwen3.6-27b     # llamacpp.extraArgs --alias; actually Qwen3.8-27B Q4_K_M
```

## Why this hostname

`friends.jomcgi.dev` was the alternative and does not work for this caller. Every authentik lane on that host runs an Envoy `oidc` block, which answers an unauthenticated request with a browser redirect an SDK cannot follow. Reaching it headlessly would have meant a new authentik `client_credentials` provider minting short-lived JWTs pasted in as the "api key".

Cloudflare Access already admits non-browser callers natively: a Service Auth policy accepts `CF-Access-Client-Id` / `CF-Access-Client-Secret` and injects the `Cf-Access-Jwt-Assertion` header the SecurityPolicy validates. So the cluster side is the same two objects every other private route already ships.

## Design notes

- **Match `/llm/v1`, not `/llm`.** Rewriting the bare prefix to `/` would publish everything llama-server routes, including `/props` and the `/metrics` endpoint `--metrics` turns on. `/slots` is not in play either way, since `--slots` is not passed.
- **Explicit `timeouts: 600s`.** `cf-ingress.httproute` emits no timeouts, so Envoy's 15s route default would apply and truncate any real generation. This route therefore writes its own HTTPRoute and takes only the SecurityPolicy from the library. 600s matches the monolith's private route against this same backend.
- **Not under `/app/`**, which on this hostname is the convention for browser UIs.
- **No `audiences` pin**, matching argocd/signoz/longhorn/kargo. Stated in the template: isolation between "this token may call the LLM" and "this token may call argocd" lives entirely in the Cloudflare Access policy, not in the cluster.

## Safe on merge with no Cloudflare change

The Access application covering `private.jomcgi.dev` has no path scope, so it guards `/llm` the moment the route exists: a browser needs a team login, a headless client gets nothing. Adding a path-scoped Access app for `/llm` with a Service Auth policy is the separate, dashboard-side step that later admits a service token. Note such an app *replaces* the hostname-wide one for that path rather than layering on it, so a Service-Auth-only policy locks the browser out of `/llm`.

## Verification

- `helm template` renders exactly the two new objects, 9 total (was 7).
- `cf-ingress` is a `type: library` chart, so vendoring it adds no objects of its own.
- Deploy is instant on merge: `projects/inference/deploy` is a git-sourced Application on `targetRevision: HEAD`, so no chart-version write-back is involved.
- Post-merge: confirm `/llm/v1/models` returns the model list through the edge, and that `/app/argocd` and the other private routes are unchanged.

Cloudflare 524s if the origin sends no byte for 100s, which is not configurable here, so long non-streamed completions are the shape that breaks. Prefer `stream: true`.


jomcgi:
## Helm Manifest Diff

<details>

```diff

(root level)  (gateway.networking.k8s.io/v1/HTTPRoute/inference-private)
+ one document added:
  ---
  # Source: inference/templates/httproute-private.yaml
  apiVersion: gateway.networking.k8s.io/v1
  kind: HTTPRoute
  metadata:
    name: inference-private
    labels:
      helm.sh/chart: inference-2.5.4
      app.kubernetes.io/name: inference
      app.kubernetes.io/instance: inference
      app.kubernetes.io/version: b5270
      app.kubernetes.io/managed-by: Helm
      ingress-tier: trusted
  spec:
    # group/kind and the backendRef group/kind/weight below are Gateway API schema
  # defaults, spelled out to match cf-ingress.httproute rather than because this
  # Application needs it: that library declares them because its ServerSideApply
  # consumers report OutOfSync forever on any field the chart leaves implicit
  # and the apiserver then defaults. This app is client-side apply today, so the
  # cost of stating them is zero and the trap is pre-closed if that changes.
  parentRefs:
    - name: cloudflare-ingress
      group: gateway.networking.k8s.io
      kind: Gateway
      namespace: envoy-gateway-system
    hostnames:
    - private.jomcgi.dev
    rules:
    - matches:
      - path:
          type: PathPrefix
          value: /llm/v1
      filters:
      - type: URLRewrite
        urlRewrite:
          path:
            type: ReplacePrefixMatch
            replacePrefixMatch: /v1
      # Explicit, because this is NOT rendered through cf-ingress.httproute and
  # that library emits no timeouts at all: Envoy's own route default is 15s,
  # which truncates any generation longer than a sentence. 600s matches the
  # monolith's private route, whose callsites hit this same backend.
  #
  # Cloudflare is the other clock and it is not configurable here: the edge
  # returns 524 if the origin sends NO byte for 100s. Streaming responses
  # emit the first token well inside that, so a long non-streamed completion
  # is the shape that breaks. Prefer stream: true from the client.
  timeouts:
        request: 600s
        backendRequest: 600s
      backendRefs:
      - name: inference
        group: 
        kind: Service
        port: 8080
        weight: 1
  


(root level)  (gateway.envoyproxy.io/v1alpha1/SecurityPolicy/inference-private-cf-access)
+ one document added:
  ---
  # Source: inference/templates/httproute-private.yaml
  apiVersion: gateway.envoyproxy.io/v1alpha1
  kind: SecurityPolicy
  metadata:
    name: inference-private-cf-access
  spec:
    targetRefs:
    - name: inference-private
      group: gateway.networking.k8s.io
      kind: HTTPRoute
    jwt:
      providers:
      - name: cloudflare-access
        issuer: "https://jomcgi.cloudflareaccess.com"
        extractFrom:
          headers:
          - name: Cf-Access-Jwt-Assertion
        remoteJWKS:
          # Envoy Gateway's own default, declared explicitly for the same
  # ServerSideApply field-ownership reason as the HTTPRoute defaults.
  # targetRefs above already spelled its group/kind out and showed no
  # drift while this key did, on the same object, which is what
  # identified the mechanism.
  cacheDuration: 300s
          uri: "https://jomcgi.cloudflareaccess.com/cdn-cgi/access/certs"
        claimToHeaders:
        - claim: email
          header: X-Auth-Email
```

</details>

PR #5028, opened 2026-08-18
fix(inference,model-bench): let a Cloudflare service token reach the LLM

jomcgi:
Two blockers stood between the work laptop and `private.jomcgi.dev/llm/v1`. Both are fixed here.

## 1. Envoy rejected the request

Cloudflare Access injects `Cf-Access-Jwt-Assertion` for **identity-based** sessions. A **service-token** request is authenticated by Access and forwarded **without** it. Measured, not assumed: with the token the edge stopped returning its own 403 and Envoy answered

```
HTTP/2 401
www-authenticate: Bearer realm="http://private.jomcgi.dev/llm/v1/models"
Jwt is missing
```

That string is Envoy's JWT filter verbatim, so the request had cleared the edge and died at the origin. Every other private-tier route is a browser lane, so this is the first one to carry a machine caller and the first to hit it.

The SecurityPolicy is now written out in `httproute-private.yaml` rather than taken from `cf-ingress.security-policy`, for the one field that library cannot express: **`jwt.optional: true`**. Object name is unchanged, so this updates in place.

What that changes precisely: a request with **no** JWT passes. A request **with** one is still fully validated (signature, issuer, expiry) and still projects its email claim. This is not "auth off", it is "this auth applies to callers holding that kind of credential", and the service token is a different kind that Cloudflare has already checked.

What guards the route is Cloudflare Access, which was already true in substance: the policy has no `authorization` block and no `audiences`, so it never decided *who* could call, only that a caller arrived through the edge. That last check is what is given up for tokenless callers. Residual exposure is something inside the cluster addressing the Envoy gateway directly with a `private.jomcgi.dev` Host header, since nothing else reaches this listener except the cloudflared tunnel.

## 2. The harness could not authenticate even once the route allowed it

`OpenRouterClient.__init__` set `headers={"Authorization": ...}` and took nothing else, so there was no way to send `CF-Access-Client-Id` / `CF-Access-Client-Secret`, the pair Access actually checks. Adds a repeatable `--header NAME:VALUE` threaded through as `extra_headers`.

`_parse_headers` splits on the **first** colon so a value containing one survives, and **raises** on a malformed entry rather than skipping it: a silently dropped secret presents as a 401 from the edge, which reads like a credential problem rather than a typo in the flag.

## Verification

`helm lint --strict` passes, policy renders with `optional: true` under the unchanged name `inference-private-cf-access`. `ci test`: `Executed 15 out of 735 tests: 735 tests pass`.

The header test uses `MockTransport` to assert both CF headers are **on the outbound request** alongside `Authorization`, rather than merely accepted by the constructor. A test that only checked the constructor would pass against a client that swallowed them.

Usage from off-cluster:

```
bench run --base-url https://private.jomcgi.dev/llm/v1 \
  --header "CF-Access-Client-Id: $CF_ID" \
  --header "CF-Access-Client-Secret: $CF_SECRET" \
  --model qwen3.8-27b
```

### DIFF
--- a/projects/inference/deploy/templates/httproute-private.yaml
+++ b/projects/inference/deploy/templates/httproute-private.yaml
@@ -82,16 +82,70 @@ separator onto the first line of the policy, which helm lint reports as
 "invalid Yaml document separator". The same shape exists unlinted elsewhere in
 the repo, because every other chart consuming this library sets lint = False.
 
-Validates the Cloudflare Access JWT, so the only way to this route is through
-the edge rather than by reaching Envoy some other way.
+Validates the Cloudflare Access JWT when one is present.
 
-This carries NO `authorization` block and NO `audiences`, matching every other
-private-tier route (argocd, signoz, longhorn, kargo). The consequence is worth
-stating plainly: any JWT from the team issuer validates here, so the isolation
-between "this service token may call the LLM" and "this service token may call
-argocd" lives ENTIRELY in the Cloudflare Access policy, not in the cluster. Pin
-`audiences` to the /llm application's AUD tag if that ever needs to hold on its
-own.
+WRITTEN OUT RATHER THAN TAKEN FROM cf-ingress.security-policy, for the one
+field that library cannot express: `optional: true`.
+
+WHY IT IS NEEDED. Cloudflare Access injects Cf-Access-Jwt-Assertion for
+IDENTITY-based sessions. A SERVICE TOKEN request is authenticated by Access and
+forwarded WITHOUT that header, which was measured, not assumed: with the token
+the edge stopped returning its own 403 and Envoy answered
+
+    HTTP/2 401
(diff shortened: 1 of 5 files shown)


jomcgi:
## Helm Manifest Diff

<details>

```diff

metadata  (gateway.envoyproxy.io/v1alpha1/SecurityPolicy/inference-private-cf-access)
  + one map entry added:
    labels:
      helm.sh/chart: inference-2.5.4
      app.kubernetes.io/name: inference
      app.kubernetes.io/instance: inference
      app.kubernetes.io/version: b5270
      app.kubernetes.io/managed-by: Helm

spec.jwt  (gateway.envoyproxy.io/v1alpha1/SecurityPolicy/inference-private-cf-access)
  + one map entry added:
    optional: true
```

</details>

PR #5030, opened 2026-08-18
feat(inference): give bench mode a 160k context window

jomcgi:
Bench mode now runs a **163840-token** KV pool. With one slot that is the whole window for a single sequence, against the model's 262144 training context.

## Sized from measurement, not estimate

```
(1 slot,  98304 KV)  = 20942 MiB    measured 2026-08-18
(3 slots, 98304 KV)  = 21590 MiB    measured 2026-08-18
(4 slots, 131072 KV) = 22562 MiB    measured 2026-08-14, in values.yaml
```

Two unknowns, three equations. Solving on the first and third:

```
fixed base     18674 MiB    (weights + mmproj + non-slot compute)
per slot         324 MiB
per 32768 KV     648 MiB    at q8_0
```

That model then predicts the middle point at exactly **21590**, the measured value, so it has no residual across the range.

| ctxSize at 1 slot | used | free |
|---|---|---|
| 98304 (current) | 20942 | 3140 |
| 131072 | 21590 | 2974 |
| **163840** | **22238** | **2326** |
| 196608 | 22886 | 1678 |
| 262144 | 24182 | 382 (too tight) |

163840 leaves 2326 MiB. Production has run all week at 2974, so this is only marginally tighter than normal steady state.

**The figure this replaces claimed 1.00 GiB per 32k, a 35% overestimate.** That is why 160k looked marginal in every earlier calculation in this repo. The derivation now lives in the `llamacpp.ctxSize` comment so the next person recomputes from constants rather than reusing a rule of thumb.

## Why the override is not just an edit to llamacpp.ctxSize

It lives in `benchMode.ctxSize` behind a helper, mirroring how `parallel` is handled. Left behind in production, a direct edit would run 3 slots against a pool sized for one, quietly tightening VRAM below the headroom the production value was chosen for. One flag still restores everything.

## Verification

Rendered both ways: `true` gives `--ctx-size "163840" --parallel "1"`, `false` gives `--ctx-size "98304" --parallel "3"` unchanged. Embeddings untouched at 8192.

Post-merge check is VRAM against the predicted 22238 MiB, which is the real test of the model above. The pod rolls (`Recreate`, single GPU), so qwen is down for the model load.

### DIFF
--- a/projects/inference/deploy/templates/_helpers.tpl
+++ b/projects/inference/deploy/templates/_helpers.tpl
@@ -102,6 +102,20 @@ callers still connected they would queue behind a full generation, up to about
 {{- end }}
 
 {{/*
+Total KV pool in tokens. Bench mode overrides it here rather than by editing
+llamacpp.ctxSize, so the single benchMode flag still restores everything: an
+override left behind in production would run 3 slots against a pool sized for
+one, quietly tightening VRAM below the headroom that value was chosen for.
+
+Sizing is solved from three measured points rather than estimated. See the
+llamacpp.ctxSize comment in values.yaml for the derivation and for why the
+GiB-per-32k rule of thumb it replaced was 35% too pessimistic.
+*/}}
+{{- define "inference.llamaCppCtxSize" -}}
+{{- if .Values.benchMode.enabled -}}{{ .Values.benchMode.ctxSize }}{{- else -}}{{ .Values.llamacpp.ctxSize }}{{- end -}}
+{{- end }}
+
+{{/*
 Embedding llama-server CLI arguments.
 */}}
 {{- define "inference.embeddingArgs" -}}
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

<details>

```diff

spec.template.spec.containers.llama-server.args.0  (apps/v1/Deployment/inference)
  ± value change in multiline text (one insert, one deletion)
      MODEL=$(find /model-image -maxdepth 2 -name '*.gguf' ! -name 'mmproj*' -type f | sort | head -1)
      if [ -z "$MODEL" ]; then
        echo "ERROR: no .gguf file found under /model-image" >&2
        exit 1
      
      [13 lines unchanged)]
      
      echo "Auto-discovered mmproj: $MMPROJ"
      set -- "$@" --mmproj "$MMPROJ"
      exec /app/llama-server "$@" \
        --n-gpu-layers "99" \
    -   --ctx-size "98304" \
    +   --ctx-size "163840" \
        --parallel "1" \
        --flash-attn "on" \
        --cache-type-k "q8_0" \
        --cache-type-v "q8_0" \
      
      [five lines unchanged)]
      
        "--jinja" \
        "--reasoning-format" \
        "deepseek" \
        "--metrics"
  
  
```

</details>

PR #5031, opened 2026-08-18
fix(inference): back bench context off 160k to 128k

jomcgi:
163840 runs, but it landed at **23426 of 24564 MiB with 656 MiB free**, against a predicted 2326.

656 MiB leaves no cushion for an allocation spike, and this is a `Recreate` deployment on a single GPU, so an OOM is a crash loop with the previous engine already torn down rather than a degraded pod. Not something to leave running unattended.

131072 sits **between** two measured points, so its estimate is interpolation rather than extrapolation:

```
(1 slot,  98304) = 20942 MiB   measured
(1 slot, 131072) = ~22176      interpolated, ~2388 free
(1 slot, 163840) = 23426 MiB   measured
```

## The sizing comment, which this session made worse

It previously carried a `1.00 GiB per 32k` rule of thumb. Earlier today I replaced that with a 648 MiB figure derived from a linear fit to three points spanning 98304 to 131072, called the old number "a 35% overestimate", and shipped 163840 on the strength of it.

The fit was exact inside its range and **missed 163840 by 1188 MiB**. The real cost between 98304 and 163840 is roughly 1234 MiB per 32768, about double, because KV is not the only term that grows with `n_ctx`: llama.cpp's attention and compute buffers scale too. The discarded rule of thumb was the more accurate of the two in the range that actually mattered.

The comment now records the fourth measured point, states plainly that cost is **not linear** in `ctxSize`, and says to interpolate between measurements rather than extrapolate past them, verifying against `nvidia-smi` after any change.

## Verification

Renders `--ctx-size "131072" --parallel "1"` with bench mode on, `--ctx-size "98304" --parallel "3"` off, unchanged. `helm lint --strict` passes. `ci test`: `Executed 1 out of 735 tests: 735 tests pass`.

Post-merge check is `nvidia-smi` against the ~22176 estimate, which is the whole point this time.

### DIFF
--- a/projects/inference/deploy/values.yaml
+++ b/projects/inference/deploy/values.yaml
@@ -260,25 +260,33 @@ llamacpp:
   # per-sequence like vLLM's --max-model-len. 131072 / 4 = 32768 per slot, which
   # preserves the 32k window every monolith callsite is written against.
   #
-  # VRAM COST, SOLVED FROM THREE MEASURED POINTS on the 24564 MiB card rather
-  # than estimated:
+  # VRAM COST IS NOT LINEAR IN ctxSize. Do not fit a line to it. Measured on the
+  # 24564 MiB card:
   #
-  #   (1 slot,  98304 KV) = 20942 MiB   measured 2026-08-18
-  #   (3 slots, 98304 KV) = 21590 MiB   measured 2026-08-18
-  #   (4 slots, 131072 KV)= 22562 MiB   measured 2026-08-14
+  #   (1 slot,  98304)  = 20942 MiB   2026-08-18
+  #   (3 slots, 98304)  = 21590 MiB   2026-08-18
+  #   (4 slots, 131072) = 22562 MiB   2026-08-14
+  #   (1 slot, 163840)  = 23426 MiB   2026-08-18
   #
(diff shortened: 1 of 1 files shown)


jomcgi:
## Helm Manifest Diff

<details>

```diff

spec.template.spec.containers.llama-server.args.0  (apps/v1/Deployment/inference)
  ± value change in multiline text (one insert, one deletion)
      MODEL=$(find /model-image -maxdepth 2 -name '*.gguf' ! -name 'mmproj*' -type f | sort | head -1)
      if [ -z "$MODEL" ]; then
        echo "ERROR: no .gguf file found under /model-image" >&2
        exit 1
      
      [13 lines unchanged)]
      
      echo "Auto-discovered mmproj: $MMPROJ"
      set -- "$@" --mmproj "$MMPROJ"
      exec /app/llama-server "$@" \
        --n-gpu-layers "99" \
    -   --ctx-size "163840" \
    +   --ctx-size "131072" \
        --parallel "1" \
        --flash-attn "on" \
        --cache-type-k "q8_0" \
        --cache-type-v "q8_0" \
      
      [five lines unchanged)]
      
        "--jinja" \
        "--reasoning-format" \
        "deepseek" \
        "--metrics"
  
  
```

</details>

Issue #4667, opened 2026-08-11
docs: reconcile ADRs into per-domain ARCHITECTURE.md, then drop the harvested ADRs


jomcgi:
## Problem

The ADR tree is 141 files and 303,661 words, and for 13 of 14 domains it is the
*only* narrative document that exists. An agent asking "how does X work today"
has nowhere to go except a decision log written at a point in time, so it reads
a stale ADR and treats it as current state.

Status hygiene alone does not fix this. `agents/020` ("Deprecate Context Forge,
Serve MCP Directly from the Monolith") is **Accepted** and correctly so, the
decision genuinely was accepted. It was then half-executed: the monolith side
was built, Context Forge was never removed, and it is now the agent gateway.
Nothing records that. Three sources describe it three different ways (the ADR,
CLAUDE.md, reality).

The drift is not confined to `docs/decisions/`. `check_readme_structure.py`,
whose entire job is catching structural staleness, currently passes while
asserting that `projects/grimoire` exists (it does not) and that EmberVM is an
"R0 skeleton, not yet a deployed system" (it is the largest deployed system in
the repo).

## Approach

Per-domain **reconciliation pass**, not a documentation task. Read what the
ADRs, CLAUDE.md, values-file comments and CI guards claim; read what the charts
and the cluster actually do; put every disagreement in front of Joe as a
confirm-or-deny ledger. Confirmed claims become `ARCHITECTURE.md` prose. Denied
claims fan out into ADR supersessions, CLAUDE.md edits, and issues.

`projects/embervm/ARCHITECTURE.md` is the working template. Pilot ledger for
`projects/mcp` is complete and adjudicated.

## Standing policy

1. **Link to ARCHITECTURE.md, never to an ADR.** Nine live code references
   currently point at ADR paths; each gets repointed during its domain's rollup.
2. **Delete per-domain.** An ADR is dropped only once its domain's
   ARCHITECTURE.md carries the current state and inbound links are repointed.
   Deletion is the last step of a rollup, never a standalone sweep. Git history
   is the archive.
3. **Keep writing ADRs.** They stay the full-rationale decision journal, but
   write-and-forget: nothing links to them, the rollup harvests them, and they
   are deleted once harvested.

## Inventory as found

14 tracked top-level domains. 9 have a README, 1 has an ARCHITECTURE.md.

| Domain | README | ARCHITECTURE | Note |
|---|---|---|---|
| embervm | yes | yes | the template |
[...]

PR #5167, opened 2026-08-23
docs(inference): add a README as the domain entry point

jomcgi:
## What

Rollup pass for `projects/inference` (#4667). The domain had no README; this adds one as the entry point and corrects the `docs/reference/services.md` row (said vLLM, live engine is llama.cpp).

## Runbook 5a verdict: README, not ARCHITECTURE.md

- No ADR is specific to this domain; the ADRs that mention it (agents/012, 013, 021, 024, platform/010, security/005) are about its consumers and are claimed by the monolith and platform rollups. Test 1 (partially executed or reversed ADR trail) fails.
- The consumer wiring does span several files (monolith chart and deploy values, monolith-dev, monolith-public, embervm egress allowlist, `shim.py`, model-bench), but the domain's own knowledge lives in one heavily commented `values.yaml` and its templates, and those comments are fresh (Joe, 2026-08-17 to 2026-08-21).
- PR #5162 (#5155, NInfer as the only engine) is open against this directory and deletes the engine switch. A current-state document written now would be stale on arrival. The README keeps the engine section short and points at #5162; layout, consumer map, ingress and bench mode survive the swap.

Recommend recording inference as README-only on #4667 and re-evaluating after #5162 lands.

## Ledger

| Claim | Source (date) | Observed | Verdict |
| --- | --- | --- | --- |
| LLM engine is llama.cpp serving Qwen3.8-27B GGUF under alias `qwen3.6-27b` | `deploy/values.yaml` (2026-08-21) | Live pod args: `--alias qwen3.6-27b --ctx-size 98304 --parallel 3 --spec-type draft-mtp`; logs show model, mmproj-F16 and MTP draft discovered | verified |
| Inference is "Self-hosted LLM inference (vLLM)" | `docs/reference/services.md` (2026-08-07) | stale | fixed in this PR |
| Chart deploys via OCI publish and write-back like other services | CLAUDE.md general rule | `application.yaml` is a git-path source at `targetRevision: HEAD`; chart absent from `//bazel/images:push_charts`; `Chart.yaml` version last touched by hand 2026-08-17. Deploys on merge, like `projects/platform/*` | verified, documented |
| Chart.yaml sits inside `deploy/` against convention | #4667 task | confirmed; `deploy/BUILD` also exports `values.yaml` to `//projects/embervm/runtimes/claude` | documented, not restructured |
| `inference-chat-template` ConfigMap and `inference-vllm-cache` PVC render unconditionally | templates | both live (121d / 122d), consumed by nothing while `llm.engine: llamacpp`; values comment says the vLLM block is kept intact for the revert | deliberately inert; #5162 deletes them |
| `pi_context_window_sync_test.py` guards `PI_CONTEXT_WINDOW` against the engine's context | test docstring | reads `vllm.maxModelLen` (32768), a key the live engine does not consume; equal to 98304/3 by coincidence | inert guard; already item 4 on #5155, fixed in #5162 |
| External API at `private.jomcgi.dev/llm/v1` behind Cloudflare Access service token | `httproute-private.yaml` (2026-08-21) | HTTPRoute and SecurityPolicy (`jwt.optional: true`) live 5d | verified |
| Bench mode off, Service is ClusterIP | `values.yaml` | `service/inference` ClusterIP, 47h old | verified |
| Qwen synthetic never pages, `/health` has no inference dependency | `values.yaml` comment (verified live 2026-08-18) | not re-verified here | carried as stated |
| embervm allowlist comment says "in-cluster vLLM endpoint"; monolith-dev comment says "Qwen vLLM" | `embervm/chart/values.yaml:1521`, `monolith/deploy/values.yaml:290` | wording stale, mechanism correct | harmless, left alone |

Docs-only, pushed with `SKIP_CI_TEST=1` after `ci lint` (prettier, 2 files, done). No auto-merge.

Refs #4667, #5155.

### DIFF
--- a/docs/reference/services.md
+++ b/docs/reference/services.md
@@ -33,7 +33,7 @@ This document provides an overview of all services running in the cluster.
 | **Monolith**           | Primary application backend and frontend    | [projects/monolith](../../projects/monolith/)                                       |
 | **Monolith Public**    | Read-only public tier serving jomcgi.dev    | [projects/monolith-public](../../projects/monolith-public/)                         |
 | **EmberVM**            | Firecracker microVM orchestration for agent sandboxes | [projects/embervm](../../projects/embervm/)                               |
-| **Inference**          | Self-hosted LLM inference (vLLM)            | [projects/inference](../../projects/inference/)                                     |
+| **Inference**          | Self-hosted LLM inference (llama.cpp LLM + CPU embeddings) | [projects/inference](../../projects/inference/README.md)                      |
 | **Context Forge Gateway** | MCP gateway for the GitHub and monolith tool surfaces | [projects/mcp/context-forge-gateway](../../projects/mcp/context-forge-gateway/) |
 
 ## Development Services (dev)
--- a/projects/inference/README.md
+++ b/projects/inference/README.md
@@ -1,0 +1,172 @@
+# Inference
+
+On-cluster LLM serving for everything in this repo that calls a model and
+does not want to pay per token: the monolith's Discord, chat, summarize,
+vision and classifier callsites, monolith-dev (which points every provider
+here), monolith-public retrieval, EmberVM pi-runtime qwen sessions, and the
+`model-bench` harness. One GPU (the RTX 4090 on node-4), one generative
+model, one CPU embeddings model.
+
+This is the entry point for the domain. The detailed reasoning lives in the
+values-file comments, which are deliberately long; this file says where to
+look and what holds across the whole directory.
+
+## What runs
+
+| Workload | Engine | Model | Where |
+| --- | --- | --- | --- |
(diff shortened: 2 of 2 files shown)


jomcgi:
Closing as superseded rather than rebasing.

#5162 merged about two hours after this branch was cut and removed the exact machinery this README documents: the `llm.engine` switch, the vLLM templates, the `llamacpp.*` keys, and the ConfigMap/PVC pair. Main's `projects/inference/deploy/values.yaml` now opens with "The LLM engine is NInfer". Separately, a1cba1655 inverted the health claim: the composite `/health` now fails fatally when inference is down, where this README says it carries no inference dependency.

Four of the eight sections describe deleted mechanism, so this is a rewrite rather than a rebase. The underlying gap is real and stays open: `projects/inference` still has no README, and `docs/reference/services.md:35` still says vLLM.

Follow-up filed as #5400, which lists the sections worth salvaging from this branch.

Issue #5221, opened 2026-08-23
embervm: backfill image-lane serving inventory on daemon boot, and select activators that hold the image


jomcgi:
Two follow-ups surfaced by the #5180 serving-prover drill (round 3), both downstream of ADR embervm/038:

1. Boot-time backfill. Image-lane inventory entries register at BuildBase time only; ScanServingHandlerArtifacts repopulates zip-lane entries on daemon boot but there is no equivalent for handler-less image-lane bases. Consequence: any brick restart after a serving base was built loses the entry until an external re-key forces a rebuild. Fix: during boot reconciliation, for each serving-class workload whose base exists on disk with a provisioned runtime ref, register a handler-less entry (mirror ReconcileBasesFromDisk; class comes from the SyncRegistry replay).

2. Activator selection. The published empty-cluster fallback is a single activator endpoint, but image-lane entries are per-brick: wakes landing on a brick that never built the workload refuse while the holding brick could serve. Either publish every advertising activator endpoint and accept per-brick refusal retries, or have Envoy/publisher prefer endpoints whose NodeStatus reports a serving_image_ref for the requested workload.

Both are prerequisites for multi-brick image-source serving; single-brick fleets work around (2) by pinning maxInstances placement.

jomcgi:
## Agent brief

### Outcome
Gap 1 of this issue (boot-time backfill of image-lane serving inventory) is already merged via f86ce9eca, which persists a runtime.ref marker at BuildBase time and re-seeds handler-less entries on the startup rescan; gap 2 (activator selection) is still open because the endpoint publisher renders a single activator endpoint that prefers a node whose base snapshot is READY, while the noded activator refuses wakes on bricks with no local serving image, so the remaining actionable work is to make the publisher prefer endpoints whose NodeStatus reports a serving_image_ref for the workload.

### Acceptance
- `node_advertised_activator` prefers advertisers reporting a non-empty serving_image_ref for the workload, keeping the deterministic configured_id ordering and the single-endpoint shape.
- Fallback chain unchanged: any advertiser, then the CP-injected endpoint, then an empty cluster.
- Publisher tests cover an image holder beating a base-READY-only node, and unchanged fallback when no node reports a ref.
- A multi-brick wake for an image-lane workload lands on a holding brick (the #5180 round-3 refusal no longer reproduces).

### Files
[...]

PR #5257, opened 2026-08-24
chore(embervm): re-key ping to image-lane-4 after the overnight brick rolls

jomcgi:
Follow-up to #5225 for the #5180 drill. The image-lane serving-images registration is still BuildBase-time and in-memory (#5221 gap 1): the overnight brick restarts wiped every entry #5225's rebuild created, so wakes 502 with "no serving image provisioned" again this morning (verified 14:47Z, all five attempts, on brick-2gi-node-1).

One more re-key forces the rebuild that repopulates inventory on the current pods so today's end-to-end drill can run. The durable fix (persist a runtime-ref marker per base bundle and re-seed it at boot) lands via #5221.


jomcgi:
## Helm Manifest Diff

PR #5258, opened 2026-08-24
feat(monolith): make public health fail when inference is unavailable

jomcgi:
## What

Registers a fatal `inference` health component on the monolith public tier, so `https://jomcgi.dev/health` reports `status: unhealthy` (503) whenever the in-cluster inference lane is unavailable.

## Why

The public endpoint carried no direct signal for inference. The only qwen-adjacent component was `ember_qwen`, an hourly EmberVM synthetic latch, so an inference outage stayed invisible for up to an hour and then surfaced as a demo probe failure rather than as the dependency it actually is.

## How

The component probes the inference server's own `/health` path, the same one the kubelet readiness probe uses (`projects/inference/deploy/values.yaml`), so it agrees with the Service's definition of "up" rather than inventing a second one.

It lives in `chat_public` because that domain already holds the endpoint config (`CHAT_PUBLIC_INFERENCE_URL`) and already calls inference for public chat. So this needs **no new network privilege, no new env var, and no chart change**. The rule in `core/platform_probe.py` that the public tier must never need privilege to report health is satisfied without the private-writer latch that `cd` uses.

Fails open when `CHAT_PUBLIC_INFERENCE_URL` is unset, matching the bootstrap fail-open in `ember_public/health.py`. A test asserts the module is composed into `PUBLIC_MODULES`, so the check cannot silently become unwired.

## Also

`projects/inference/deploy/values.yaml` claimed the composite `/health` carries no qwen, llama or inference dependency. That was already false when written: `b27104a07` armed the qwen synthetic and registered `ember_qwen` on 2026-08-16. The claim authorised a deliberate outage, so it is corrected rather than left to mislead.

## Known tradeoff

Inference deploys with `strategy: Recreate` and loads a 27B model at start, so an inference rollout will report public health down for the length of that load. That is the honest reading of "inference is unavailable"; revisit if it proves too noisy.

## Verification

`ci` green. `Executed 74 out of 759 tests: 759 tests pass.` The new target runs:

```
$ ci test -- //projects/monolith:chat_public_health_test
Executed 0 out of 1 test: 1 test passes.
```

### DIFF
--- a/projects/inference/deploy/values.yaml
+++ b/projects/inference/deploy/values.yaml
@@ -104,10 +104,10 @@ runtimeClassName: nvidia
 # ENABLING THIS IS A DELIBERATE OUTAGE for every in-cluster qwen consumer:
 # the monolith's Discord, summarize, chat, vision and classifier callsites,
 # monolith-dev, and EmberVM pi-runtime qwen sessions. They fail at DNS, so
-# expect fast errors and dead letters rather than hangs. Nothing pages: the
-# qwen synthetic CronWorkflow is suspended and has never run, and the composite
-# /health carries no qwen, llama or inference dependency (both verified live,
-# 2026-08-18). Embeddings are untouched, so retrieval keeps working.
+# expect fast errors and dead letters rather than hangs. The Service rename
+# also takes public /health unhealthy: the inference component in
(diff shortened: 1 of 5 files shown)


jomcgi:
## Helm Manifest Diff

PR #5260, opened 2026-08-24
feat(embervm): make image-lane serving registrations survive a daemon restart

jomcgi:
Addresses gap 1 of #5221 (boot-time backfill of image-lane inventory entries). Found this morning: after the overnight brick rolls, ping woke 502 again on all five attempts because #5225's registrations lived only in the pre-roll pods' memory.

## The hole

An IMAGE-lane serving base registers its inventory entry at BuildBase time and nowhere else. The zip lane never had this problem: its startup rescan re-seeds every `bases/<key>/handler.zip` bundle from disk. An image-lane base writes no handler.zip, so the rescan had nothing to find and every daemon restart silently un-wakeable'd the workload until something forced another rebuild.

## The fix

Give the image lane the same durability with a runtime-ref-only marker:

- `Driver.WriteServingImageMarker` persists ONLY `bases/<key>/runtime.ref` into the bundle dir.
- `BuildBase`'s image-lane branch writes it with the same contract as its zip-lane twin: write failure fails the build (`FailedPrecondition`) rather than reporting READY on an un-wakeable base.
- `ScanServingHandlerArtifacts` now returns two shapes: zip+ref (artifact entry, unchanged) and ref-alone (HANDLER-LESS entry, `Path ""`, `SizeBytes 0`). A bundle with neither file stays invisible, so task/session bases are unaffected. The existing boot re-seed in `NewServer` picks marker entries up with no change there; `startServingFresh` already boots `Path ""` entries from the image entrypoint.

## Tests

- Driver round-trip: marker write leaves no handler.zip; rescan yields exactly one handler-less entry; a bare task-bundle dir stays skipped.
- Server: image-lane BuildBase records a scannable marker; a restarted `New(Options{ServingDriver: fsd})` re-seeds the wakeable entry keyed by workload from the base-key prefix; marker-write failure fails the build with an empty inventory.
- Epoch bumped to `image-lane-5`: bases built before this binary carry no marker, so one rebuild is what makes ping restart-durable.

`ci` green locally (153 affected tests). Gap 2 of #5221 (single-endpoint activator selection) remains open.

### DIFF
--- a/projects/embervm/chart/values.yaml
+++ b/projects/embervm/chart/values.yaml
@@ -1887,7 +1887,7 @@ pingWorkload:
   nodeLocalWake: true
   # Base-signature re-key for the ping workload (see the template's
   # EMBER_BASE_EPOCH). Inert to the guest; bumping it forces a rebuild.
-  baseEpoch: "image-lane-4"
+  baseEpoch: "image-lane-5"
   guestImage:
     repository: ghcr.io/jomcgi/homelab/projects/embervm/runtimes/ping
     tag: latest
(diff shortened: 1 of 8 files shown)


jomcgi:
## Helm Manifest Diff

Issue #5224, opened 2026-08-23
embervm: minimal functional promotion check for a demonstrated coverage gap


jomcgi:
Approved worth-review scope (2026-09-14): retain the objective of functional verification before Ember promotion, gated on a concrete gap. Replace needs-human with needs-thought.

Gate: establish an actual supported dev-to-prod promotion path and name a functional failure that existing checks cannot catch before release. Record the current promotion flow, relevant checks and missing evidence. Readiness alone does not establish functional correctness, but environment availability alone does not justify the original runner programme.

When the gate is met:

- Revalidate current workloads, interfaces, authentication, storage and promotion behavior. Historical dev topology, retired runtime names and timing assumptions are not implementation requirements.
- Reuse existing checks and facilities for the smallest representative lifecycle test that addresses the named failure. Establish what the test actually proves, including cleanup and bounded execution.
- Bind fresh results to the exact candidate being promoted and the relevant test/configuration identity. Missing, stale, running, incomplete or VACUOUS evidence is insufficient; do not substitute a healthy pod or unrelated prior pass.
- Verify that the intended failure is detected and the valid path passes, including the selected handling of unavailable evidence. Preserve existing authorization, isolation and any explicit operator override's truthful audit trail.
- Scope any eventual integration into promotion separately and validate it in a suitable existing environment.

Removed from mandatory scope: a new conformance-runner Deployment/service and verdict API, recurring chaos phase, fixed minimum invariant count, arbitrary scenario budgets or soak counts, old workload assumptions, and a general scenario framework. Do not revive #4763 or make #4761's gated expansion a prerequisite without its own demonstrated gap.

This decision does not authorize environment creation, pod deletion, new privileges, live fault injection or a promotion configuration change. Existing useful functional checks remain intact.

Historical context: #4745, #4857, #4856 and ADR embervm/034. The earlier prescribed platform and rollout sequence are superseded by this evidence-gated scope.

jomcgi:
G0 scope reconciliation, 2026-09-06, Codex coordinator.

The conformance runner and its chart/Kargo integration are already implemented: `projects/embervm/conformance/`, the three conformance chart templates, the dev `conformance` values, and the optional `verdictURL` HTTP step in `projects/platform/kargo/templates/promotion.yaml`. Current Kargo values explicitly record that the home-cluster dev stage used that functional gate. Reuse this implementation; the issue body is historical implementation scope, not evidence that a new runner still needs building.

[...]

Issue #4949, opened 2026-08-15
embervm: dev exercises only the task workload class, so session and stateful reach prod untested


jomcgi:
## The gap

Live Workload CRs:

| Namespace | Workload | Class |
|---|---|---|
| embervm | sandbox, semgrep, bazel-query | task |
| embervm | sandbox-session, claude-runtime | session |
| embervm | demo-postgres | stateful |
| **embervm-dev** | **sandbox** | **task** |

Dev covers one of three classes. Session and stateful lifecycle (volume attach,
blessing, banking, autoWake, warm restore, snapshot) reach production having run
nowhere first.

## Sizing data, measured

| Workload | Class | Guest RAM | Image (compressed) |
|---|---|---|---|
| sandbox (dev runs this) | task | 512 MiB | 141 MiB |
| **demo-postgres** | **stateful** | **150 MiB** | **119 MiB** |
| sandbox-session | session | 2048 MiB | 141 MiB, SAME image as sandbox |
| semgrep | task | 1536 MiB | n/a |
| bazel-query | task | 1024 MiB | n/a |
| claude-runtime ("pi") | session | 4096 MiB | 287 MiB |

Two things fall out:

**Postgres is already the lightest guest in the fleet.** 119 MiB image, 150 MiB
RAM, smaller in both dimensions than the workload dev runs today. Replacing it
with a SQLite guest would mean a new apko image, guest init, Workload template
and values block, to replace something cheaper than its replacement, and it
would exercise a stateful path production does not run. Explicitly rejected.
The 4 GiB figure in dev's values comments is `runtime-claude`, not postgres.

**Session coverage needs no new image.** `sandbox-session` uses the same guest
image as `sandbox` (identical frozen shim contract: vsock 1027, `/shim/ready`,
`/invoke`; session mode is selected by the request body, not a different
image), so the base dev already builds covers it. `memMib` is values-driven
(`sandboxSessionWorkload.memMib`), so dev can declare a slimmer session.

**claude-runtime / pi is the wrong end**: 4096 MiB, 287 MiB image, plus the
egress lane and credentials dev does not have wired.

## Proposal

Add to dev:
- `demoPostgres` for stateful class coverage
- `sandboxSessionWorkload` with a reduced `memMib` for session class coverage

Nothing new gets built.

## Open decision: brick sizing

Dev's 1gi brick has 512 MiB usable (1024 nameplate minus the 512 daemon
reserve), and `sandbox` alone declares exactly 512.

[...]

PR #5265, opened 2026-08-24
fix(embervm): stop the bank dial fail-open destroying sessions, and self-heal the spec_trace seq

### LINKED ISSUE #5249: embervm: bank dial fail-open to bare node_id destroys a healthy session after CP restart, failing the conformance gate
First live firing of the #5224 conformance gate (2026-08-24 01:51-01:53 UTC, chart 0.42.2) failed dev Promotion `dev.01m0rnf9e0cy8539kd3me00xds.0727c7a`, holding prod at 0.41.7. The gate relayed a real control-plane defect, not a runner bug.

## What the suite saw

- S1 (task cold boot): pass, 1.0s.
- S2: vacuous, detail `bank never observed`, 60.4s.
- S3/S4: vacuous cascade (no baseline, zero invariant coverage).

## What actually happened (dev CP logs, pod embervm-dev-embervm-645df77bbc-4js6p)

The CP pod itself rolled in the same 0.42.2 sync (started ~01:51). The runner created its S2 session 13s later, and the idle-bank then failed three times at the 20s cadence and destroyed the session:

```
01:51:54 embervm session created            workload=pi-runtime principal=...conformance
01:52:14 embervm session bank failed        reason={:error, :unknown_node}
01:52:34 embervm session bank failed        reason={:error, :unknown_node}
01:52:54 embervm session bank failed        reason={:error, :unknown_node}
```

Three strikes hit `@bank_fail_limit`, `handle_bank_failure` called `fail_session_and_destroy`, the runner's poll saw the terminal state and reported S2 vacuous.

## Mechanism

`spawn_bank_worker` (projects/embervm/control/lib/embervm/session_manager.ex:1589) resolves its dial with:

```
dial_key = Embervm.WakeInstance.dial_for_session_vm(state.capacity_table, node_id, vm_id)
```

and the comment says "fail-open to node_id". When the capacity table has no instance reporting this vm (here: freshly restarted CP, brick reports not yet absorbed for a vm created 20s earlier), the fallback is the bare node name. The repo already documents twice that the node-name alias is an anchor, not a dial key, and dialing it fails `:unknown_node` forever (restore_then_prime comment at session_manager.ex:866, retire_session_volume at session_manager.ex:4060). The bank worker has the same latent class: a transient registry miss becomes three guaranteed-failing dials 20s apart, and then destruction of a healthy live session at the 60s mark.

S1 passing in the same minute shows create-path dialing (place_create's own resolved instance dial) worked; only the capacity-table vm lookup missed.

## Defects

1. CP: the bank dial's fail-open to a never-dialable key converts a transient capacity-table miss into session destruction. Either fall back to the placed instance dial (as restore_then_prime does via placed_dial_id), or treat `:unknown_node` as retryable without counting a strike, or both.
2. Runner observability: `runS2` discards the `waitForSessionState` error and reports every failure as `bank never observed` (projects/embervm/conformance/scenarios.go:341). The session was actually destroyed under it; the verdict detail should carry the terminal state or poll error.
3. Process gap: when the runner's next 30m interval run passes on the warm CP, the verdict flips to pass but nothing re-promotes; the failed Promotion is terminal and only new Freight or a manual promotion retries. Worth deciding whether that is acceptable.

## Hypothesis and falsifier

Hypothesis: boot race, the suite ran before the brick's capacity reports repopulated the restarted CP's vm mapping. Falsifier: the runner's second interval run at ~02:21 UTC on the warm CP. If S2 fails again, dev session banking is broken outright and this is bigger than a race.

Refs #5224, #4949. Promotion event: `PromotionFailed: step "conformance": HTTP (200) response met failure criteria` at 01:53:23 UTC.

### LINKED ISSUE #5253: embervm: dev spec_trace writer collides on seq forever, recording nothing and keeping conformance S4 vacuous
The dev control plane (embervm-dev, chart 0.42.4, pod started 2026-08-24 08:24 UTC) logs this on every flush cadence, indefinitely:

```
warning embervm spec_trace store write failed reason="UNIQUE constraint failed: spec_trace.seq"
```

Observed continuously at ~2s intervals at 14:29-14:30 UTC, six hours into the pod's life. Every batch is dropped (`note_drop`), so the trace records NOTHING while `EMBERVM_SPEC_TRACE` reads as enabled. That is precisely the silent state #4841's resume fix was meant to end, and it is one of the two reasons the #5224 conformance gate stays red: S4 reports every invariant `vacuous(coverage=0)` because the trace is empty, and any non-pass scenario fails the suite verdict. Even with the bank defect (#5249) fixed, S4 keeps the gate red until this is fixed.

Why it should not be able to happen, and two candidate mechanisms:

`Embervm.SpecTrace.Writer.init/1` resumes `seq` from `safe_max_seq` (projects/embervm/control/lib/embervm/spec_trace.ex:163 onward), so a restart should climb past the old maximum after at most one collision window. A collision that repeats on every flush for six hours means something is persistently ahead of this writer's counter:

1. Two writers appending to one store: a rolling-update overlap window is too short to explain six hours, but a second Writer instance under another supervisor in the same VM would do it.
2. `max_seq` reading a different backend than `write` lands in: the store has SQLite and Postgres implementations and #5245 just added a parity suite; if init's `safe_max_seq` consults one while flush writes the other, seq resumes from the wrong head forever.

Also worth noting `safe_max_seq` maps ANY store error to 0 by design, so a store that errors only on the bookkeeping query reproduces the eternal-collision state too.

Repro/verification: the dev CP logs; and `curl /verdict` on the conformance Service shows S4 `all invariants vacuous ... coverage=0` on every run since 08:28 UTC.

Refs #5249 (the other gate blocker), #4841 (the original seq-reset defect), #5224, #5245.

jomcgi:
Fixes the two control-plane defects holding the #5224 conformance gate red and prod embervm at chart 0.41.7. Full diagnosis with live log evidence on the issues.

## What was wrong

**#5249, the bank/destroy dial.** `spawn_bank_worker` and `destroy_vm` resolved their gRPC dial via `dial_for_session_vm`, which fails open to the bare node name on a capacity-table miss. The bare node name is never a dial key, so every idle bank in dev got `{:error, :unknown_node}`, three strikes destroyed a healthy session 60s after first idle, and the destroy mis-dialed the same way so the VM leaked, poisoning the node-scoped live-VM count every subsequent conformance run read.

**#5253, the spec_trace seq.** The writer seeds its seq from `safe_max_seq`, which rescues any boot-time store error to 0. With the Postgres store coming up late under `:rest_for_one`, the writer resumed at 0 under a table whose max was far ahead and collided on the primary key on every flush, recording nothing for hours while reading as enabled. That kept conformance S4 at coverage=0.

## The fix

- A manager-state `session_dials` map keeps the placed instance dial from create/restart; bank and destroy prefer capacity-table ownership, then the stored dial, and only then the legacy bare-node fallback. Cleared on every terminal transition; a re-placement that resolves to the bare node clears the stale entry.
- A dial-resolution miss no longer counts as a bank strike (it never reached the node), but is bounded by its own counter (`@bank_dial_miss_limit 30`, roughly ten minutes at the 20s idle cadence) so a session on a genuinely dead instance still terminalizes, with fail reason `bank_dial_miss_limit`.
- The spec_trace writer detects a seq unique-collision (both store shapes), re-queries `max_seq`, and fast-forwards, making the rescue-to-0 boot state self-heal within one flush interval.
- The conformance S2 vacuous verdict now carries the poll error instead of discarding it (the live incident hid "session entered terminal state" behind "bank never observed" for a full diagnosis round).

## Verification

- `ci` green; the embervm CP ExUnit genrule executed remotely against the final branch: `1485 tests, 0 failures, 21 skipped` (judged by the genrule's own line, as required).
- New tests: bank dials the stored instance dial on a table miss; a dial miss does not strike and does not destroy; the dial-miss limit terminalizes; a successful bank resets the counter; three real strikes still destroy; spec_trace fast-forward for both error shapes; a non-collision error does not fast-forward.
- Opus review completed; all five findings addressed in commits 414b66284 and 0509d61b3.

After merge: chart publish rolls dev via Kargo, the conformance gate re-runs on the new version, and S2/S4 tell us whether the dev session lane clears or the next layer surfaces.

Closes #5249. Closes #5253. Refs #5224, #4949.

https://claude.ai/code/session_011iHVK9WyNFFjUkhn8q4s49

### DIFF
--- a/projects/embervm/conformance/scenarios.go
+++ b/projects/embervm/conformance/scenarios.go
@@ -338,7 +338,7 @@ func runS2(ctx context.Context, cfg config, client *controlPlaneClient) s2Result
 		return s2Result{verdict: scenarioVerdict{Verdict: verdictFail, Detail: "idle bank wait interrupted: " + err.Error()}, liveDelay: liveDelay}
 	}
 	if _, err := waitForSessionState(ctx, client, session, bankedSessionStates); err != nil {
-		return s2Result{verdict: scenarioVerdict{Verdict: verdictVacuous, Detail: "bank never observed"}, liveDelay: liveDelay}
+		return s2Result{verdict: scenarioVerdict{Verdict: verdictVacuous, Detail: "bank never observed: " + err.Error()}, liveDelay: liveDelay}
 	}
 
 	invokePath := "/v1/sessions/" + url.PathEscape(session.ID) + "/invoke"
--- a/projects/embervm/control/lib/embervm/session_manager.ex
+++ b/projects/embervm/control/lib/embervm/session_manager.ex
@@ -325,6 +325,10 @@ defmodule Embervm.SessionManager do
       # process, which stops on admission) so three strikes across bank attempts fails
       # the session. Cleared on a successful bank or a successful invoke-driven relight.
       bank_failures: %{},
+      # session_id -> placed instance dial. The capacity table is eventually
+      # consistent, so keep the dial that actually created or restarted the VM for
+      # bank and destroy when its live-vm fact is temporarily absent.
+      session_dials: %{},
       # Snapshot-disk low watermark (bytes): when a node's snapshot_disk_free_bytes
       # is below this, disk-pressure eviction fires. nil = eviction disabled.
       disk_low_watermark_bytes: Keyword.get(opts, :disk_low_watermark_bytes, nil),
@@ -476,7 +480,9 @@ defmodule Embervm.SessionManager do
     # the session process's own FIFO parks the caller. So we reply the pid to the
     # caller path via a spawned forwarder, keeping the manager responsive. Simpler:
     # forward synchronously from a short task so the manager is not the bottleneck.
-    case resolve_route(state, session_id) do
+    {route, state} = resolve_route(state, session_id)
+
+    case route do
       {:live, pid} ->
         # Forward off the manager so a long guest round-trip does not serialize other
         # sessions' routing through this one GenServer.
@@ -503,7 +509,7 @@ defmodule Embervm.SessionManager do
   def handle_call({:destroy, session_id}, _from, state) do
     case SessionStore.get(state.session_store, session_id) do
       {:ok, %{state: session_state}} when session_state in [:expired, :evicted, :destroyed, :failed] ->
-        {:reply, {:ok, :already_terminal}, state}
+        {:reply, {:ok, :already_terminal}, clear_session_tracking(state, session_id)}
 
       {:ok, %{state: :destroying}} ->
         {:reply, {:ok, :destroying}, state}
@@ -917,7 +923,7 @@ defmodule Embervm.SessionManager do
             state
           end
 
-        result =
+        {result, state} =
           case outcome do
             {:ok, vm_id, restored} ->
               register_and_start(state, workload, principal, entry, node_id, vm_id, dial_id, lineage_id, restore_lineage, restored)
@@ -937,7 +943,7 @@ defmodule Embervm.SessionManager do
                 _ = retire_session_volume(state, %{volume_node_id: node_id, workload: workload, lineage_id: restore_lineage})
               end
 
-              {:error, {:denied, reason}}
+              {{:error, {:denied, reason}}, state}
           end
 
         GenServer.reply(from, result)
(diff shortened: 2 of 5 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4404, opened 2026-08-06
embervm: per-VM bounded serial output sink via PUT /serial


jomcgi:
Out of the Firecracker v1.12.1 to v1.16.1 research pass on #4389 (Sol report, 2026-08-06).

Firecracker v1.14.0 added PUT /serial with serial_out_path (upstream #5350) and v1.16.0 added a serial rate limiter plus a uart.rate_limiter_dropped_bytes metric (#5824). Today every FC process inherits the daemon's stdout/stderr (ExecLauncher.Launch), so all guest consoles interleave into the noded log, and during the #4389 investigation the probe VM's console was effectively unfindable there.

Sketch: add the minimal /serial request to fcclient, pre-create a per-VM serial file in the bundle, configure it before Start on cold boot AND before LoadSnapshot on restore (serial config is deliberately not snapshotted, so both paths must issue it), apply a conservative byte rate so a noisy guest cannot flood the node, and retain the tail in failure diagnostics.

Also relevant: v1.16.0 #5764 fixed the restored-UART-loses-TX-interrupt bug, so a silent-console-after-restore class of confusion goes away with the binary upgrade; a per-VM sink makes the remaining cases diagnosable.

Blocked on the v1.16.1 rollout (#4389) landing first.

PR #5268, opened 2026-08-24
fix(embervm): PUT /serial rate limiter is a flat token bucket on n v1.16

jomcgi:
Unblocks the fleet: every snapshot restore has been failing since #4404's serial sink landed, which turned Kargo's conformance gate red (prime_failed -> session create 500 on chart 0.44.0) and blocked all promotions.

## Evidence (dev, 15:49:39Z)

```
[vm-cce3215b0b80f10c:fc_api] The API server received a Put request on "/serial" with body
  {"serial_out_path":".../serial.log","rate_limiter":{"bandwidth":{"size":1114112,"one_time_burst":1048576,"refill_time":1000}}}
[vm-cce3215b0b80f10c:fc_api] SerdeJson(Error("missing field `size`", line: 1, column: 207))
```

Column 207 is exactly where the `bandwidth` object closes: n v1.16.1 parses `rate_limiter` as a FLAT token bucket (`{size, refill_time, one_time_burst}`), not the drive/net `{bandwidth: ...}` wrapper this endpoint inherited from upstream Firecracker convention.

## Fix

- `Serial.RateLimiter` becomes `*TokenBucket` (flat); drop the now-unused `RateLimiter` type.
- Pin the flat wire shape in both the client wire-format test (with an explicit anti-wrapper assertion) and the driver cold-boot ordering test.
- Limiter semantics unchanged: same size/burst/refill values.

`ci` green locally (153 affected tests). After this merges, the next freight should pass conformance and let today's queued ping re-keys (#5257/#5260) finally reach prod.

### DIFF
--- a/projects/embervm/noded/fcvm/driver/serial.go
+++ b/projects/embervm/noded/fcvm/driver/serial.go
@@ -85,13 +85,16 @@ func (d *Driver) prepareSerialOutput(threadID string) (string, error) {
 // burst for boot-time kernel spew, then 64 KiB/s sustained. Once the bucket
 // empties Firecracker drops further UART bytes, so a flooding guest loses
 // middle output but the most recent console lines keep flowing.
-func serialRateLimiter() *fcclient.RateLimiter {
-	return &fcclient.RateLimiter{
-		Bandwidth: &fcclient.TokenBucket{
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

PR #5269, opened 2026-08-24
feat(embervm): idempotent session creation with a client-supplied key

### LINKED ISSUE #4919: embervm: idempotent session creation with client-supplied key
## Status 2026-08-22

Still valid, but the motivating consumer changed. The merge-queue reconciler (#4915) was retired in favour of GitHub's native merge queue (e8d15bad9), so there is no reconciler escalation step to dedupe. The property itself is still wanted: this platform has already produced duplicate sessions from a `uuid4` generated inside a retried DBOS step (see the dbos-graph notes on at-least-once steps), and every monolith-to-Ember spawn from a workflow has the same exposure (swarm runs, voice companion asks, future issue-verifier agents).

Files:
- `projects/embervm/proto/embervm/node/v1/` and the control plane's create path (`projects/embervm/control/lib/embervm/session_manager.ex`, `session_store.ex`): an optional `idempotency_key` on session create, unique per principal, enforced by the store rather than check-then-create
- `projects/monolith/agent_sessions/` and `swarm/`: pass a deterministic key from the workflow (e.g. `swarm/{run_id}/{node_id}`) and treat create as an upsert
- a list/query by key so a crash between spawn and record recovers the existing session

Acceptance: two concurrent creates with one key return the same session id; a create with a reused key after the session is destroyed returns a conflict rather than a silent new session; an ExUnit test plus a monolith test covering the retried-step path.

---

The reconciler's escalation step is at-least-once (DBOS), so session creation must dedupe on a caller-supplied idempotency key, e.g. `merge-queue/{pr}/{headSHA}/rebase`. Add client-token dedupe to EmberVM session creation (server-side unique constraint, not just check-then-create) and a way to list/query sessions by key so the tick's spawn is an upsert and a crash between spawn and record cannot create a duplicate agent. This platform has already produced duplicate sessions from uuid4 generated inside a retried step. Benefits every future monolith-to-Ember flow, not just the merge queue.

jomcgi:
Closes #4919.

Optional `Idempotency-Key` on `POST /v1/workloads/{name}/sessions` makes session create an upsert, so a retried DBOS step (or any at-least-once caller) can no longer mint duplicate sessions:

- same-key retry against a live holder replays the SAME session (200, capability token returned exactly once); concurrent same-key creates park and resolve to the winner; reuse after the holder went terminal conflicts (compaction frees the key); malformed/oversized keys 400 before anything runs
- enforcement lives in the op-log projection, not check-then-create: both backends gain `sessions.idempotency_key` plus a unique partial `(principal, idempotency_key)` index, collision rolls back the whole append; the self-holder no-op preserves adopt-and-backfill's ON CONFLICT DO NOTHING contract
- SessionStore indexes the key in ETS, rebuilds from the projection on boot; docs/api.yaml documents the header

Verified locally with the full pinned-toolchain replication (incl. the fake noded and live Postgres 18): new tests 23/0, affected suites green after one test fix, full suite 1465 tests with 1 pre-existing RouterTest flake (baseline-proven on clean main). `ci` green (149 affected targets); the ExUnit genrule runs in pr-checks/queue rather than the local affected set.

### DIFF
--- a/projects/embervm/control/lib/embervm/op_log/postgres.ex
+++ b/projects/embervm/control/lib/embervm/op_log/postgres.ex
@@ -141,6 +141,7 @@ defmodule Embervm.OpLog.Postgres do
     CREATE TABLE IF NOT EXISTS sessions (
       session_id TEXT PRIMARY KEY,
       lineage_id TEXT,
+      idempotency_key TEXT,
       tenant TEXT NOT NULL,
       principal TEXT,
       workload TEXT,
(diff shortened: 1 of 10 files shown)


jomcgi:
## Helm Manifest Diff

PR #5270, opened 2026-08-24
fix(monolith): repair GitHub calls broken by the org move

jomcgi:
## The bug

The repo moved from `jomcgi/homelab` to the org `jomcgi-org/homelab`. Nine monolith modules still hardcoded the old slug, GitHub answers those paths with a **301**, and httpx does not follow redirects by default. Every one of those calls has been failing silently since the move.

It was visible in production on the `cd` health component:

```
ci check unavailable (Redirect response '301 Moved Permanently' for url
'https://api.github.com/repos/jomcgi/homelab/commits?sha=main&per_page=20'
Redirect location: 'https://api.github.com/repositories/847803371/commits?sha=main&per_page=20')
```

The CI half of `cd` had been blind for over 12 hours. The other eight call sites were failing the same way with no surface at all: `agent_sessions/mcp.py`, `chat/acl.py`, `goosecracker/repo_catalog.py`, `home/dashboard.py`, `home/observability/stats.py`, `semgrep_scan/perf_harvest.py`, `semgrep_scan/router.py`, `swarm/compare_router.py`.

## The fix

One source of truth in `core/github.py` (the leaf package all of these already depend on), read from the environment so the next move is a values change rather than a code change:

```python
GITHUB_API = "https://api.github.com"
GITHUB_REPO = os.environ.get("GITHUB_REPO", "jomcgi-org/homelab")
```

All nine call sites now import it, and every GitHub client passes `follow_redirects=True`. A rename should degrade to a slower call, never to a silent outage.

## The guard test is the point

`github_repo_slug_test.py` walks every non-test `.py` under `projects/monolith` and fails if the stale literal reappears.

Getting that to actually work took two corrections worth recording, because both failure modes are ones this repo has hit before:

1. **It first passed vacuously.** A bazel test only sees its own runfiles, so with `srcs` alone the walk found nothing, the violations list was empty, and the assertion was trivially true. The target now carries a `data = glob(["**/*.py"])`, and the test asserts a **floor of 200 scanned files** so a glob that silently stops matching fails loudly instead of going quiet. (`app/architecture_test.py` does the same kind of walk and has no bazel target at all, so there was no working precedent to copy.)
2. **The glob then broke collection.** pytest auto-collects any `conftest.py` under its rootdir, which in a bazel test is the runfiles tree, and importing those pulls `sqlmodel` and friends this target does not depend on. `**/conftest.py` is excluded, with the reason in a comment.

Verified the guard actually bites rather than passing for a new reason: against the real tree it scans **384** files, finds **0** violations, and detects an injected stale literal.

## Verification

`ci` green. `Executed 1 out of 762 tests: 762 tests pass.`

### DIFF
--- a/projects/monolith/BUILD
+++ b/projects/monolith/BUILD
@@ -2136,6 +2136,40 @@ py_test(
 )
 
 py_test(
+    name = "github_test",
+    srcs = ["core/github_test.py"],
+    imports = ["."],
+    deps = [
+        ":pkg_core",
+        "@pip//pytest",
+    ],
+)
+
+# The sources are `data`, not just `srcs`: this test WALKS the tree rather than
+# importing it, and a bazel test only sees its own runfiles. Without the glob the
+# walk finds nothing, the violations list is empty, and the test passes while
+# guarding nothing. The test asserts a floor on the file count for the same
+# reason, so a future glob that silently stops matching fails loudly.
+py_test(
+    name = "github_repo_slug_test",
+    srcs = ["github_repo_slug_test.py"],
+    data = glob(
+        ["**/*.py"],
+        exclude = [
+            "**/*_test.py",
+            "**/test_*.py",
+            # pytest auto-collects any conftest.py it finds under its rootdir,
+            # which here is the runfiles tree. Collecting one imports it, and
+            # these pull sqlmodel and friends that this target deliberately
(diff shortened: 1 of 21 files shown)


jomcgi:
## Helm Manifest Diff

PR #5271, opened 2026-08-24
fix(embervm): split snapshot_ref into local presence and store fetchability

### LINKED ISSUE #4893: embervm: snapshot_ref conflates "present on this node" with "fetchable from the store", at three layers

jomcgi:
Closes #4893.

`snapshot_ref` was read at three layers as two different claims (present on this node vs fetchable from the store), and the two came apart in exactly the ways the issue pins. BaseBuilder now tracks the distinction explicitly and the three read sites consume the claim they actually need; tests cover the split at each layer including refs local-but-gone-from-store and store-only.

Local full ExUnit on the branch converged to the pre-existing RouterTest flake only; in CI the full suite is green (1480 tests, 0 failures). `ci` green (760 targets).

### DIFF
--- a/projects/embervm/control/lib/embervm/base_builder.ex
+++ b/projects/embervm/control/lib/embervm/base_builder.ex
@@ -85,9 +85,28 @@ defmodule Embervm.BaseBuilder do
       gap), `BaseBuilt=False`.
 
   `status.snapshotRef` is only ever ADVANCED to a base the daemon has already
-  confirmed built, never to an in-progress one, which is what makes the
-  acceptance property hold: at any point, the `snapshotRef` in status names a
-  restorable base.
+  confirmed built, never to an in-progress one. What it NAMES is deliberately
+  narrow (#4893): a ref the daemon confirmed BUILT, i.e. it was present on a
+  node and boot-provable at that moment. It does NOT claim "fetchable from the
+  store". Store fetchability is a separate claim, tracked per vendor in the
+  workload's `store_confirmed` ledger and populated ONLY from node-reported
+  facts (`WorkloadCapacity.exported == true` on a READY base, noded's meta.json
+  completeness marker). No single field answers both questions, and every
+  reader says which it means:
+
+    * `Ready`/`snapshotRef`: a built base exists (a local-presence claim; #4937
+      keeps health out of it, #5009 keeps the claim at BOOT);
+    * `vendor_built`: which vendor's build the CP can vouch for, at signature;
+    * `store_confirmed`: the store provably holds that vendor's current ref;
+    * hydrate (restore-first): requires the ref recorded AND absent locally
+      AND affirmatively fetchable from the store. "Recorded" alone never
+      implies "fetchable".
+
+  When a recorded ref is absent locally and NOT fetchable, the builder does not
+  fire a doomed RestoreArtifact: if some node still holds the base READY it
+  exports that copy first (the reconcile backstop), and if no node holds it the
+  builder converges straight to BuildBase (the same destination the hydrate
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

PR #5272, opened 2026-08-24
fix(embervm): prewarm codex and pi CLIs like claude

### LINKED ISSUE #4423: embervm: prewarm the codex and pi CLIs like claude, so every family pays zero init on the first turn
Follow-up to the first-message latency work. With warm restore armed, VM start is ~2.5ms and the remaining first-turn cost is workspace hydration plus CLI startup. Hydration is being cut ~20x separately (shallow clone). This issue is the CLI half, for the two families that get no prewarm at all today.

## Current state

`shim.py` declares:

```python
SUPPORTED_PREWARM_CLIS = ("claude",)
```

and `prewarm()` only branches on `cli == "claude"`. `EMBER_PREWARM_CLIS` is set to `"claude"` in `projects/embervm/deploy/values.yaml`. So a codex or pi session spawns its CLI lazily on the first turn and pays full startup, while a claude session finds a parked process already running in the base snapshot.

## Why this should be tractable

Both adapters are ALREADY long-lived processes with late-bound identity, which is the hard part and is done:

| adapter | shape | late-bind |
|---|---|---|
| `CodexProcess` (shim.py:1567) | long-lived `codex app-server` | `thread/resume` (shim.py:1848) |
| `PiProcess` (shim.py:2130) | long-lived `pi --mode rpc` | `switch_session` (shim.py:2385) |

Unlike claude, neither needed a patched binary to late-bind identity. So the work is mostly: add them to `SUPPORTED_PREWARM_CLIS`, add prewarm branches that spawn them, and extend `EMBER_PREWARM_CLIS`.

## The real risk, which needs live validation not just tests

A prewarmed process is captured in the BASE SNAPSHOT, so every restored session starts from byte-identical process memory. Two specific hazards:

1. **Captured connections.** If `codex app-server` or `pi --mode rpc` opens an upstream connection (or resolves DNS, or seeds TLS state) at startup, that state is frozen into the base and is dead after restore. Every restored session would inherit the same stale socket. This is the same class as the stale gRPC channel problems seen elsewhere in this system.
2. **Restore-time triggers cannot derive from in-process state.** Restored guest memory is identical to the base, so anything that needs to happen "on restore" must key off EXTERNAL state the restore changed (device superblock, mount table), never an in-process flag. Prior primed-CLI attempts fell into this hole repeatedly.

`PiProcess._spawn` also takes a `model`, so prewarm needs a defensible default and a story for what happens when the first turn asks for a different model within the family.

## Scope

- add `codex` and `pi` to `SUPPORTED_PREWARM_CLIS` and to `prewarm()`
- decide and document the pi prewarm model default
- set `EMBER_PREWARM_CLIS: "claude,codex,pi"` in deploy values (this re-keys the claude-runtime base via `initEnv`, forcing a rebuild, which is the intended mechanism)
- validate live per family: a session of each that restores, takes a turn, and bills, plus one park/relight cycle

Blocked-ish on #4393: the init-wait bug should be fixed first so prewarm failures are not masked by a spurious 30s timeout.

jomcgi:
Closes #4423.

Prewarm was wired to a dead knob: `EMBER_PREWARM_CLIS` sat in workload initEnv, which never reaches a task-class guest (#4429 class), so codex and pi paid full init on their first turn. Config now reaches the guests the only way it can: a `prewarm-clis` file baked into each runtime image via `pkg_tar` + `apko_image(tars)`, with the claude runtime's prewarm path extended to codex and pi per their init profiles.

Note: this changes guest images, so rollout is vendor-keyed and slow (#4937); the report documents what CI must prove (image builds are bazel-only). ExUnit + shim pytest green locally (200/233 passed); `ci` green (760 targets, full ExUnit 1476/0).


jomcgi:
## Helm Manifest Diff

PR #5279, opened 2026-08-24
fix(embervm): a wake denied on memory pressure queues instead of failing the turn

### LINKED ISSUE #4355: embervm: relight denied on pressure:mem surfaces as a failed turn instead of queueing
Observed live (2026-08-05 ~00:54Z): turn 2 on a fresh /agents session returned 502 with reason {:relight_failed, {:prime_failed, GRPC status 8 "noded: pressure:mem (need 4096 MiB, floor 512 MiB)"}}, retryable: false, so the monolith surfaced a dead turn to the user. The session itself parked seconds later and the ledger shows every other claude VM already evicted (idle_ttl) within the preceding minutes, so the denial was either momentary co-residency or reclaim lag between eviction and noded's memory ledger (the recurring CP-ledger-vs-brick-ledger shape).

Structural context: claude-runtime VMs are 4096 MiB and only the single 16gi brick can host them (2gi bricks never can), so real concurrency is roughly (16384 - 512 daemon floor - other workloads) / 4096, about 3. Any overlapping use plus a relight contends.

The failure handling is the actual bug: memory pressure is inherently transient (idle VMs park or evict on TTLs measured in minutes), so a prime denial on pressure should queue, not kill the user's message.

Fix directions, smallest first:
1. Monolith transport: treat prime/relight RESOURCE_EXHAUSTED as retryable with backoff for a bounded window (2 to 3 minutes), same lane as the mid-wake 409/429 retries from PR #4335; the pending-message machinery already redelivers, so this may be mostly a retryable-classification change.
2. CP: classify {:prime_failed, pressure} as retryable: true in the invoke error body so callers know.
3. CP/noded: on pressure at prime, proactively park or evict the idlest live VM for the workload (harvest) instead of denying, bounded to one attempt.

Also worth recording the capacity math in projects/embervm docs: one 16gi brick is the entire claude-session fleet budget today.

Refs #4330 (the console work that surfaced it), memory of the recurring ledger-ahead shape.

jomcgi:
Closes #4355.

A relight denied with RESOURCE_EXHAUSTED (`pressure:mem`) surfaced as `{:relight_failed, ...}` to every parked caller, rendering a dead 502 turn. Memory pressure is transient, so denial is now an admission delay: callers stay parked in the relighting ledger, the session unwinds to its resting state, and a retry timer re-drives the same wake lane.

Bounded: after `pressure_wait_bound_ms` (default 180s, values-wired) the parked callers get a distinguishable `{:pressure_wait_expired, reason}` error, still classified retryable; the session stays alive. `snapshot_lost` keeps priority and is never queued, including mid-episode (regression test).

Verified: 4 new tests stable across 3 runs; targeted suites 130/0; full-suite flakes baseline-equal on pristine main; `ci` green with full ExUnit 1504 tests 0 failures.

### DIFF
--- a/projects/embervm/control/lib/embervm/application.ex
+++ b/projects/embervm/control/lib/embervm/application.ex
@@ -986,12 +986,33 @@ defmodule Embervm.Application do
       node_confirmed_destroy: node_confirmed_destroy_enabled(),
       destroying_alarm_ms: destroying_alarm_ms(),
       orphan_grace_ms: orphan_grace_ms(),
+      # #4355 pressure-parked wake retry cadence and bound.
+      pressure_retry_interval_ms: session_pressure_retry_interval_ms(),
+      pressure_wait_bound_ms: session_pressure_wait_bound_ms(),
       # The Direction-2 adopt-and-backfill discriminator asks AsyncWriter whether a
       # rowless reported VM has a write in flight (ADR embervm/014 decision 2), so it
       # needs both the gate and the writer reference.
       async_lifecycle_writes: async_lifecycle_writes_enabled(),
       async_writer: Embervm.AsyncWriter
     ] ++ wake_opts()
+  end
+
+  # Retry cadence for a wake parked behind node memory pressure (#4355;
+  # EMBERVM_SESSION_PRESSURE_RETRY_INTERVAL_MS), default 5s.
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

Issue #4417, opened 2026-08-06
embervm: lane connection #2 wedges (multi-connection vsock), constrains in-session git fetches


jomcgi:
## Approved scope: bounded two-connection diagnosis and correction

Keep agent-ready. The historical report under #4389 reproduced a second guest-initiated lane connection completing connect but delivering no first bytes host-side, across cold/restored guests and two Firecracker versions. In-session git fetch/pull while another tunnel lingers is a potential factory impact. These are recorded findings, not proof of the current transport state.

## Reproduction

Use the current transport and a minimal two-connection scenario. Control whether connection A fully closes or completes and lingers while connection B starts. Observe enough of the existing guest/host path to locate the stall rather than assuming the Firecracker muxer or guest driver is responsible.

Cover concurrent and sequential connections, with bounded deadlines and cleanup so the diagnostic cannot itself leave indefinite tunnels. Check cold/restored behavior where relevant to the reproduced fault.

## Correction

Fix only the confirmed fault through its existing transport/lifecycle owner. Preserve working sequential traffic, established tunnel semantics and current hydration behavior. No transport rewrite, new proxy service or unrelated multiplexing framework.

If the issue is already fixed, record evidence from the current reproduction and existing correction instead of adding redundant code. If an upstream limitation remains, record the exact evidence and smallest viable mitigation rather than broadening this task.

## Acceptance

The reproduction detects the failure when present and passes with the correction. Both connections make progress under their supported lifetime semantics; completed/failed connections clean up within the documented bounds. Verify representative affected git/HTTPS behavior through available supported validation, without treating a connect return alone as successful delivery.

Required Linux validation covers changed paths. Do not run prohibited local full suites or claim production verification from a test-only result.

This scope supersedes the worth report's closure recommendation. #4389 and PRs #4414/#4415 remain historical evidence to revalidate.

jomcgi:
## Triage recommendation

**Recommendation:** Keep (needs-thought)

[...]

PR #5280, opened 2026-08-24
feat(embervm): node-local git mirror over smart http for hydration

### LINKED ISSUE #4473: embervm: serve the git mirror over http so hydration is node-local again
Hydration currently clones from GitHub over https (#4470). That was the right call when a clone was occasional, but the MCP agent tool now defaults every voice/MCP session to `jomcgi/homelab`, so the clone is on essentially every session.

## Why it costs more than it looks

Volumes are per lineage, so the clone is per **session**, not per repo. There is no shared object cache between sessions, and the VM listing shows heavy session churn with frequent `idle_ttl` evictions. So this is ~23 MB from GitHub per interaction, with rate limits that scale the same way.

Measured on `jomcgi/homelab` against real GitHub:

| shape | wall | `.git` | commits |
| --- | --- | --- | --- |
| `--depth=1` (old) | 2.46s | 12 MB | 1 |
| `--filter=blob:none` (current) | 4.12s | 23 MB | 11,553 |

Against the mirror the same content was ~1.3s server-side over loopback.

## Why the mirror was dropped, and why that reason is fixable

Not because of the mirror. Because of the transport. #4389 stopped propagating half-close onto the vsock leg, and that suppression is scoped to one port:

```
GIT_DAEMON_PORT = "9418"
half_close_upstream = not host_port.endswith(":" + GIT_DAEMON_PORT)
```

So a `git://` tunnel never closes, and a second lane connection opened while it lingers wedges (#4417). That is what forced `--depth=1`, since a blob filter lazy-fetches on exactly that second connection.

Over any other port the tunnel tears down normally, which is why sequential CLI HTTPS and `gh` both work.

## Proposal

Serve the mirror over http (`git-http-backend`) on a non-9418 port and point hydration at it:

- node-local again (~1.3s), no GitHub rate limit on session create
- **no credential at all** on the clone path, unlike today
- keeps full history plus lazy blobs: the mirror already sets `uploadpack.allowFilter` ("Enable partial/shallow clones so guests can use `--filter=blob:none`")
- invisible to the MCP tool and the console: same `repo` argument, different clone URL

## Before building

Run the two-connection diagnostic #4417 has been waiting for: two connections on a **non-9418** port, with connection A lingering vs closed. Sequential HTTPS working is strong evidence the lane is fine there, but #4417 notes the wedge appeared "even in earlier configurations" and the splitting diagnostic was never run. That experiment de-risks this and unblocks #4417 either way.

## Caveat

The mirror lags GitHub by up to one refresh interval (60s), and it clones private repos only if its own token has access. Today's GitHub-direct path has neither limitation, so this is a real tradeoff, not a pure win.

Refs #4417, #4470, ADR agents/050, ADR agents/041

jomcgi:
Closes #4473.

Serves the git mirror over smart http and points session hydration at it, taking the clone off GitHub's rate limit and the ~9x-slower credentialed egress lane:

- new `embervm-git-mirror` sidecar (Go server + refresh loop): anonymous read-only upload-pack from bare clones, filter capability always advertised so guests keep `--filter=blob:none`
- egress-proxy gains a reserved-host seam (`git-mirror.egress.internal` -> co-located loopback), fail-closed when unset, mirroring the caFetchHost pattern; scoped away from the port-9418 half-close suppression (#4417)
- shim hydration tries the mirror credential-free first, falls back to the credentialed github.com clone for unmirrored/private repos
- chart: `gitMirror` values block gated on egress.enabled, image pinned via the images map; prod values arm it for jomcgi/homelab
- amd64 apko image (wolfi git, uid 65532), lock bootstrapped two-pass under podman

Verified locally: go vet/test green incl. a real `git clone --filter=blob:none` smoke against the served fixture; shim pytest 25/25 + full runtime suite (2 macOS-only AF_VSOCK failures baseline-proven pre-existing); helm renders on/off/prod all checked; lock checksum verified. CI must prove: rules_apko lock resolution + image build/push and digest pin (called out in the worktree report). `ci` green (763 targets, full ExUnit 1500/0).

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -590,7 +590,13 @@ apko.translate_lock(
     name = "embervm_runtime_pi_lock",
     lock = "//projects/embervm/runtimes/pi:apko.lock.json",
 )
-use_repo(apko, "embervm_conformance_lock", "embervm_lock", "embervm_noded_lock", "embervm_runtime_bazel_lock", "embervm_runtime_claude_lock", "embervm_runtime_k3s_agent_lock", "embervm_runtime_k3s_server_lock", "embervm_runtime_pi_lock", "embervm_runtime_postgres_lock", "embervm_runtime_python_lock", "embervm_scratch_prep_lock", "embervm_tokenbroker_lock", "embervm_xds_lock", "monolith_frontend_lock", "rootfs_builder_lock", "sandbox_elixir_lock", "sandbox_go_lock", "sandbox_javascript_lock", "sandbox_ocaml_lock", "sandbox_python_lock", "sandbox_rust_lock", "semgrep_guest_lock", "shotter_lock", "whatsapp_lock")
+
+apko.translate_lock(
+    name = "embervm_mirror_lock",
+    lock = "//projects/embervm/mirror/image:apko.lock.json",
+)
+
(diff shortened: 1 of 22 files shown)


jomcgi:
## Helm Manifest Diff

PR #5297, opened 2026-08-25
fix(monolith): repair the hourly homelab changelog broken by the org move

jomcgi:
chat-changelog-homelab has failed every hour for hours with `main: Error (exit code 1)`. The cause is a stale repository slug in config, not in code: `projects/monolith/deploy/values.yaml` still named `jomcgi/homelab`, the pre-move owner.

## Chain

1. `githubRepo: "jomcgi/homelab"` is the pre-org-move slug.
2. GitHub answers it with `301` to `jomcgi-org/homelab`.
3. `httpx.AsyncClient` defaults to `follow_redirects=False`, and `raise_for_status()` returns early only on `is_success`, so the 301 raises `HTTPStatusError`.
4. That propagates out of the job, exit 1.

`chat-changelog-colin-homelab` (`ColinCee/homelab`) and `chat-changelog-loom` (`weave-hand/loom`) point at repos that never moved, which is why only this one entry failed.

Verified against the live API: the old slug returns `301` with `Location: .../repos/jomcgi-org/homelab/...`, the new slug and both sibling repos return `200`.

## Changes

- Correct the slug to `jomcgi-org/homelab`.
- Add `follow_redirects=True` to both `AsyncClient` constructions in `chat/changelog.py`, matching the hardening #5270 applied to the GitHub call sites in code.

Either change alone fixes the symptom. They are kept together because they fail differently: the slug fix is exact but goes stale again on the next move, and the redirect fix degrades any future move into a redirect rather than an outage. The two land safely in either deploy order, since the corrected slug returns `200` under the old code and the old slug is followed under the new code.

## Why #5270 did not catch it

That PR swept GitHub call sites in code. This slug is a *value*, delivered through the `$values` git ref rather than baked into the image, so a call-site sweep could not reach it.

## Tests

The existing suite injects an `AsyncMock(spec=httpx.AsyncClient)`, so no test in it could observe redirect behaviour at all: the seam that broke is the client construction, which nothing reached. That is why this was green in CI while failing hourly in production.

The new tests drive a real `AsyncClient` over `httpx.MockTransport`, one asserting the org-move redirect is followed and one pinning the 301 failure mode. Confirmed the second one fails against the unfixed code path before being corrected to assert the real exception.

`ci test`: `Executed 1 out of 360 tests: 360 tests pass.`

### DIFF
--- a/projects/monolith/chat/changelog.py
+++ b/projects/monolith/chat/changelog.py
@@ -174,7 +174,9 @@ async def run_changelog_iteration(
 
     since = datetime.now(timezone.utc) - timedelta(hours=config.interval_hours)
 
-    async with httpx.AsyncClient(timeout=httpx.Timeout(600.0)) as client:
+    async with httpx.AsyncClient(
+        timeout=httpx.Timeout(600.0), follow_redirects=True
+    ) as client:
         commits = await _fetch_commits_since(
             client, config.github_repo, github_token, since
         )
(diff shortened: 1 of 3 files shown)


jomcgi:
## Helm Manifest Diff

Issue #5308, opened 2026-08-25
observability: un-pin ClickHouse from 25.5.6 by doing the 25.12 migration properly


jomcgi:
`projects/platform/signoz/values-prod.yaml` pins `signoz.clickhouse.image.tag: 25.5.6` as of #5307, to end the outage in #5298. That pin is a stopgap. Removing it without doing the migration reproduces the outage exactly.

## Why the pin exists

Chart 0.135.1 moved the default from 25.5.6 to 25.12.5. The chart README calls it "a significant ClickHouse version jump" and says to read the upgrade guide first. 25.12.5 cannot load the parts 25.5.6 wrote, and aborts startup on `signoz_logs.logs_v2` with 133 broken parts against a ceiling of 100.

## The work

1. Read the ClickHouse 25.5 to 25.12 upgrade guide and establish what actually breaks. `logs_v2` uses a `JSON(max_dynamic_paths = 100)` column, which is one of the areas that moved between these versions and a plausible specific cause worth confirming rather than assuming.
2. Decide the migration: an intermediate version step, an explicit part conversion, or accepting the loss of the affected parts as a deliberate choice rather than as the side effect of a chart bump.
3. Take a backup first. Whatever the route, this is the telemetry store.
4. Unpin, upgrade, verify ClickHouse serves and SigNoz reports healthy.
5. Confirm what the retention policy means here. If the affected parts age out under the SigNoz retention window anyway, waiting is itself a valid migration and much cheaper than converting them.

## The wider problem this exposed

Two failures made a chart bump into a 30-hour outage, and both outlive this ticket.

**A 16-day gap between config and effect.** The chart bumped 2026-08-08. The pod adopted the image 2026-08-24. A change that lands green and does nothing for two weeks, then breaks production on an unrelated restart, cannot be associated with its cause by anyone reading the commit. Whatever restarted the pod got blamed. Worth deciding whether image changes to stateful workloads should roll deliberately at merge, so the blast radius lands next to the change that caused it.

**Nothing alerted.** The thing that would alert on SigNoz being down is SigNoz. This surfaced from a WARNING line in an unrelated cron job log, more than a day in. The alerting stack needs a liveness check that does not depend on the alerting stack. That is worth its own issue, and this one should not absorb it.

## Do not close this by deleting the pin

The pin is load-bearing until the migration is done. Closing this by unpinning and watching it break is the failure mode this issue exists to prevent.

jomcgi:
[...]

PR #5309, opened 2026-08-25
fix(signoz): revert the ClickHouse 25.5.6 pin, it fails worse than 25.12.5

jomcgi:
Reverts the pin from #5307. It deployed, and the evidence falsified the theory behind it.

## What happened

25.5.6 does not start either. It fails on a **different** table:

```
Code: 231. Suspiciously many (126 parts, 964.09 MiB in total) broken parts
... Cannot attach table `signoz_metrics`.`samples_v4_agg_5m`
```

with each part failing as:

```
Code: 246. Unknown version of serialization infos (1). Should be less or equal than 0.
(CORRUPTED_DATA)
```

That is the reverse incompatibility. 25.12.5 **wrote** those parts in a newer serialization format that 25.5.6 cannot read.

## What I got wrong in #5307

I claimed 25.12.5 never served, and that the pod adopted the new image 16 days after the chart bump. Both were wrong, and they came from the same mistake: I read `Pulled 25.12.5 first-seen 2026-08-24T00:55:42` as the first ever pull. That field is truncated by event retention, not a first-occurrence guarantee.

The parts 25.5.6 now rejects carry partition dates of **2026-08-18 and 2026-08-20**, days before the 08-24 crash. So 25.12.5 was running and writing normally, almost certainly from the chart bump on 08-08, and something separate broke `logs_v2` on 08-24.

## Why revert rather than stay pinned

The two failures are not equivalent:

| version | blocks on | size |
|---|---|---|
| 25.12.5 | 133 parts in `signoz_logs.logs_v2` | **0.00 B** (empty) |
| 25.5.6 | 126 parts in `signoz_metrics.samples_v4_agg_5m` | **964 MiB** (real data) |

Staying on 25.5.6 leaves a gigabyte of good metrics data one raised ceiling away from being discarded by whoever picks this up next. Reverting puts the blocker back on parts that are worth nothing.

I left a `DO NOT PIN BACK TO 25.5.6` comment where the pin was, with the reason, so the downgrade is not tried a third time.

## This also inverts the recommendation I gave

In #5307 I argued against raising `max_suspicious_broken_parts` because it would "discard 133 parts of real log data". That was wrong: those parts total **0.00 B**. They are empty. Raising the ceiling on 25.12.5 discards nothing of value and is now the most likely correct fix, which is close to the opposite of what I advised.

I am not doing that in this PR. It is the destructive-shaped option and it should be a deliberate call with the corrected numbers in hand, not something folded into a revert.

## State

SigNoz is still down. This restores the better of two broken states; it does not fix it. #5298 carries the diagnosis and #5308 the migration work, both updated.

### DIFF
--- a/projects/platform/signoz/values-prod.yaml
+++ b/projects/platform/signoz/values-prod.yaml
@@ -232,23 +232,12 @@ signoz:
       # Health probes are exempt and stay on /api/v1/health at the root.
       signoz_global_external__url: https://private.jomcgi.dev/app/signoz
   clickhouse:
-    # PINNED to the version that last started cleanly. Chart 0.135.1 moved this
-    # default 25.5.6 -> 25.12.5; the chart README calls that "a significant
-    # ClickHouse version jump" and says to read the upgrade guide first. It was
-    # not read. The pod adopted the new image 16 days after the chart bump, on
-    # 2026-08-24T00:55, and has crashlooped ever since: 25.12.5 cannot load the
-    # parts 25.5.6 wrote, and aborts startup on signoz_logs.logs_v2 with
-    # "Suspiciously many (133 parts, 0.00 B in total) broken parts".
-    #
-    # The alternative fix, raising max_suspicious_broken_parts past 133, starts
-    # the server by DISCARDING those 133 parts of real log data. This pin is
-    # reversible and destroys nothing, so it comes first.
-    #
-    # Removing this pin without performing the documented 25.5 -> 25.12 data
-    # migration reproduces the outage. See #5298 for the outage and #5308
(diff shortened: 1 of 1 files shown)


jomcgi:
## Helm Manifest Diff

<details>

```diff

spec.templates.podTemplates.pod-template.spec.containers.clickhouse.image  (clickhouse.altinity.com/v1/ClickHouseInstallation/signoz/signoz-clickhouse)
  ± value change
    - docker.io/clickhouse/clickhouse-server:25.5.6
    + docker.io/clickhouse/clickhouse-server:25.12.5
```

</details>

PR #5310, opened 2026-08-25
fix(embervm): refuse to adopt bases the local firecracker cannot load

### LINKED ISSUE #4407: embervm: the hypervisor version belongs in base identity, not a manual epoch ride-along
Out of the PR #4406 review. Base rebuilds key on the spec signature (image ref, zip sig, vcpus, memMib, guest port, ready path, init env: base_builder.ex signature computation). The hypervisor version is not an input, so a Firecracker binary change invalidates every snapshot (formats are version-locked) while the BaseBuilder sees identical signatures and takes the idempotent no-op branch, or worse re-hydrates the same stale bytes from the store. The noded boot reconcile registers bases READY on file presence alone, with no notion of which Firecracker wrote the snapfile, and Claim has no cold-boot fallback once a base ref is recorded: a failed LoadSnapshot aborts the dispatch.

PR #4406 papers over this with a shared hypervisorEpoch chart value rendered into every workload's initEnv, which works but is a manual contract ("bump the epoch in the same PR as the binary"): the failure mode when someone forgets is every dispatch failing fleet-wide, discovered post-merge.

Durable fix: make the hypervisor generation a first-class input to base identity. Options, roughly in order of preference:

1. noded stamps the running Firecracker version (it ships the binary, so it knows) into the base bundle metadata at build time and into its capacity/base facts; the CP folds that version into the signature and into restore eligibility, so a version-mismatched base is simply not eligible and triggers the existing rebuild path (the same mechanics that handle a missing base today, proven by the #4400 forced-rebuild flow).
2. Validate at register time: the boot reconcile checks the snapfile's recorded snapshot data format version against what the local binary supports (firecracker --describe-snapshot exists for exactly this) and refuses to register incompatible bases, which converts the failure from per-dispatch aborts into the normal absent-base rebuild.

Option 2 is smaller and catches any mechanism that changes the format, including a future kata bundle bump; option 1 additionally covers same-format-but-behavior-relevant version moves. Doing 2 first is probably right.

Related: #4407-adjacent relight classification (banked/stateful snapshots), tracked separately.

jomcgi:
## What

Closes #4407 (option 2, the register-time gate the issue recommends doing first).

`ReconcileBasesFromDisk` adopted every base dir with a snapfile as READY on file presence alone. A base written by a different Firecracker generation is unloadable there (formats are version-locked), so every dispatch resting on it aborted with a hard `PUT /snapshot/load` failure. The only protection was the manual `hypervisorEpoch` ride-along, whose failure mode is a forgotten bump and a fleet-wide post-merge outage.

The boot reconcile now asks ground truth instead of humans:

- the shipped binary's supported format via `firecracker --snapshot-version` (probed once, cached; verified against v1.16.1: prints `v10.0.0`),
- each snapfile's recorded format via `firecracker --describe-snapshot <path>`.

A base is refused when its format cannot load here, mirroring upstream's own gate (`Snapshot::load_without_crc_check`: same major, file minor not newer) or when the binary cannot describe the file at all (restore runs the identical parse plus stricter checks). Refusal reports `BASE_BUILD_STATE_NONE` with the reason in `buildErr`, exactly like the missing-rootfs branch, so it lands in the control plane's normal absent-base rebuild machinery (`node_reports_base_absent?/3`) instead of per-dispatch aborts.

## Deliberate choices

- **Fail open** when our own binary's support is unknown (probe failure, BinPath unset): absent information never invalidates working state, so all pre-existing tests and dev flows are untouched.
- **Refused dirs stay on disk**: a downgrade would make them usable again; reclamation is retention's job.
- No proto or CP changes needed: `NONE` is already the one state the CP acts on. The kv-logger whitelist is untouched.
- Covers the hydrate path too: `reregisterRestored` routes through the same reconcile, so a store copy exported by a different Firecracker generation is refused on adoption and rebuilt locally.

## Validation

- `ci lint`, regen clean; remote affected Bazel tests green (BuildBuddy invocation cc6af9d4-dfbe-4465-be6f-2fd4f83fdef5), incl. `//projects/embervm/noded/server:server_test`.
- New hermetic unit tests cover: parse robustness against Firecracker's stdout log noise, the compatibility matrix, compatible adoption, incompatible/undescribable refusal (NONE + buildErr + dir survives + capacity advertises affirmative absence), fail-open when the probe fails, and probe caching.
- The FC CLI contract was verified empirically against the pinned v1.16.1 release binaries in a Linux VM (stdout shape, exit codes, load-compat rule read from the v1.16.1 source).


jomcgi:
## Helm Manifest Diff

<details>

```diff

spec.templates.podTemplates.pod-template.spec.containers.clickhouse.image  (clickhouse.altinity.com/v1/ClickHouseInstallation/signoz/signoz-clickhouse)
  ± value change
    - docker.io/clickhouse/clickhouse-server:25.12.5
    + docker.io/clickhouse/clickhouse-server:25.5.6
```

</details>

Issue #4659, opened 2026-08-10
observability: alert on policy-denied ingress drops to monolith, the silent failure mode of the new allowlist


jomcgi:
Follow-up to #4628, from a post-merge review of the now-armed policy.

## Why

`monolith-api-ingress` is enforcing (armed 2026-08-10T19:57:57Z, PR #4654). Its failure mode is a **silent dial timeout**, not a readable deny: a caller that is not in the allowlist gets no error it can report, it just hangs and gives up. That is the whole risk profile of an `ingress:` CNP, and today nothing watches for it.

Worse, the obvious manual check does not work retroactively. The hubble ring buffer on these nodes is only a few minutes deep, so "no drops since arming" can only ever be verified for the last few minutes. Any question asked an hour later is unanswerable.

## What to build

Hubble/cilium drop metrics are already scraped into SigNoz (`projects/platform/signoz/values-prod.yaml:102`). Add an alert on policy-denied INGRESS drops whose destination is the monolith app pod.

Note `cilium_drop_count_total{reason="Policy denied", direction="INGRESS"}` is node-wide and cumulative, so it is not usable as-is: it is already nonzero for unrelated reasons. The alert needs to be endpoint-scoped or namespace-scoped to be meaningful, and should fire on a rate increase rather than an absolute value.

## Why it is worth doing properly

This generalises beyond the monolith. Every `ingress:` CiliumNetworkPolicy in the repo has the same property: it fails closed and silently. `monolith-public` already has two ingress policies behind gates. An alert on policy-denied drops is the only thing that turns "we think the allowlist is complete" into something that would actually tell us if it is not.

Related: #4294, where a Cilium egress rule targeting the wrong port failed as a silent dial timeout rather than a readable deny, and cost real debugging time for exactly this reason.

PR #5313, opened 2026-08-25
fix(monolith): narrow the remaining namespace-wide grants in the ingress allowlist

### LINKED ISSUE #4660: security: narrow the remaining namespace-wide grants in the monolith ingress allowlist
Follow-up to #4628, from a post-merge review. The `monolith-workflows` grant has already been narrowed to this chart's own job pods (evicting renovate). Two namespace-wide grants remain, and neither is being fixed in that change because the risk profiles differ.

## envoy-gateway-system to :3000 and :8000

The namespace holds the envoy proxy **and four cloudflared pods**. cloudflared's only legitimate path is to envoy, so the direct grant to `:3000` and `:8000` buys nothing.

It also costs something specific: the gateway-wide `ClientTrafficPolicy` that strips inbound `X-Auth-Email` runs at the **envoy listener**. A direct cloudflared to monolith path bypasses the exact control that makes the header trustworthy, as well as bypassing HTTPRoute routing.

The discriminator exists and is clean:

| pod | `app.kubernetes.io/name` |
|---|---|
| envoy proxy | `envoy` (plus `component: proxy`) |
| cloudflared | `cloudflare-tunnel` |

**Why this was deliberately NOT done alongside the workflows narrowing.** The downside is asymmetric. If envoy-gateway ever relabels its proxy pods during an upgrade, the selector stops matching and **all ingress dies as silent dial timeouts**, a total outage. The workflows narrowing had no such exposure: that label is stamped by the same chart that renders the selector, so the two cannot drift. cloudflared is also first-party, memory-safe infrastructure, so the threat it represents is far smaller than renovate's.

So this is worth doing, but as its own change, consciously, with the relabel failure mode documented next to the rule and a stated rollback. Do it when there is appetite to watch a deploy, not as a ride-along.

Prerequisite worth having first: #4659, an alert on policy-denied drops. With that in place this becomes much safer to attempt, because the failure would be visible rather than silent.

## mcp to :8000

Grants the whole namespace, which includes Context Forge's redis and postgres, when only the gateway pod federates to the monolith API. Low value, low risk, tidy it up whenever the file is next open.

## The thing that actually matters more than either

Neither of these makes the header trustworthy. Even fully narrowed, the allowlist leaves several workloads able to send a forged `X-Auth-Email`. The policy is defence in depth and must never become load-bearing.

The test is simple: if this policy were sufficient, you could authorize on the header. You cannot. So the preview authorization work must verify the CF Access or authentik JWT **in the handler** rather than trusting network position. The CNP's `spec.description` now says so on the live object, so an operator reading it does not over-trust it.

jomcgi:
Closes #4660.

- **envoy-gateway-system** now matches only the envoy proxy pods (`app.kubernetes.io/name: envoy`, `component: proxy`, stamped by Envoy Gateway itself; verified against the pinned v1.8.3 chart's `EnvoyAppLabel()` source). The four cloudflared pods in that namespace dial only the envoy service on :80, so the direct grant bought nothing and would have bypassed HTTPRoute routing and the `ClientTrafficPolicy` that strips inbound `X-Auth-Email` at the listener.
- **mcp** now matches only Context Forge's gateway Deployment; its redis, postgres, and team-mapping CronJob get nothing (the CronJob dials its own namespace's gateway, not the monolith).
- embervm's two namespace-wide entries stay by design (ADR embervm/051, /035 §4) and the new test pins that they are the *only* ones.

Caller enumeration was done before narrowing, from repo manifests plus the hubble flow evidence recorded at arm time (65-minute window, 151,264 flows, source set exactly: envoy-gateway-system, mcp, monolith-workflows, host). Full inventory of both namespaces is in the worktree report.

**Risk, stated plainly:** if an Envoy Gateway upgrade ever relabels proxy pods, all ingress dies as silent dial timeouts. The tripwire is the #4659 policy-drop alert (merged), and the rollback is reverting two label keys. Documented next to the rule.

New `cilium_ingress_allowlist_test` renders prod and dev and pins the allowlist shape. Note: `ci` caught the first cut of its BUILD target missing `Chart.lock`/`charts/`/`migrations/`, so `helm template` aborted on subchart deps; fixed in a follow-up commit. `ci` now green (774 pass).

### DIFF
--- a/projects/monolith/chart/BUILD
+++ b/projects/monolith/chart/BUILD
@@ -90,6 +90,39 @@ py_test(
     deps = ["@pip//pytest"],
 )
 
+# Hand-written guard for the ingress allowlist's SHAPE (#4660): no
+# namespace-wide grants outside the two documented embervm entries, envoy and
+# mcp narrowed to pod identity, cloudflared never granted. Renders the chart,
+# so it needs helm plus both environments' values and Applications; the env
+# fallbacks in the file also let it run under plain pytest outside Bazel.
+py_test(
+    name = "cilium_ingress_allowlist_test",
+    srcs = ["cilium_ingress_allowlist_test.py"],
+    data = [
+        "Chart.yaml",
+        "values.yaml",
+        "//projects/monolith/deploy:application",
+        "//projects/monolith/deploy:deploy_values",
+        "//projects/monolith/dev/deploy:application",
+        "//projects/monolith/dev/deploy:deploy_values",
+        "@multitool//tools/helm",
+    ] + glob([
+        "templates/**/*",
+    ]),
+    env = {
+        "HELM_BIN": "$(rootpath @multitool//tools/helm)",
+        "DEPLOY_VALUES": "$(rootpath //projects/monolith/deploy:deploy_values)",
+        "DEV_VALUES": "$(rootpath //projects/monolith/dev/deploy:deploy_values)",
+        # The Applications, so releaseName is read from where it is DECLARED.
+        "PROD_APPLICATION": "$(rootpath //projects/monolith/deploy:application)",
+        "DEV_APPLICATION": "$(rootpath //projects/monolith/dev/deploy:application)",
+    },
+    deps = [
+        "@pip//pytest",
+        "@pip//pyyaml",
+    ],
+)
+
 semgrep_test(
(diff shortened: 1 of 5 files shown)


jomcgi:
## Helm Manifest Diff

Issue #5142, opened 2026-08-22
security: scope monolith-public egress to its four dependencies instead of toEntities: cluster


jomcgi:
## Status 2026-08-22

Expanded from the one-line pointer so it can be worked from the body.

`projects/monolith/THREAT-MODEL.md` finding 3: the public tier's off-cluster egress lockdown scopes by "in-cluster or not", never by destination. `projects/monolith-public/chart/templates/cilium-policy.yaml` grants the backend `toEntities: [cluster, host, remote-node]` (line 156), and the frontend (198) and imgproxy (222) the same, while the backend rule's own comment names the intended reach as Postgres read/write, inference, embeddings, and SeaweedFS S3. A compromised public-tier pod therefore has pod-to-pod reach across the whole cluster, which is the opposite of what ADR security/004 documents this control as providing.

Fix: replace the backend's `toEntities: cluster` with `toEndpoints` selectors for the four dependencies (CNPG pooler for the `public_reader` role, the inference service, the embeddings service, SeaweedFS S3) plus kube-dns on 53; keep `host` / `remote-node` only if a health probe needs them. Frontend and imgproxy need only the backend (and imgproxy SeaweedFS). The apiserver is not on the list because the public tier has no RBAC. Roll out in Cilium audit mode first and read `hubble observe --verdict AUDIT` for a day of public traffic before enforcing, since a missed dependency is a silent dial timeout. Acceptance: `helm template` of monolith-public shows no `toEntities: cluster` on any public-tier pod, and the public chat, imgproxy, and `/health` paths stay green after enforcement.

---

see THREAT-MODEL.md finding 3 (#5140)

jomcgi:
Closing as done. Destination-scoped egress policies landed in projects/monolith-public/chart/templates/cilium-policy-scoped-egress.yaml (fc1c213fb audit, afbfb020d enforce) and projects/monolith-public/deploy/values.yaml sets scoped: enforce, with a guard test asserting no cluster/world/all entity in the rendered chart.

PR #5314, opened 2026-08-25
fix(monolith-public): enforce destination-scoped egress

jomcgi:
Refs #5142. Completes step 2 of the ladder started in fc1c213fb: with `ciliumPolicy.egress.scoped: enforce`, the broad `toEntities: cluster` grants stop rendering, the scoped policies carry default-deny, and the Turnstile siteverify CIDR allow moves into the web policy.

**Re-enumerating every dial against code found two problems with the audited target list:**

- **A fifth and sixth destination the issue's four missed.** The demo-postgres DSN lands on the EmberVM *serving* Envoy DaemonSet (different selector, port 5401), not the control plane. Enforcing the previously merged shape would have black-holed `/api/ember/postgres` as a silent dial timeout. Added as `embervmServing`.
- **There is no CNPG pooler.** Both `-ro` and `-rw` select the instance pods directly and the pod holds both credentials, so one postgres rule covers public_reader and public_writer.
- **Dead allow removed:** the signoz otel-collector 4317 target claimed an `OTEL_EXPORTER_OTLP_ENDPOINT` that does not exist on these pods; the public binary returns before `_setup_otel` (`framework/core.py`), verified independently.
- `host`/`remote-node` entities stay, port-pinned to 4318, because the frontend forwards browser traces to the hostNetwork OTLP agent.

Full destination inventory with file:line evidence for every dial (and the ruled-out ones: kube-apiserver, private-tier scrapers) is in the worktree report.

Guard test now renders the chart and pins the enforced shape: no cluster/world/all entity anywhere, broad policies gone, per-component allow sets equal to the audited inventory. `ci` green (772 pass).

Live-only confirmation after rollout: watch for policy-denied drops (the #4659 alert) and exercise `/api/ember/postgres` plus a chat turn.

### DIFF
--- a/projects/monolith-public/chart/BUILD
+++ b/projects/monolith-public/chart/BUILD
@@ -18,6 +18,24 @@ filegroup(
     visibility = ["//projects/monolith:__pkg__"],
 )
 
+# Guard (#5142): render inputs so the guard test can run `helm template` against
+# the REAL values (chart + deploy) and pin the enforced policy shape. Chart.lock
+# plus the charts/ glob are load-bearing: helm aborts when a locked dependency
+# has no packaged tarball on disk. This chart has no migrations/ directory, so
+# there is no migrations glob to carry (unlike projects/monolith/chart).
+filegroup(
+    name = "chart_render_inputs",
+    srcs = [
+        "Chart.yaml",
+        "Chart.lock",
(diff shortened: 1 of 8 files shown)


jomcgi:
## Helm Manifest Diff

PR #5316, opened 2026-08-25
feat(embervm): health tiers for durability signals

### LINKED ISSUE #4338: embervm: health tiers for durability signals (immediate on export failures, >24h on GC sweep stall)
Policy decided by Joe 2026-08-04 (ADR to follow, branch docs/adr-embervm-health-tiers): EmberVM health signals classify by time-to-impact, and BOTH tiers end in the health surface, not alert-only.

**Tier 1, immediate unhealthy (platform impact now):** sustained artifact export failures. During #4317, workspace.img PUTs 500'd for hours while the synthetic probe stayed green (it only exercises create/send); any session parking in that window rehydrates blank, which is user-visible data loss. Implementation sketch: the CP (or noded via CP aggregation) tracks export failure streaks per kind; the monolith health composite gains an ember-durability component that latches unhealthy on a sustained failure streak (a handful of consecutive failures, minutes not hours). Alternative or complement: a low-cadence durability synthetic (plant marker, destroy to force export, restore, verify) — note each run costs a session slot and a full workspace export (~10 GiB), so cadence must be slow and cleanup religious (parked sessions count against the workload cap).

**Tier 2, unhealthy after >24h sustained (longer horizon):** the S3 warmth GC failing to complete sweeps. Measure as the age of the newest gc-manifests/ object exceeding 24h + sweep interval: survives CP restarts, covers every abort reason (fleet-freshness, empty-store, list failures), and is exactly the signal that went unread for 11 days before #4317. 

**Also:** Discord alerts accompany both tiers (abort-streak and bucket-size-growth per ADR platform/007's observation that nothing alerts on object-storage growth), but the health surface is the system of record.

**Standing rules:** detectors land suspend: true / dark and flip only after live verification (feedback: verify-prober-reach); grep tests before changing any threshold numbers; new monolith endpoints reading cluster resources need matching ClusterRole verbs.

Related: #4317 (closed incident), #4334, #4336, #4337.

jomcgi:
Closes #4338. Implements ADR embervm/031: durability signals classify by time-to-impact and both tiers end in the health surface, never alert-only.

- **Tier 1 (immediate)**: `Embervm.Durability` aggregates per-kind artifact export failure streaks from the node capacity facts noded already reports. Crossing 10 consecutive rounds (~5 min) latches unhealthy, because a session parking in that window rehydrates blank (the #4317 data-loss shape the synthetic probe cannot see).
- **Tier 2 (>24h)**: age of the newest gc-manifests object beyond 24h + sweep interval. Reads durable S3 state, so it survives CP restarts and covers every GC abort reason by construction.
- Both surface on `GET /v1/health/durability`; the monolith gains a **fatal-tier** `ember-durability` component reading it.

**No vacuous green**: missing data reads `unknown`, never healthy. Unset `expectedNodes`, silent nodes, or an empty/failed manifest listing hold the tier at unknown, and a latched streak does not heal just because witnesses go quiet.

**Lands dark on both sides** (`durabilityHealth.enabled: ""`, `emberDurability.url: ""`), so nothing evaluates until you flip it with a values-only change after live verification. Worth doing in that order given the monolith component is fatal-tier.

Known gap, documented in the module: `SessionVolume` carries no exported flag on the wire, so a workspace-only export failure stays invisible to the CP; #4317 itself would still have fired within minutes because every other kind was failing fleet-wide.

Verified: new suites 26/0 and the durability router tests; full ExUnit failures baseline-compared against stashed-change runs and all pre-existing; `ci` green (218 pass, 87 executed).

### DIFF
--- a/projects/embervm/chart/templates/deployment.yaml
+++ b/projects/embervm/chart/templates/deployment.yaml
@@ -575,6 +575,36 @@ spec:
             - name: EMBERVM_WARMTH_S3_GC_INTERVAL_MS
               value: {{ .Values.warmthS3Gc.intervalMs | quote }}
             {{- end }}
+            {{- if .Values.durabilityHealth.enabled }}
+            # Durability health detector (#4338, ADR embervm/031): BOTH tiers
+            # (export-failure streaks; gc-manifests age vs 24h + sweep
+            # interval) surface on GET /v1/health/durability. Unset keeps the
+            # detector dark: no evaluation runs and the route 404s. See the
+            # durabilityHealth block in values for the flip procedure.
+            - name: EMBERVM_DURABILITY_HEALTH
+              value: {{ .Values.durabilityHealth.enabled | quote }}
+            {{- end }}
+            {{- if .Values.durabilityHealth.expectedNodes }}
+            # Tier 1's fleet contract: every listed node must report fresh
+            # capacity facts or tier 1 reads unknown (never green). Empty =
+            # tier 1 always unknown.
+            - name: EMBERVM_DURABILITY_EXPECTED_NODES
+              value: {{ join "," .Values.durabilityHealth.expectedNodes | quote }}
+            {{- end }}
+            {{- if .Values.durabilityHealth.roundIntervalMs }}
+            - name: EMBERVM_DURABILITY_ROUND_INTERVAL_MS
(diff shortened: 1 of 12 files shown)


jomcgi:
## Helm Manifest Diff

Issue #5300, opened 2026-08-25
embervm: primed VM can still be handed to a second session across a control-plane restart


jomcgi:
#5293 and its fix in #5296 narrow the primed-VM double-hand-out window but do not close it. Do not treat #5293 as having closed this.

Found during the review of #5296. Control-plane side, pre-existing.

## The residue

Between session create and that session's first operation, the primed VM still sits in `primed_vm_ids` on the control-plane side. So:

1. Session A is created and a primed VM is recorded for it.
2. The control plane restarts before A performs any operation.
3. On rebuild, the same primed VM can be re-adopted and handed to a NEW session B.
4. Both sessions now believe they own that `vm_id`. Whichever one banks first consumes it, and the other session's operations fail against a `vm_id` that noded no longer holds in either registry.

Before #5296 the window was wider, because a never-invoked VM could never be consumed by a bank at all. The fix shrinks it: a bank now consumes the VM, which closes the idle-forever case but leaves the restart case.

## Fix direction

The durable session row already records `vm_id` at create. The primed inventory and the session rows need to agree on ownership across a control-plane restart, so that rebuilding cannot hand an already-owned primed VM to a second session. Options worth weighing before implementing: drop primed entries that a live session row already claims during rebuild, or make adoption on the noded side reject a `session_id` that does not match the one the entry was first claimed for.

The second option is cheap to reason about because `claimForSession` already runs under the registry mutex, but it changes a daemon-side contract, so it wants a decision rather than a drive-by patch.

jomcgi:
Confirmed live, and the trigger is a bank rather than a control-plane restart. Raising priority: this now alternates the conformance gate red on every other cycle.

## Mechanism

`adopt_inventory` in `projects/embervm/control/lib/embervm/dispatcher.ex` (around line 1381) is additive only. Its own comment states it:

> ADDITIVE only (never drops on a status read) ... A vm_id the node destroyed out of band (base turnover) self-corrects on its next use (one failed-then-retried assign), not here.

Since #5296, `Bank` adopts a primed session VM, and `claimForSession` DELETES that vm_id from noded's task registry. noded then stops advertising it in `WorkloadCapacity.primed_vm_ids`. The control plane never drops it, so the next session is dispatched onto a vm_id that exists in neither of noded's registries.

The documented recovery, one failed-then-retried assign, does not exist on the bank path: the idle bank simply fails three times 20s apart and the session goes terminal `failed`.

## Evidence

[...]

Issue #5293, opened 2026-08-25
embervm: Bank refuses a never-invoked session VM, so an idle bank of a fresh session always fails


jomcgi:
`Bank` refuses a session VM that has never served a turn, so an idle-bank of a freshly created session always fails and conformance S2 can never establish its baseline.

## Symptom

Live in `embervm-dev` on chart 0.46.5/0.46.6, S2 is vacuous and S3/S4 fall over behind it:

```
S1 pass     guest exited 0 and printed conformance ok; VM reap observed in 1ms
S2 vacuous  bank never observed: session entered terminal state "failed"   (61s)
S3 vacuous  S2 did not establish a baseline
S4 fail     3 passing invariants with coverage, need 4
```

Control plane, three attempts 20s apart (`conformance.idleBankSeconds: 20`), then terminal:

```
embervm session bank failed
reason: GRPC.RPCError{status: 9, message: "noded: session vm \"vm-f53df1b191853c78\"
        not bankable (unknown, task-class, or a call is already in flight)"}
```

## Cause

noded keeps a session VM in one of two registries:

- `Prime` boots the VM into `s.vms` (the primed/task registry) and returns the `vm_id` the control plane durably records on the session row (`session_manager.ex`, `PrimeResponse.vm_id`).
- `adoptPrimedSession` promotes it `s.vms` to `s.sessionVMs`, and its ONLY caller is `SessionAssign` (`noded/server/server.go:1373`).
- `Bank` requires an entry in `s.sessionVMs` (`noded/server/server.go:1462`).

So a session that is created and then idles WITHOUT ever being invoked never leaves the primed registry, and its idle-bank is refused as "unknown". The VM is real, booted, and sitting one registry over.

`SessionAssign` already carries the fallback and its comment states the intent, that a freshly created session's VM "still lives in the task registry on its FIRST invoke". `Bank` never got the symmetric branch, even though `BankRequest` already carries both `session_id` and `trace.workload`, which is everything `adoptPrimedSession` needs.

This is required behavior rather than an edge case. `piRuntimeWorkload` is documented in `chart/values.yaml` as ADR 027's `memory:true` quadrant, where "an idle session banks a memory snapshot and relights from it", and conformance S2 exercises exactly that: create a `pi-runtime` session, never dispatch a turn, wait for the idle bank.

## Evidence that no turn ever ran

The `CapacityObserver` on the only dev brick reports, at S2 session creation and through every bank attempt:

```
[...]

PR #5319, opened 2026-08-25
fix(embervm): route the guest shim to the CLI its image actually ships

### LINKED ISSUE #5315: embervm: guest shim defaults to the claude CLI on images that do not ship it
`ProcessManager._adapter` in `projects/embervm/runtimes/claude/shim.py` falls back to `self.claude` for any model it does not recognise:

```python
def _adapter(self, model):
    if model == "qwen":
        return self.pi
    if isinstance(model, str) and model in CODEX_MODELS:
        return self.codex
    return self.claude
```

That fallback is a hardcoded binary name, not a property of the guest image. The `pi-runtime` image ships only `pi` (`projects/embervm/runtimes/pi/prewarm-clis`), so an invoke that omits `model` returns:

```
422 {"error":"[Errno 2] No such file or directory: 'claude'"}
```

Observed live in embervm-dev across several conformance cycles, for example at 13:05:45 and 14:07:23 on 2026-08-25.

## Why it holds the conformance gate red

Conformance S2 (`projects/embervm/conformance/scenarios.go`) invokes a relit session to prove the relight worked. `classifyInvokeResponse` passes the scenario on a 2xx OR on a model-unreachable marker, because dev has no reachable model. A missing-binary 422 is neither, so S2 fails.

The harness clearly expects the pi adapter on this workload: one of its accepted model-unreachable markers is the pi-specific string `pi turn produced no output`.

## Fix direction

Resolve the fallback against the CLIs the guest image actually carries rather than hardcoding claude. `ProcessManager` already reads that set into `self._prewarm_clis` via `_read_prewarm_clis()`. Keep claude as the fallback whenever the image lists it, so the claude-runtime guest (`claude,codex,pi`) is unaffected.

Do not fix this by having the conformance runner send a `model` field: any real caller omitting `model` against a pi-runtime guest hits the same 422.

## Related

Distinct from #5293 and #5296. Banking now works; this is the next failure S2 reaches. Blocks #5224.

jomcgi:
Fixes #5315.

`ProcessManager._adapter` fell back to a hardcoded `self.claude` for any model it did not recognise. That is a property of the binary name rather than of the guest image, so a `pi-runtime` session invoked without a `model` field returned:

```
422 {"error":"[Errno 2] No such file or directory: 'claude'"}
```

The pi-runtime image ships only `pi` (`projects/embervm/runtimes/pi/prewarm-clis`).

## Why it matters now

This is what conformance S2 hits immediately after banking starts working (#5296). S2 invokes a relit session to prove the relight happened, and `classifyInvokeResponse` passes on a 2xx or on a model-unreachable marker, because dev has no reachable model. A missing-binary 422 is neither, so S2 fails and the Kargo dev to prod gate stays red.

That the harness expects the pi adapter here is not an inference: one of its accepted markers is the pi-specific string `pi turn produced no output`.

## Change

Resolve the fallback against the CLIs the image actually carries, using the `_prewarm_clis` set `ProcessManager` already reads. When the image lists `claude` the behaviour is unchanged, so the claude-runtime guest (`claude,codex,pi`) is untouched. An image with exactly one CLI and no claude routes to that one. Empty or ambiguous multi-CLI sets keep the claude fallback rather than becoming unpredictable.

Explicit `qwen` and `CODEX_MODELS` routing keep priority over the fallback.

`getattr(self, name)` is safe here because `_read_prewarm_clis` raises `StartupError` on anything outside `SUPPORTED_PREWARM_CLIS`, and `ready()` already uses the same pattern.

## Verification

The regression test was falsified rather than assumed. With the test file unchanged and ONLY the three-line fix reverted:

```
projects/embervm/runtimes/claude/shim_test.py::test_manager_routes_unspecified_model_to_only_configured_cli FAILED
Executed 1 out of 137 tests: 136 tests pass and 1 fails remotely.
```

With the fix restored: `Executed 1 out of 139 tests: 139 tests pass.`

Five cases covered: single-CLI routing, claude-runtime unchanged, qwen and codex routing under both configurations, empty set, and the ambiguous multi-CLI set.

## Not this

Distinct from #5293/#5296 (banking) and #5300 (the control plane recycling a consumed vm_id). Fixing this alone will not turn the gate green; #5300 is the other half.

### DIFF
--- a/projects/embervm/runtimes/claude/shim.py
+++ b/projects/embervm/runtimes/claude/shim.py
@@ -3742,6 +3742,9 @@ class ProcessManager:
             return self.pi
         if isinstance(model, str) and model in CODEX_MODELS:
             return self.codex
+        configured_clis = getattr(self, "_prewarm_clis", ())
+        if "claude" not in configured_clis and len(configured_clis) == 1:
+            return getattr(self, configured_clis[0])
         return self.claude
 
     def ready(self):
--- a/projects/embervm/runtimes/claude/shim_test.py
+++ b/projects/embervm/runtimes/claude/shim_test.py
@@ -753,6 +753,48 @@ def test_manager_routes_qwen_to_pi(tmp_path, monkeypatch):
     assert manager._adapter(None) is manager.claude
 
 
+def _adapter_manager(prewarm_clis):
+    manager = _new_process_manager()
+    manager._prewarm_clis = prewarm_clis
+    manager.claude = object()
+    manager.codex = object()
+    manager.pi = object()
+    return manager
+
+
+def test_manager_routes_unspecified_model_to_only_configured_cli():
+    manager = _adapter_manager(("pi",))
+
+    assert manager._adapter(None) is manager.pi
+
+
+def test_manager_routes_unspecified_model_to_claude_when_configured():
+    manager = _adapter_manager(("claude", "codex", "pi"))
+
+    assert manager._adapter(None) is manager.claude
+
+
+@pytest.mark.parametrize("prewarm_clis", [("pi",), ("claude", "codex", "pi")])
+def test_manager_explicit_model_routing_precedes_cli_fallback(prewarm_clis):
+    manager = _adapter_manager(prewarm_clis)
+    codex_model = next(iter(shim.CODEX_MODELS))
+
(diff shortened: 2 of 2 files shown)


jomcgi:
## Helm Manifest Diff

PR #5322, opened 2026-08-25
fix(embervm): drop a banked or destroyed vm from the dispatcher inventory

### LINKED ISSUE #5300: embervm: primed VM can still be handed to a second session across a control-plane restart

jomcgi:
Fixes #5300.

`adopt_inventory` in `dispatcher.ex` is additive only, and deliberately so. Its own comment gives the reason a blanket drop on a status read would be wrong: a vm_id reserved for an in-flight assign is still reported primed by the node until the Assign lands, so dropping on status would race the reserve and could double-dispatch.

Since #5296 a `Bank` adopts a primed session VM, and `claimForSession` DELETES that vm_id from noded's task registry, so noded stops advertising it. The control plane kept it forever and dispatched a later session onto a vm_id that exists in neither of noded's registries. Its idle bank then failed three times 20s apart until the session went terminal `failed`. The recovery the comment relies on, "one failed-then-retried assign", does not exist on the bank path.

## Live evidence

Alternating conformance results in embervm-dev on 2026-08-25, noded stable throughout:

```
13:05  S2 fail (guest 422, #5315)        bank SUCCEEDED, consumed the primed VM
13:35  S2 vacuous "bank never observed"  dispatched onto the consumed vm_id
14:06  S2 fail (guest 422, #5315)        primed fresh, bank succeeded again
```

In the 13:35 cycle `adopt_idempotent` is back to `vacuous(coverage=0)` and `prime_before_checkpoint` to `vacuous(coverage=13)`, the pre-#5296 pattern.

## Change

Drop the vm_id where the control plane KNOWS it consumed the VM, rather than inferring a removal from a status snapshot:

- a successful bank, via the `vm_id` now carried in the `{:bank_done, ...}` completion message
- a confirmed destroy, which had the same staleness and no existing removal

Both are consumption events the control plane performed itself, so no in-flight assign can be racing them. `adopt_inventory` stays additive and the reserve race it guards against stays guarded (there is a test for that).

The removal is idempotent: an unknown vm_id is a no-op.

## On the synchronous call

`drop_vm` is a `GenServer.call` so a placement after a confirmed consumption cannot race stale inventory, but its exit is caught. Both callers run inside the SessionManager's serialized loop, where a `GenServer.call` that EXITS (a dispatcher busy enough to exceed the call timeout) would take down session lifecycle for what is only a bookkeeping cleanup. This file already carries a comment about being bitten by exactly that hazard on the node-channel call.

Catching costs nothing functionally: a timed-out call is still delivered and still handled, so the drop lands either way. Only this one caller loses its synchronous guarantee, in the rare case where it was going to crash instead.

## Verification

The regression tests were falsified rather than assumed. With the tests unchanged and ONLY `drop_vm_id`'s body neutered (API kept, so failures are assertions rather than a compile error):

```
1) test dropping an adopted vm clears provenance and later adoption still works (Embervm.DispatcherTest)
2) test a successful bank drops stale dispatcher inventory before another placement (Embervm.SessionManagerTest)
3) test a confirmed destroy drops stale dispatcher inventory (Embervm.SessionManagerTest)
1518 tests, 3 failures, 22 skipped
```

Restored: `1518 tests, 0 failures, 22 skipped`.

Note for reviewers: `ci test` gives NO signal on this suite. The ExUnit suite is a genrule, so `bazel test` reports "Found 1 target and 0 test targets" and the run looks cached-green. It has to be built (`bb remote ... build //bazel/erlang:mix_test_smoke`) and judged by its `N tests, N failures` line.

## Not this

#5319 fixes the guest shim routing that S2 hits next. Neither PR turns the gate green alone.

### DIFF
--- a/projects/embervm/control/lib/embervm/dispatcher.ex
+++ b/projects/embervm/control/lib/embervm/dispatcher.ex
@@ -160,6 +160,36 @@ defmodule Embervm.Dispatcher do
   end
 
   @doc """
+  Removes a VM consumed by another control-plane lifecycle path from the primed
+  inventory and its adoption-provenance set. The operation is idempotent, so an
+  unknown VM is a no-op. This is the targeted consumption seam from #5300.
+
+  This is synchronous so a placement after a confirmed consumption cannot race
+  stale dispatcher inventory.
+
+  The exit is caught deliberately. Both callers run inside the SessionManager's
+  serialized loop, where a GenServer.call that EXITS (a dispatcher busy enough to
+  exceed the call timeout) would take down session lifecycle for what is only a
+  bookkeeping cleanup. A timed-out call is still DELIVERED and still handled, so
+  swallowing the exit costs this one caller its synchronous guarantee and nothing
+  else: the drop still lands.
+  """
+  @spec drop_vm(GenServer.server(), String.t()) :: :ok
+  def drop_vm(server \\ __MODULE__, vm_id) do
+    case GenServer.whereis(server) do
+      nil ->
+        :ok
+
+      _pid ->
+        try do
+          GenServer.call(server, {:drop_vm, vm_id})
+        catch
+          :exit, _reason -> :ok
+        end
+    end
+  end
+
+  @doc """
   Whether `principal` may submit another task to `workload` without breaching its
   per-principal queue-depth cap. A read-only advisory check the router runs
   BEFORE the durable submit to 429; the actual reservation happens on `enqueue`,
@@ -322,6 +352,10 @@ defmodule Embervm.Dispatcher do
 
   @impl true
   def handle_call(:stats, _from, state), do: {:reply, snapshot(state), state}
+
+  def handle_call({:drop_vm, vm_id}, _from, state) do
+    {:reply, :ok, drop_vm_id(state, vm_id)}
+  end
 
   def handle_call(:sweep, _from, state) do
     {:reply, :ok, run_sweep(state)}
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

PR #5323, opened 2026-08-25
fix(embervm): adopt a sibling brick's published base instead of wedging BuildBase

### LINKED ISSUE #4866: embervm: co-located bricks share scratch but not base registries, a sibling's published bundle wedges BuildBase at the publish rename
## Status 2026-08-22

Still open and still true on main. `driveBuild` in `projects/embervm/noded/server/server.go` (around line 708) short-circuits `AlreadyBuilt` only on the in-memory `s.bases.get(baseKey)`, and the publish `os.Rename(buildingDir, finalDir)` in `projects/embervm/noded/fcvm/driver/driver.go` (around line 1382) still refuses an existing destination. Neither consults disk. The base-scratch and ownership follow-ups (#4960 closed, #4962 closed) did not touch this path.

Files: `projects/embervm/noded/server/server.go` (`driveBuild`, the inventory completeness rule near the `ReconcileBasesFromDisk` scan), `projects/embervm/noded/fcvm/driver/driver.go` (publish rename), `projects/embervm/noded/server/server_test.go`.

Acceptance:
- a unit test that pre-creates a complete bundle at `finalDir` (imageref, memfile, snapfile) and calls `BuildBase`: returns `AlreadyBuilt: true` with the bundle registered READY, without booting a build guest
- a unit test with an incomplete `finalDir`: the stale directory is removed and the build proceeds
- the publish rename never returns `file exists` for a complete sibling bundle

---

## Defect

Two brick pods co-located on one node share the same hostPath scratch (`/var/lib/embervm/scratch/embervm-noded/`), but each noded keeps a PER-PROCESS base registry, and both the BuildBase idempotency gate and the publish rename trust only that in-memory view:

- `driveBuild` short-circuits `AlreadyBuilt` only on `s.bases.get(baseKey)` (`projects/embervm/noded/server/server.go`, driveBuild), never consulting the disk.
- The final publish `os.Rename(buildingDir, finalDir)` refuses an existing destination (`projects/embervm/noded/fcvm/driver/driver.go:1236`), which is correct against partial bundles but turns a sibling's completed publish into a permanent build failure.
- `ReconcileBasesFromDisk` runs at daemon start and after restores, so a bundle published by a sibling (or a previous incarnation) AFTER this daemon's startup scan is invisible until the next restart.

## Observed 2026-08-14 (fleet roll for the amd64-only image rebuild)

The pre-restart control plane dispatched a semgrep base build to the then-live `brick-16gi-node-4`, which published `semgrep__076cff3e18ee` and then rolled; its successor adopted the bundle at 14:50:38 and serves restores from it. Meanwhile the sibling `brick-16gi` (same node, fresh scratch scan at 14:38:35, bundle not yet on disk) had been given the SAME build at 14:39:58, spent ~11 minutes on it (bazel warming), and died at the publish rename at 14:51:05:

```
BuildBase for embervm/semgrep failed: noded: build base "semgrep__076cff3e18ee":
snapshot: driver: publish base bundle: rename .../bases/semgrep__076cff3e18ee.building
.../bases/semgrep__076cff3e18ee: file exists
```

The workload stayed `BaseReady: True` throughout (the sibling serves it), so this is waste and noise rather than an outage: every control-plane retry on `brick-16gi` burns a full multi-minute 16 GiB build VM and fails at the same rename, until that pod restarts and its startup scan adopts the bundle.

Not the cause of #4865 (demo-postgres never issues a BuildBase at all; that is a control-plane anchor-recovery defect tracked there), but surfaced by the same fleet roll.

## Fix shape

On publish collision (or better, in the driveBuild idempotency gate before booting anything): if the destination bundle exists on disk and is complete (`imageref`, `memfile`, `snapfile`, the same completeness rule the inventory scan uses at `server.go:1940`), adopt it (register READY, as ReconcileBasesFromDisk would) and return `AlreadyBuilt` instead of failing. An incomplete destination is the existing half-written case and should be removed before the rename rather than wedging. Either way the invariant becomes: a completed bundle on shared scratch is a cache hit for every brick on the node, not just the one that wrote it.

## Immediate operational step

Restart the `brick-16gi` pod on node-4; its startup reconcile adopts the bundle and the retry loop ends.

jomcgi:
Closes #4866.

Co-located bricks share the node's base scratch but keep per-process registries, so a sibling's completed build was invisible to both the BuildBase idempotency gate and the publish rename: every control-plane retry burned a full multi-minute build guest and died at `rename(bases/<key>.building, bases/<key>)` with "file exists" until a pod restart finally rescanned the bundle.

Two layers:

- `driveBuild` adopts a COMPLETE bundle for the same derived key straight off disk (same completeness rule as the inventory scan, plus the #4407 snapshot-format check and the rootfs rule) and reports `AlreadyBuilt` without booting anything.
- `SnapshotBase` resolves a publish collision instead of failing on it: a complete destination is adopted, a stale one is cleared and the rename retried.

Review notes: the worker skipped local verification, and its own new test then failed in CI (adoption declined because the fixture's `RootfsPath` pointed at a nonexistent `/rootfs.ext4`, so a build guest still booted). Fixed here in follow-up commits: the fixture now creates a real backing rootfs, and the identity assertion compares against it. The feature itself is confirmed by that same test now passing: adoption fires, the bundle registers READY at the sibling's size, and the build driver records zero claims and zero snapshots.

pr-checks and semgrep green.

Implemented by ox-alpha via opencode, corrected and reviewed on Opus.

### DIFF
--- a/projects/embervm/noded/fcvm/driver/driver.go
+++ b/projects/embervm/noded/fcvm/driver/driver.go
@@ -1437,7 +1437,21 @@ func (d *Driver) SnapshotBase(ctx context.Context, h substrate.Handle, baseKey s
 		return substrate.SnapshotRef{}, fmt.Errorf("driver: publish base snapfile: %w", err)
 	}
 	if err := os.Rename(buildingDir, finalDir); err != nil {
-		return substrate.SnapshotRef{}, fmt.Errorf("driver: publish base bundle: %w", err)
+		// #4866: co-located bricks share this scratch dir but not their
+		// per-process base registries, so two bricks can be handed the SAME
+		// BuildBase while only one wins the publish. Resolve the collision
+		// instead of wedging on it (every control-plane retry would otherwise
+		// burn a full multi-minute build VM and die at this same rename).
+		adopted, cerr := resolveBasePublishCollision(buildingDir, finalDir)
+		if cerr != nil {
+			return substrate.SnapshotRef{}, fmt.Errorf("driver: publish base bundle: %w", errors.Join(err, cerr))
+		}
+		if !adopted {
+			// The destination was cleared as stale; retry the publish once.
+			if rerr := os.Rename(buildingDir, finalDir); rerr != nil {
+				return substrate.SnapshotRef{}, fmt.Errorf("driver: publish base bundle: %w", rerr)
+			}
+		}
 	}
 	return substrate.SnapshotRef{
 		ID:        baseKey,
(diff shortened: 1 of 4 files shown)


jomcgi:
## Helm Manifest Diff

<details>

```diff

spec.configuration.settings  (clickhouse.altinity.com/v1/ClickHouseInstallation/signoz/signoz-clickhouse)
  - one map entry removed:
    merge_tree/max_suspicious_broken_parts: 200
```

</details>

PR #5325, opened 2026-08-25
fix(ci): surface tool-build failures instead of a bogus missing-tool error

### LINKED ISSUE #4749: ci: chart-diff stage reports "DYFF not found in bazel-bin" when the tool build failed
## Status 2026-08-22

The proposed fix is already in place and has been since the script was created (commit 8572f5c8a, 2026-03-11): `bazel/helm/ci-diff-manifests.sh:10` is `set -euo pipefail`, so a failed `bazel build ... | tail -1` DOES abort the script. That falsifies the diagnosis below: on #4748 the build exited 0 and `find -L "$BAZEL_BIN/external" -name dyff` still returned nothing.

So the real defect is in tool discovery, not exit masking. Candidates: the `bazel-bin` convenience symlink pointing at a different output base than `--config=ci` wrote to (the comment at line 30 already warns `bazel info bazel-bin --config=ci` is unusable), or the multitool output layout not matching `-name dyff -type f`.

**Remaining work (one PR):**

1. Replace the four `find` lookups with `bazel cquery --config=ci --output=files @multitool//tools/<tool>` (or read the `bazel build --config=ci ... --build_event_json_file` outputs), so the path comes from the build that just ran rather than from a symlink walk.
2. On a lookup miss, print `ls -la "$BAZEL_BIN"` and `find "$BAZEL_BIN/external" -maxdepth 3 -name '*dyff*'` before exiting so the next occurrence is diagnosable.
3. Keep the full build output on failure instead of `tail -1` (write to a temp file, cat it when the pipeline fails).

**Acceptance:** the chart-diff stage on a PR still passes; a deliberately broken tool target (for example a bad label) fails with the bazel error text visible, not "DYFF not found in bazel-bin".

---

`bazel/helm/ci-diff-manifests.sh` swallows the exit status of its tool build, so
a build failure surfaces several lines later as a missing-tool error that names
the wrong problem.

## The lines

```bash
bazel build @multitool//tools/helm @multitool//tools/dyff @multitool//tools/gh @multitool//tools/yq --config=ci 2>&1 | tail -1
BAZEL_BIN="$(git rev-parse --show-toplevel)/bazel-bin"

HELM=$(find -L "$BAZEL_BIN/external" -name "helm" -type f -perm /111 2>/dev/null | head -1)
DYFF=$(find -L "$BAZEL_BIN/external" -name "dyff" -type f -perm /111 2>/dev/null | head -1)
...
    echo "ERROR: $tool_name not found in bazel-bin"
```

A pipeline returns the status of its LAST command, so `tail` succeeding means a
failed `bazel build` passes silently. Whatever bazel printed about the failure
goes into a `tail -1` that discards it.

## What it looked like

On #4748 (a vendored chart bump, no Elixir, no Python):

```
==> Building tools...
  HELM: /home/buildbuddy/workspace/repo-root/bazel-bin/external/.../helm
ERROR: DYFF not found in bazel-bin
Command failed: exit status 1
```

That message sends you to look at multitool config and lockfiles. `dyff` is
correctly declared there:

```
MODULE.bazel:194  "multitool.dyff.linux_arm64",
MODULE.bazel:195  "multitool.dyff.linux_x86_64",
bazel/tools/tools.lock.json:146  "dyff": { ... }
```

and #4746, also a chart change, passed the same stage an hour earlier. So the
tool is not missing and the real event was a transient build failure whose
reason was thrown away. Re-running the PR cleared it.

Note that HELM resolved and DYFF did not **in the same build invocation**,
which is itself a hint the build partially failed rather than the toolchain
being wrong.

## Fix

One line. Either:

```bash
set -o pipefail   # near the top, with the existing set -euo
```

or check explicitly:

```bash
bazel build ... --config=ci 2>&1 | tail -1
[ "${PIPESTATUS[0]}" -eq 0 ] || { echo "ERROR: tool build failed (see above)"; exit 1; }
```

The value is diagnosis rather than correctness: the stage already fails, it
just fails while describing something that is not true. Worth pairing with
keeping more than `tail -1` of the build output on failure, since one line of a
bazel failure is rarely the useful one.

## Related

This is the same class the repo already guards against for `ci` / `bb remote`
invocations via `bazel/tools/hooks/check-ci-pipe-mask.sh`, which blocks piping
their output into a filter for exactly this reason. That hook covers what
humans and agents type; it does not cover pipes inside CI scripts.

jomcgi:
Closes #4749.

The chart-diff stage piped the tool build into `tail -1` and then resolved binaries by walking the `bazel-bin` convenience symlink. Two failure modes both ended at the same misleading message:

- a failed build showed one arbitrary line, then fell through to "DYFF not found in bazel-bin"
- a *successful* build could still miss tools when the symlink pointed at a different output base than `--config=ci` wrote to

Phase 1 now captures the whole build log and, on non-zero exit, prints it under `ERROR: tool build FAILED` and exits, so a build failure is reported as a build failure. Tool paths resolve via `bazel cquery --config=ci --output=files` (the artifacts the build just produced), with the old find walk kept only as a fallback.

Verified both paths against a scratch copy of the script with a stubbed `bazel`: the failure path prints the real bazel error text and never reaches the missing-tool message; the success path resolves all four tools and behaves as before. Both runs are pasted in the worktree report. Only CI can exercise the real `cquery` path resolution. `ci` green (774 pass).

### DIFF
--- a/bazel/helm/ci-diff-manifests.sh
+++ b/bazel/helm/ci-diff-manifests.sh
@@ -31,21 +31,87 @@ main() {
 	# (.bazelrc `build:ci --disk_cache=`). BuildBuddy's workflow pool is
 	# mixed-arch, so a plain build pulls a wrong-arch helm/dyff/gh out of that
 	# cross-arch-contaminated disk cache and it dies with "cannot execute binary
-	# file: Exec format error" (why the manifest comment never posted). Read the
-	# output path from the bazel-bin convenience symlink, NOT `bazel info
-	# bazel-bin --config=ci` (that fails to resolve @@buildbuddy_toolchain).
-	bazel build @multitool//tools/helm @multitool//tools/dyff @multitool//tools/gh @multitool//tools/yq --config=ci 2>&1 | tail -1
+	# file: Exec format error" (why the manifest comment never posted).
+	#
+	# Capture the whole build log instead of piping into `tail -1` (#4749):
+	# one line of a bazel failure is rarely the useful line, and a failed
+	# build must be reported AS a build failure, never fall through to the
+	# missing-tool error below.
+	TOOL_BUILD_LOG="$TMPDIR_BASE/tool-build.log"
+	if ! bazel build @multitool//tools/helm @multitool//tools/dyff @multitool//tools/gh @multitool//tools/yq --config=ci >"$TOOL_BUILD_LOG" 2>&1; then
+		echo "ERROR: tool build FAILED (bazel exited non-zero), full output:"
+		cat "$TOOL_BUILD_LOG"
+		exit 1
+	fi
+	tail -1 "$TOOL_BUILD_LOG"
 	BAZEL_BIN="$(git rev-parse --show-toplevel)/bazel-bin"
 
-	HELM=$(find -L "$BAZEL_BIN/external" -name "helm" -type f -perm /111 2>/dev/null | head -1)
-	DYFF=$(find -L "$BAZEL_BIN/external" -name "dyff" -type f -perm /111 2>/dev/null | head -1)
-	GH=$(find -L "$BAZEL_BIN/external" -name "gh" -type f -perm /111 2>/dev/null | head -1)
-	YQ=$(find -L "$BAZEL_BIN/external" -name "yq" -type f -perm /111 2>/dev/null | head -1)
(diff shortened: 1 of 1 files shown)


jomcgi:
## Helm Manifest Diff

PR #5326, opened 2026-08-25
fix(ci): size deploy above a wide apko rebuild so the chart publish survives

### LINKED ISSUE #4871: ci: the deploy action can exceed BuildBuddy's 1 hour wall when many apko images rebuild, losing the chart publish
## Status 2026-08-22

Still open. `buildbuddy.yaml` sets no `timeout` on either action (grep is empty), and the `deploy` action's ordering is unchanged: image pushes first, chart publish and write-back last. The move to the GitHub merge queue (e8d15bad9) touched only `pr-checks`' format-drift handling and the push URL. A wide apko lock refresh still has the same shape.

Files: `buildbuddy.yaml` (`deploy` action: `timeout`, stage ordering), `docs/runbooks/argocd-outofsync.md` (add this as the third "merged but not deployed" cause).

Acceptance, minimum viable (option 1): an explicit `timeout` on `deploy` sized for a 14-image cold rebuild plus publish, and the runbook entry. Option 2 (publish in its own action or stage that runs even when image work is truncated) is better and should be scoped as its own PR if picked; it is Bazel/BuildBuddy workflow wiring, which CLAUDE.md keeps on Opus.

---

The `deploy` action for `1e63fee7653aecca8cd2f7c1a4c9265133042cbe` ran for **exactly 3600 seconds** and reported `Failed`. That is BuildBuddy's default workflow timeout, not a build error: `buildbuddy.yaml` sets no `timeout` on either action.

Invocation `06feb469-7a7a-4fc5-b22f-11d7a89ba859`, `createdAtUsec` 1786719344953000, `updatedAtUsec` 1786722946635000.

## What was lost

All 14 `image.push` child invocations **succeeded**, spanning roughly 52 minutes. The images are in the registry. What never ran is the stage that follows them: computing chart versions, publishing the OCI charts, and committing the write-back.

So the symptom is the familiar "merged but not deployed", with a cause that is neither of the two already documented in the runbook. It is not the ruleset rejection (#3 in that list) and not the digests-match early exit (#2). It is simply that the action ran out of wall clock before reaching the publish.

## What triggered it

#4862 regenerated 14 apko locks, which moved x86_64 package versions (`glibc 2.43-r11` to `r13`, `libcrypto3 3.6.3-r3` to `r4`, and others). Every one of those images therefore had to rebuild from scratch rather than hit cache, and 14 cold apko builds plus pushes consumed the budget.

This is not specific to that PR. **Any** change that invalidates a large number of apko images has the same shape: a Renovate lock refresh, a base image bump, or the weekly `renovate-apko-lock-maintenance` PR.

## Why it is not worse than it looks

I checked for a registry orphan, which is the dangerous outcome when a publish is interrupted midway. There is none: the highest published `charts/embervm` version is `0.12.9`, exactly matching main's `Chart.yaml`. The timeout landed before anything was published, so the state is consistent and the next successful deploy simply computes and publishes.

That is luck rather than design. A timeout a few minutes later could have published charts without the write-back, which is the orphan case that hard-stops a future publish.

## Options

1. Set an explicit `timeout` on the `deploy` action above the default, so a wide image rebuild fits. Cheapest, but only moves the wall.
2. Split image build and push from chart publish and write-back, so a slow build cannot starve the publish. The publish is the part that must not be skipped.
3. Make the publish stage resumable, or ensure it runs even when the image stage is truncated.

Worth noting the ordering is currently the wrong way round for safety: the expensive, retryable work runs first, and the cheap, must-not-be-skipped work runs last.

jomcgi:
Closes #4871.

main's deploy for 1e63fee7 was killed at exactly 3600s by BuildBuddy's default workflow wall clock: all 14 `image.push` children were green after ~52 minutes of cold apko rebuilds (the #4862 lock refresh), and the chart publish plus write-back never ran. The images shipped and nothing deployed.

Sets an explicit `timeout: "2h"` on the deploy action, sized from the observed stages (~10m warm main suite + ~55m cold 14-image rebuild + ~15m publish/write-back/format) with deliberate headroom: a kill landing *during* the publish is strictly worse than a long run, since charts can reach the registry without the write-back commit and wedge later publishes. Every stage is idempotent, so a re-run after truncation converges.

Also documents this as the third "merged but not deployed" cause in the ArgoCD out-of-sync runbook.

Verified: the `timeout` field name checked against BuildBuddy's own per-action struct tag, and the config round-tripped through a yaml.v2 unmarshal to confirm deploy carries `2h` while pr-checks is untouched. Caveat recorded in the file: an org-level `remote_execution.remote_runner_default_timeout` experiment clamps requests at or above its value, so if a wide rebuild still dies near one hour, that cap is the cause rather than this setting. `ci` green (774 pass).


jomcgi:
## Helm Manifest Diff

PR #5337, opened 2026-08-26
feat(ci): gate chart admissibility with kubeconform

### LINKED ISSUE #4831: CI renders charts but never asks the API server if the result is admissible
## Status 2026-08-22

Still open, nothing landed. No `kubeconform` or `kubectl --dry-run=server` anywhere in the repo (checked `bazel/`, chart dirs, `buildbuddy.yaml`). The helm-manifest-diff PR check also skips multi-source apps (embervm, monolith, monolith-public), so the charts most likely to carry this defect class get the least rendering coverage in CI.

Decision on the open question: the Test action has no cluster access, so server-side dry run is unreachable; offline schema validation is the only option reachable today and it would have caught the `secretKeyRef.key: Required value` case (`key` is required in `SecretKeySelector`).

Files:
- `projects/*/chart/` plus each chart's `deploy/values.yaml` and `dev/deploy/values.yaml` (the real value stacks)
- `MODULE.bazel` / `bazel/` for vendoring a `kubeconform` binary (Bazel toolchain work, which CLAUDE.md routes to Opus rather than a Codex implementer)
- CRD schemas: `OnePasswordItem`, `HTTPRoute`, `SecurityPolicy`, `CiliumNetworkPolicy`, `Workload`, Kargo and ArgoCD kinds will need schemas or a `-skip` list; `-ignore-missing-schemas` is the honest first step

Acceptance:
- one Bazel test target renders every chart against its prod and dev value stacks and pipes the result through `kubeconform -strict`
- a fixture reproducing #4830 (a `secretKeyRef` with an empty `key`) fails the target
- the target runs under `ci test` and shows in `Executed N out of M`

---

## What happened

PR #4830 fixes a duplicate `chat:` key in `projects/monolith/chart/values.yaml` that made `GITHUB_API_TOKEN` render as:

```yaml
valueFrom:
  secretKeyRef:
    name: monolith-dev-chat-secrets
    key:
```

`helm template` renders this happily, because it is structurally valid YAML. The Kubernetes API server rejects it (`secretKeyRef.key: Required value`), so every ArgoCD sync of that chart version failed, `monolith-dev` sat `OutOfSync`, four Kargo dev promotions timed out in `argocd-wait`, and prod stayed stranded on `0.304.9` for roughly six hours. Nothing alerted.

## The gap

Chart tests today verify that templates render and that rendered content matches expectations. None of them verify that the rendered manifests are **admissible**. That is a different question, and it is the one that decides whether a deploy lands.

## Rough shape of a fix

A test that runs `helm template` for each chart's real value stacks and pipes the result through `kubectl apply --dry-run=server`, or an equivalent schema validation (`kubeconform` with the cluster's CRD schemas) for the offline case. Server-side dry run is stronger, since it catches admission webhooks and required-field validation, but it needs cluster access from the test action, which CI does not have today. Offline schema validation would have caught this specific defect, since `key` is `required` in the `SecretKeySelector` schema.

Worth deciding which of the two is actually reachable from the Test action before scoping.

## Related

- #4830 the fix
- #4821 `RespectIgnoreDifferences=true`, a real and separate sync problem found while diagnosing this

jomcgi:
Closes #4831.

CI rendered charts but never asked whether the result would be admitted. New `//bazel/helm:chart_admissibility_test` validates all 28 rendered chart x value-stack manifest sets against Kubernetes schemas, with a testdata chart carrying a deliberate defect as a negative control.

**It is hermetic.** The first cut used kubeconform's default remote schema fetch, which passed on a laptop and failed every set in CI's network-less sandbox with DNS errors (`Valid: 0, Errors: 27`) while the negative control reported INCONCLUSIVE. Schemas are now vendored at `bazel/helm/k8s-schemas`: the standalone-strict set for the cluster's 1.35.8, from `yannh/kubernetes-json-schema` pinned at commit `5a69f83`, pruned to the kinds this repo actually renders (4.7 MB instead of 250 MB), with per-file sha256s the test re-verifies each run.

**It can fail, for the right reason.** The negative control is rejected specifically on `secretKeyRef/key: got null, want string`, and a guard fires if the fixture is ever rejected for some other reason (that guard is what caught the non-hermetic bug). CRD kinds are skipped by name rather than with a blanket `-ignore-missing-schemas`, which would have quietly restored the no-op; an unrecognised built-in kind fails loudly.

**It already caught a live defect**: keda set `prometheus.io/*` annotations the upstream chart also renders, producing duplicate YAML keys in one annotations map that the API server rejects. Fixed here.

Verified: offline run with proxies pointed at a closed port gives all 28 sets `Valid > 0, Errors: 0` and zero network attempts in the log; `ci` green (775 pass). Two follow-up commits from review: the vendored-schema fix above, and declaring the schema anchor file as a prerequisite so `$(rootpath)` resolves.

Implemented by ox-alpha via opencode, corrected and reviewed on Opus.


jomcgi:
Superseded by #6245 (merged 2026-09-19), which delivers the kubeconform admissibility gate against the current chart set with hermetic schemas pinned to the hub's 1.36.3 API server.

PR #5358, opened 2026-08-27
fix(embervm): open the egress lane for base-build guests

### LINKED ISSUE #5357: embervm: base-build guests have no egress lane, so claude-runtime cannot build a base
`claude-runtime` cannot build a base in production and has been down since 2026-08-26 16:47 UTC, retrying every 60s. The workload reports `Ready=False/BaseNotBuilt`, and neither 16gi node holds any claude base: the old one is gone and the new one cannot be produced.

## Root cause

Base-build guests get NO egress lane, and since #4423 the Claude CLI prewarm runs during base builds and its init requires network.

`startEgress` is called from exactly two places in `noded/server/server.go`: Prime (around line 1226) and Relight (around line 1655). It is never called from the base-build path, so `runBuild` boots a guest whose vsock egress port has no forwarder behind it.

Before `6ea698f0b` ("prewarm codex and pi CLIs like claude", #4423) the prewarm list came from `EMBER_PREWARM_CLIS`, which was rendered by nothing and therefore inert, so no CLI was prewarmed during a base build and the missing lane never mattered. That commit ships `prewarm-clis` (`claude,codex,pi`) INSIDE the image, so `ProcessManager.prewarm` now runs on every base build.

The guest's own serial log:

```
ember-claude-shim: cli-probe: exit=0 stdout='2.1.220 (Claude Code)\n' stderr=''
ember-claude-shim: egress listening on 127.0.0.1:1024
ember-claude-shim: listening on vsock port 1027
ember-claude-shim: egress vsock connect failed after 3 attempts: [Errno 104] Connection reset by peer
   (repeated ~20 times)
ember-claude-shim: initialization timed out after 30 seconds, sending SIGINT
ember-claude-shim: CLI prewarm failed: timed out waiting for Claude initialization after 30 seconds
```

The `--version` probe succeeds, so the binary is present and sound. The real spawn blocks waiting for Claude's `{"type":"system","subtype":"init"}` event, which does not arrive without network.

`prewarm` then sets `fatal_error`, and `ready()` returns False whenever `fatal_error` is set, so `/shim/ready` never flips. `runBuild`'s `WaitReady` times out with `guest readiness: vsockhttp: timed out waiting for guest ready`, the build fails, and the control plane retries forever.

## Note on the liveness gate

`ready()` also requires every configured prewarm CLI to be alive when the placeholder volume has no ext4, which is exactly the base-build condition. That guard is correct and deliberate: its comment says "a snapshot with a dead CLI would restore warmth that is not there." The bug is not the guard. The bug is that warming Claude needs egress and base builds have none, so the guard's precondition is unsatisfiable rather than merely unsatisfied.

## Fix

Open the egress lane for base-build guests, scoped exactly as Prime and Relight scope it.

`runBuild` (`noded/server/server.go` around line 959) already computes `uds := bd.VsockUDSPath(h.ThreadID)` and already has a teardown `defer` that Releases the guest. Start the forwarder just after the uds is known and cancel it in that same teardown.

`startEgress` is safe to call unconditionally: it returns a no-op cancel when `EgressEnabled` is false, and another when the workload is outside `EgressWorkloads`, logging the skip. So this widens nothing for a workload that was not already in scope. `claude-runtime` is already in prod's scope (`["claude-runtime","pi-runtime","shotter"]`).

`runBuild` does not currently receive the workload name. `baseKey` is `<workload>__<hash>`, but do NOT split the string: thread the workload through from the caller (around line 820), which has the request. `startEgress`'s own comment warns that a wrong or missing workload name is indistinguishable from a guest that was never meant to have egress, so passing it explicitly is the safe form.

## Scope note for review

This gives base-build guests the same network reach their Primed counterparts already have, for workloads already in the egress scope. It is a deliberate change to the base-build network posture and should be reviewed as one rather than as a bug fix, even though it restores behaviour that #4423 broke by accident.

## Alternatives considered

- Skipping prewarm during base builds: defeats the point of a warm base and contradicts the guard's stated intent.
- Making prewarm failure non-fatal during builds: knowingly ships bases whose warmth is a lie, which is what the guard exists to prevent.
- Reverting #4423's image-side prewarm: restores service but loses the codex and pi prewarm that commit wanted, and leaves the same trap for the next workload whose CLI needs network.

Found during the 0.41.7 to 0.47.11 prod jump (#5224). Twelve other workloads built cleanly.

jomcgi:
Fixes #5357. **Live production outage**: `claude-runtime` has been unable to build a base since 2026-08-26 16:47 UTC, retrying every 60s. `Ready=False/BaseNotBuilt`, and neither 16gi node holds any claude base.

## Cause

`startEgress` was called from Prime and Relight only, never from the base-build path, so a build guest booted with nothing behind its vsock egress port.

That was harmless until #4423 moved the prewarm list into the image. Prewarm now runs on every base build, and the Claude CLI's init blocks on network. The guest's own serial log:

```
ember-claude-shim: cli-probe: exit=0 stdout='2.1.220 (Claude Code)\n' stderr=''
ember-claude-shim: egress listening on 127.0.0.1:1024
ember-claude-shim: listening on vsock port 1027
ember-claude-shim: egress vsock connect failed after 3 attempts: [Errno 104] Connection reset by peer
   (x20)
ember-claude-shim: initialization timed out after 30 seconds, sending SIGINT
ember-claude-shim: CLI prewarm failed: timed out waiting for Claude initialization after 30 seconds
```

The `--version` probe succeeds, so the binary is sound. The real spawn blocks waiting for Claude's `{"type":"system","subtype":"init"}` event, which never arrives. `prewarm` sets `fatal_error`, `ready()` refuses while that is set, `/shim/ready` never flips, `WaitReady` times out, and the control plane retries forever.

## Change

Seven lines in `runBuild`. Start the forwarder as soon as the build guest's uds is known, which is before Prime, Hydrate and WaitReady, so it is listening before the shim dials. Cancel it first in the existing teardown, ahead of releasing the guest. The cancel initializes to a no-op so the defer is safe on every path, including a Claim failure.

The workload comes from `driveBuild`'s `req.GetTrace().GetWorkload()`, threaded explicitly rather than split out of `baseKey`. `startEgress`'s own comment warns that a wrong or missing workload name is indistinguishable from a guest that was never meant to have egress, so the name has to come from the request, exactly as Relight takes it.

## This widens nothing

`startEgress` already no-ops when `EgressEnabled` is false, and no-ops with a logged skip when the workload is outside `EgressWorkloads`. Build guests now get the same reach their primed counterparts already have, for workloads already in scope. `claude-runtime` is already in prod's scope (`["claude-runtime","pi-runtime","shotter"]`).

That said, please review this as a deliberate change to the base-build network posture rather than as a pure bug fix. It restores behaviour #4423 broke by accident, but the posture change is real.

## The prewarm gate is deliberately untouched

`ready()` requires every configured prewarm CLI alive when the placeholder volume has no ext4, which is the base-build condition. That guard is correct: "a snapshot with a dead CLI would restore warmth that is not there." The bug was that warming Claude needs egress and base builds had none, making the precondition unsatisfiable rather than merely unsatisfied. Weakening the guard would ship bases whose warmth is a lie.

## Verification

Falsified rather than assumed. With the tests unchanged and ONLY the `startEgress` call removed:

```
--- FAIL: TestBuildBaseEgressUsesExactWorkloadAndCancelsOnSuccess (2.00s)
--- FAIL: TestBuildBaseEgressCancelsOnWaitReadyFailure (2.00s)
--- FAIL: TestBuildBaseEgressSkipsWorkloadOutsideScope (0.10s)
Executed 1 out of 1 test: 1 fails remotely.
```

Restored: `1 test passes`.

Note for reviewers: a bare `ci test` on this branch reports "affected-targets: no changes, nothing to test" and exits 0 despite the branch being ahead of origin/main. That is a false green of the #4118 family. Run the target explicitly: `ci test -- //projects/embervm/noded/server:server_test`.

## Alternatives rejected

- Skipping prewarm during base builds: defeats the point of a warm base.
- Making prewarm failure non-fatal during builds: ships bases whose warmth is a lie, which is what the guard prevents.
- Reverting #4423's image-side prewarm: restores service but loses the codex and pi prewarm, and leaves the same trap for the next workload whose CLI needs network.

Found during the 0.41.7 to 0.47.11 prod jump (#5224). Twelve other workloads built cleanly.

### DIFF
--- a/projects/embervm/noded/server/egress_test.go
+++ b/projects/embervm/noded/server/egress_test.go
@@ -1,12 +1,18 @@
 package server
 
 import (
+	"bytes"
+	"context"
+	"errors"
 	"io"
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
+
+	nodev1 "github.com/jomcgi/homelab/projects/embervm/proto/embervm/node/v1"
 
 	"github.com/jomcgi/homelab/projects/embervm/noded/config"
 )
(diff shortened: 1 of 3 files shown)


jomcgi:
## Helm Manifest Diff

PR #5360, opened 2026-08-27
fix(embervm): self-heal an unopenable SQLite op-log instead of crashlooping

### LINKED ISSUE #5359: embervm: an unopenable SQLite op-log crashloops the control plane instead of self-healing
An unopenable SQLite op-log takes the whole embervm control plane down and keeps it down. Dev has been crashlooping for roughly 12 hours because of it, and because the conformance gate needs a live control plane, it also blocked a verified production fix from promoting while `claude-runtime` was down in prod.

## What happened

```
Application embervm exited: Embervm.Application.start(:normal, []) returned an error:
  shutdown: failed to start child: Embervm.OpLog.SQLite
  ** (EXIT) {:open_failed, :database_open_failed}
Kernel pid terminated (application_controller)
```

`embervm-dev-embervm-69bb87fb8b-qgzmh`, container `control-plane`, 135 restarts, backoff-capped, since the pod was created at 2026-08-26T16:19:13Z. The conformance suite passed at 16:19:51 on that same pod, so the store opened once and then stopped opening.

The most likely trigger is an unclean kill mid-write: the control plane was liveness-killed during that roll, which is how a SQLite file ends up unopenable. Not confirmed, because there is no read-only way to inspect the volume (see below).

Not a storage fault: the Longhorn volume is `healthy`, `attached`, 654 MB of 1 Gi.

## The actual defect

The store is a supervised child, so an open failure propagates to `Embervm.Application.start` and the BEAM exits. Every restart re-opens the same file and dies the same way, so the control plane cannot self-heal from a disposable dev op-log.

A dev op-log is disposable. A control plane that will not boot is not.

## Blast radius beyond the control plane

The conformance runner talks to the control plane, so a dead control plane means no verdict, so the Kargo dev promotion sits in its verdict poll, so no Freight is ever verified for prod. During this window production could not receive an unrelated, already-verified fix (#5357) while `claude-runtime` was down. One dev data file blocked a production remediation.

## Fix

On a SQLite op-log open failure, do not take the application down:

1. Log the failure loudly WITH diagnostics that distinguish the causes: the path, the file size, and the filesystem free space. `database_open_failed` alone does not separate corruption from a full disk or a permissions problem, and that ambiguity cost real time here.
2. Quarantine the unopenable file by renaming it (and its `-wal` / `-shm` siblings) to a timestamped name. Never delete: keep it for forensics.
3. Recreate and continue, degraded, with a clear log line saying history was quarantined.
4. If the retry ALSO fails, fail for real. A second failure means the problem is the filesystem or permissions, not the file, and quarantining cannot help.

Scope this to the SQLite store only. Production uses Postgres for the op-log, so this exact path is dev-only, and Postgres failure semantics should not change as a side effect.

## Related latent problem, not fixed here

`EMBERVM_OPLOG_PATH=/var/lib/embervm/oplog.db` and `EMBERVM_SPEC_TRACE_PATH=/var/lib/embervm/spec-trace.db` share ONE 1 Gi PVC. spec-trace grows continuously on a 24h TTL, and #5342 made every checkpoint larger by adding the queued-task set. If the volume ever does fill, it takes the op-log down with it by the same crash path. Worth either separating the volumes or adding a size guard.

Found while remediating #5357 during the 0.41.7 to 0.47.x prod jump (#5224).

jomcgi:
Fixes #5359.

**This is currently blocking a production remediation.** The embervm dev control plane has been crashlooping for ~12 hours (135 restarts) on an unopenable SQLite op-log. The conformance gate needs a live control plane, so no Freight can be verified, so #5357's already-merged egress fix cannot promote to prod while `claude-runtime` is down there. Dev is already on 0.47.12 with that fix; prod sits at 0.47.11 waiting on a gate that cannot run.

## The defect

`Embervm.OpLog.SQLite` is a supervised child, so an open failure propagated to `Embervm.Application.start/2` and took the BEAM down:

```
shutdown: failed to start child: Embervm.OpLog.SQLite
  ** (EXIT) {:open_failed, :database_open_failed}
Kernel pid terminated (application_controller)
```

Every restart reopened the same file and died identically. A dev op-log is disposable; a control plane that will not boot is not.

## Change

Recovery lives at `start_link/1`, deliberately not in `init/1`. `GenServer.start_link/3` waits for `init/1`, so the open error is still in hand before `start_link/1` hands it to the application supervisor. `init/1` cannot quarantine and then raise without letting the failure escape the child start contract, which would have fixed nothing.

On an open failure:

1. Log the path, the reason, the file size, and the filesystem free space. `database_open_failed` alone does not separate a corrupt file from a full disk from a permissions problem, and that ambiguity is what made this expensive to diagnose. The next occurrence says which it is.
2. Rename the database and its `-wal` and `-shm` siblings under one timestamped basename. Never unlink: the bad file stays for forensics.
3. Retry exactly once, then run degraded with a fresh database and an unmistakable log line.
4. A second failure escapes as before. At that point the filesystem or permissions are the problem and quarantining cannot help, so failing hard is correct.

A healthy database is opened normally and never quarantined.

## Scope

SQLite store only. Production uses Postgres for the op-log and its failure semantics are unchanged. The spec-trace store, which shares the same PVC, is untouched and tracked separately in #5359.

## Verification

Falsified rather than assumed. With the tests unchanged and only the recovery branch removed:

```
1) test a second consecutive open failure fails hard after one retry (Embervm.OpLog.SQLiteTest)
2) test an unopenable database is quarantined and the supervised store starts with fresh history (Embervm.OpLog.SQLiteTest)
1555 tests, 2 failures, 22 skipped
```

Restored: `1555 tests, 0 failures, 22 skipped`.

Reviewers: `ci test` gives NO signal on this suite. It is a genrule, so `bazel test` reports "Found 1 target and 0 test targets" and reads as cached-green (#4118). Build it with `bb remote ... build //bazel/erlang:mix_test_smoke` and judge the `N tests, N failures` line.

## Expected effect on merge

The published chart carries this plus #5357. Dev receives it through `argocd-update`, which runs before the verdict gate, so the control plane recovers without manual intervention, the gate runs again, and prod promotes with the egress fix. If dev's volume turns out to be full rather than the file corrupt, this correctly fails hard on the second attempt and the new diagnostics say so plainly.

## Still latent, not fixed here

`oplog.db` and `spec-trace.db` share one 1 Gi PVC. spec-trace grows continuously on a 24h TTL and #5342 enlarged every checkpoint. If that volume fills it takes the op-log down by this same path, and this fix will not save it. Noted in #5359.

### DIFF
--- a/projects/embervm/control/lib/embervm/op_log/sqlite.ex
+++ b/projects/embervm/control/lib/embervm/op_log/sqlite.ex
@@ -113,6 +113,7 @@ defmodule Embervm.OpLog.SQLite do
   @default_compact_batch_size 500
   # The meta key the durable ops-journal prefix marker lives at; absent reads 0.
   @marker_key "compacted_through_seq"
+  @sqlite_header "SQLite format 3\0"
 
   @ddl [
     """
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

Issue #5363, opened 2026-08-27
Restore the demos trace waterfall on the Honeycomb-backed span store


jomcgi:
`home/observability/traces.py` queried SigNoz's ClickHouse (`signoz_traces.distributed_signoz_index_v3`) for the private demos trace waterfall. With SigNoz removed (#5362) both `fetch_trace_spans` and `fetch_correlated_spans` return `[]`.

This is not a bug in the UI: `TraceWaterfall.svelte` already treats an empty list as "still ingesting" and polls, so the panel spins rather than errors.

## To restore

- Repoint both functions at whatever span store Phase 2 lands on. Honeycomb has a query API, but per-trace span fetch on a heavily-sampled dataset will usually miss, so this may want its own narrow span sink instead.
- `_TRACE_ID_RE` and `_normalize_spans` were kept deliberately; the normalized span shape the frontend consumes is unchanged.
- The goose correlation path matters: `goose-coding` does not honor an inbound TRACEPARENT, so its spans live in their own trace and are recovered by the `caller.trace_id` resource attribute. Whatever replaces this must preserve resource attributes.
- `projects/monolith/frontend/src/lib/public/fcstory/bake-fc-story.sh` regenerates the public fc-story from the same store and is inoperative until this lands. The already-baked `fcstory/data/` keeps the public page live in the meantime.

jomcgi:
## Agent brief

### Outcome

This issue is not yet actionable because its target span store does not exist: #5362 (remove SigNoz, ship OTel collector to Honeycomb) is still OPEN, and the issue body explicitly defers the store choice to whatever Phase 2 lands, flagging that Honeycomb per-trace fetch on a heavily sampled dataset will usually miss and that a narrow span sink may be needed instead.

### Acceptance

Done means both `fetch_trace_spans` and `fetch_correlated_spans` return real spans from the Phase 2 store for a fresh demo run, `GET /trace/{trace_id}` reports `complete: true` once spans ingest, the goose `caller.trace_id` correlation renders its sub-timeline, `bake-fc-story.sh` is rewritten off ClickHouse (its cold-trace path shells into the `signoz` ClickHouse pod, which #5362 removes), and the polling empty state in the waterfall is kept for genuine ingest lag.

### Files

- `projects/monolith/home/observability/traces.py` (both fetch functions return `[]`; `_TRACE_ID_RE` and `_normalize_spans` retained for reuse)
[...]

Issue #5362, opened 2026-08-27
Remove SigNoz: replace with a minimal opt-in OTel collector shipping to Honeycomb


jomcgi:
SigNoz (chart, addons operator, 28 alert templates, dashboard sidecar, dashboards library) is being removed from the cluster. Telemetry moves to a minimal, opt-in OpenTelemetry Collector exporting heavily-sampled traces to Honeycomb.

Data loss is accepted: no ClickHouse migration, no alert history.

## Phases

- [ ] **Phase 1: migrate references.** Remove every functional monolith dependency on SigNoz while it is still running. GPU ticker metrics move from ClickHouse to a direct DCGM exporter scrape; trace waterfall stubbed; firing-alerts path deleted.
- [ ] **Phase 2: minimal collector.** New `projects/platform/otel-collector/`: OTLP receiver on 4317/4318, `httpcheck` receiver probing public and service URLs, tail sampling, OTLP exporter to `api.honeycomb.io`. No cluster-wide log or metric scraping. No services opt in on day one.
- [ ] **Phase 3: delete SigNoz.** `projects/platform/{signoz,signoz-addons,signoz-dashboards-library}/`, the kustomization entries, the two Kyverno policies, the dashboard-sidecar image, the monolith `signoz-dashboards` chart dependency and vendored subchart, and every OTLP endpoint value across monolith, monolith-public, embervm, inference, model-bench, kyverno, cloudflare-gateway, cilium, keda, seaweedfs, argo-workflows.
- [ ] **Phase 4: docs.** observability.md, observability-alerting.md, services.md, contributing.md, three runbooks, platform/ARCHITECTURE.md, README.md, .claude/CLAUDE.md. ADRs stay as historical rationale; one new ADR records this decision.

## Monitoring after the cutover

The 5 HttpCheck synthetic probes and the incident.io channel are dropped. Replacement:

1. UptimeRobot points at a public route reaching the collector's health endpoint (metamonitoring).
2. The collector's `httpcheck` receiver probes public and in-cluster service URLs.
3. Honeycomb holds the probe results and alerting.

## Constraints found during scoping

- No cluster-wide default-deny egress. The monolith uses `egressDeny`, which denies only what it names, so neither the DCGM scrape nor Honeycomb egress needs a new Cilium rule.
- `nvidia-dcgm-exporter.gpu-operator.svc:9400` already serves Prometheus text and has been up 294d. One RTX 4090 on node-4.
[...]

PR #5364, opened 2026-08-27
refactor(monolith): drop the SigNoz dependency from stats, traces, and alerts

jomcgi:
Phase 1 of #5362. Removes every functional monolith dependency on SigNoz **while SigNoz is still running**, so the deletion PR is pure removal and this one is independently revertible.

Nothing under `projects/platform/signoz*` is touched here.

## What changes

**GPU metrics move off ClickHouse.** The public stats ticker's GPU block was the only part of it that touched SigNoz, via three `signoz_metrics.distributed_samples_v4` queries. It now scrapes `nvidia-dcgm-exporter.gpu-operator.svc:9400/metrics` directly, which has been serving Prometheus text for 294 days. Values are averaged across GPU series so a second card would not skew them, and the rounding mirrors the old SQL exactly.

The one deliberate behaviour change: the ClickHouse query averaged over a 5 minute window, the scrape is instantaneous. The averaging was an artifact of ClickHouse being in the path, not something the ticker needs.

Everything else on the ticker (nodes, pods, deployments, ArgoCD apps, CPU and memory, knowledge counts, latest commit, last deploy) already came from the K8s metrics API, Postgres, and GitHub. The public `/stats` endpoint reads a Postgres snapshot row, so **the public tier is unchanged**.

**Trace waterfall stubbed** (#5363). No span store exists until the replacement lands, so both fetch helpers return `[]`. `TraceWaterfall.svelte` already treats an empty list as "still ingesting" and polls, so the panel spins rather than errors. `_TRACE_ID_RE` and `_normalize_spans` are kept so the normalized shape the frontend consumes cannot drift, and the goose `caller.trace_id` correlation constraint is recorded in the docstring because it constrains what can replace the store.

**Firing-alerts path removed entirely**: the SigNoz `/api/v1/rules` fetch, the `check-firing-alerts` MCP tool, the cluster-chat `firing_alerts` tool, and the dashboard section. The `home.cluster_snapshot` `alerts` column is retained and written empty rather than migrated away, which is less churn on a live table.

**Chart wiring**: `CLICKHOUSE_*` swapped for `DCGM_EXPORTER_URL` on the stats rollup job, `SIGNOZ_URL`/`SIGNOZ_API_KEY` dropped from the cluster-snapshot job and the deployment, and the `signoz-clickhouse` 1Password binding removed.

## Two comments that would have gone stale

Two places claimed the #4659 alert as the tripwire for silent Cilium selector drift. That alert dies with SigNoz, so both now say the tripwire is gone rather than implying monitoring that will not exist.

## Deliberately deferred

- The `signoz-dashboards` subchart, `Chart.lock`, and the dashboards themselves. Those ConfigMaps are inert without the sidecar, and the vendored `charts/` tree makes it deep-Helm work that belongs with the deletion.
- Every OTLP emission endpoint (`backend.otelEndpoint`, `frontend.otelCollectorUrl`, `OTEL_*`). Repointing those is the collector cutover, not this PR.
- SigNoz mentions in CV and engineering copy, which are factual claims about the platform for Joe to word.

## Verification

`Executed 229 out of 773 tests: 773 tests pass.`

Refs #5362


jomcgi:
## Helm Manifest Diff

PR #5365, opened 2026-08-27
feat(embervm): host Apache Iggy as a stateful-class runtime

jomcgi:
Adds `iggy`, an EmberVM **stateful-class** workload running Apache Iggy 0.8.0 as a scale-to-zero microVM: one long-lived VM owning one writable volume, banking when idle and waking on the next inbound TCP connection, exposed cluster-internally on node Envoy listener port **5402**. The shape follows `scratch-postgres` (ADR embervm/001, D-R4.PR-11.1) almost line for line.

Rationale is in ADR embervm/039 (`docs/decisions/embervm/039-iggy-stateful-message-streaming-runtime.md`); operational shape in `projects/embervm/runtimes/iggy/README.md`.

## What is Iggy-specific

Everything below was checked against the 0.8.0 binary directly, not inferred from upstream's `main` (which is on the 0.9 line and differs).

**The server binary comes out of an OCI image.** `iggy-server` is not in Wolfi, and the usual `multiarch_http_archive`-against-a-GitHub-release fallback does not work: upstream routes the `server-*` tag to `registry: dockerhub`, and its per-target release tarballs are `actions/upload-artifact` artifacts with **30-day retention**, not release assets. The image *is* the release channel. A new repository rule, `//bazel/tools/oci:oci_binaries`, lifts the binary out of the digest-pinned `apache/iggy` image; it is the thin sibling of the existing `//bazel/tools/postgres:oci_postgres`, which already does this for the PG test fixture.

Upstream's image is `debian:trixie-slim` with `libhwloc15`/`libudev1`, which reads like a dependency the guest would inherit. It is not: the binary is `x86_64-unknown-linux-musl` and fully static (no `PT_INTERP`, no `DT_NEEDED`), so the rootfs stays 15 Wolfi packages. The rule **asserts** that static property on every fetch, so an upstream that switches to a dynamic build fails the fetch with a named error instead of producing an image whose binary dies at exec time on a guest boot.

**Configuration is `IGGY_`-prefixed env vars, with no config file.** Iggy falls back to the config compiled into the binary when it finds no `config.toml`, so nothing is baked; per-deployment values keep riding the existing `mmds_env` seam. Three overrides are not optional: `IGGY_TCP_ADDRESS=0.0.0.0:8090` (the compiled default is loopback, which would fail the tap-NIC health probe forever while the server looked healthy inside the guest), `IGGY_QUIC_ENABLED=false` (on by default; nothing should listen that the Envoy listener cannot route to), and `IGGY_SYSTEM_LOGGING_FILE_ENABLED=false` (rotating log files default to on, under `system.path`, capped at 4 GiB, which would compete with message segments for the volume).

**A first boot with no `IGGY_ROOT_PASSWORD` is refused.** Iggy does not fail without one and does not fall back to a fixed default: it generates a random password and prints it to stdout exactly once. A stateful volume bootstraps once ever, so that log line would be the only copy of the credential, on a datastore that outlives it by up to `bankedTtlSeconds`. `ember-iggy-init` refuses the boot instead, naming the `secretRef` and the Secret key.

"First boot" keys on the `state/log` file under the data root being **non-empty**, not on `state/info` existing: the server writes system info *before* the root user, so an `info`-keyed probe would let an interrupted first boot skip the password requirement forever. Same window the postgres runtime's `PG_VERSION` probe documents.

**No `initdb`-equivalent to gate**, unlike postgres: the server bootstraps its own state in-process and loads it idempotently on later boots. **vCPU count is the shard count**, so `vcpus` and `memMib` move together (noted in the CR).

## Shipped inert

Production gets `enabled: false` **and** an emptied `guestImage.repository`, so no node builds a base for it yet. Turning it on is a deliberate two-line flip, and the 1Password item carrying `IGGY_ROOT_PASSWORD` must exist first or the first boot is refused by design. This is more conservative than `scratch-postgres`, which keeps its repository set because `demo-postgres` shares that base; nothing shares this one.

## Verification

Green on `b99f183`: `pr-checks` successful, `route-b/semgrep` successful, no merge conflict.

The substantive gate was `bazel test //...` on the previous head (invocation `ff34b3e5`, 12,804 actions, SUCCESS), which covered the pieces that could not be checked without CI: the `oci_binaries` fetch against the real `apache/iggy` image, the apko lock resolve, the image build, the chart tests, and the helm render. `bazel build //bazel/images:push_all` also passed, and on the current head that target includes this image.

Checked before pushing: `go vet`, the guest-init unit tests (20 subtests), `linux/amd64` + `darwin/arm64` builds, `gofmt`, the apko lock against the repo's own `verify-apko-lock.sh` logic, the rendered Workload CR against the CRD's `openAPIV3Schema` (required fields present, no unknown keys, `listenPort` 5402 in the declared 5400-5409 range and unique), and the Iggy behaviours above by running the extracted 0.8.0 binary.

The `style: auto-format` commit is CI's own: it generated the `image.push` registration in `bazel/images/BUILD`, the digest mapping, `bzl_library` targets for the two new `.bzl` files, and the KG docs manifest. It did not rewrite either hand-written BUILD file.

**Left for the first real cold boot:** the guest kernel is the kata bundle's, and Iggy's shard allocator pins threads to CPUs and binds memory to a NUMA node. Those are unprivileged calls a Firecracker guest should serve, but "should" is not "does". Nothing before a real boot can settle it.

## Note

`docs/decisions/embervm/README.md` indexes up to 036; 037 and 038 are already missing from that table. I added 039's row and left that pre-existing gap alone rather than guess at their summaries.


jomcgi:
## Helm Manifest Diff

jomcgi:
Closed as stale on Joe's decision (2026-09-18): 1597 commits behind, adds a docs/decisions ADR that main deleted, and no issue asks for an Iggy runtime.

Issue #5366, opened 2026-08-27
embervm: two comments describe the egress CA delivery as unsolved after it was solved


jomcgi:
Two comments describe the egress CA as an unsolved problem. It was solved, and both comments are now actively misleading. During a live production outage they sent an investigation toward "this is ADR 023's open design problem, we cannot fix it today" when the actual fix was one missing function call.

## 1. `projects/embervm/chart/values.yaml` around line 1718

```yaml
  ca:
    enabled: false
    # Secret holding tls.crt/tls.key; defaults to <fullname>-egress-ca. Nothing in
    # this chart creates it, precisely because where the matching public cert
    # enters the guest is the hard half of the decision.
```

Both claims are false now:

- `chart/templates/egress-ca.yaml` DOES create it, via cert-manager.
- "Where the matching public cert enters the guest" was answered: the guest fetches the PUBLIC cert over its host-local vsock at spawn, from the sidecar's reserved name. `projects/embervm/deploy/values.yaml` around line 645 says so and sets `ca.enabled: true`, and the guest half is implemented in `runtimes/claude/shim.py` as `fetch_egress_ca` / `install_egress_ca` / `apply_egress_ca_trust`.

The chart DEFAULT staying `false` is fine and probably correct. The comment explaining WHY is what is stale, and because it reads as a standing design gap rather than a default, it is worse than no comment.

## 2. `projects/embervm/runtimes/claude/shim.py`, the `apply_egress_ca_trust` docstring

```
At base-build time the egress lane is not open either, so the fetch there
fails with ECONNRESET and the trust is simply absent for every session that
base ever serves.
```

True when written. #5358 opened the egress lane during base builds, which invalidated it. That stale premise is exactly why nobody added the trust step to the build path, which is the outage.

Being fixed alongside the build-time trust call, noted here so the pair is tracked together.

## Why this is worth an issue rather than a shrug

This repo leans on load-bearing comments deliberately, and that is usually a strength: several of them caught real defects during this same jump. The failure mode is the inverse. A comment that documents a limitation becomes a claim about the present, and the next reader takes it as current state rather than as history.

Both of these describe a constraint that a later commit removed. Neither was updated by the commit that removed it, because that commit had no reason to look.

[...]

PR #5367, opened 2026-08-27
fix(embervm): apply the egress CA trust during prewarm so base builds work

jomcgi:
**Ends a live production outage.** `claude-runtime` has been unable to build a base in embervm prod for roughly 14 hours, retrying every 60s. Twelve of thirteen workloads are `BaseBuilt`; this is the last one.

## Cause

`prewarm` spawns the CLIs during a base build without applying the egress CA trust. The egress proxy MITMs the catalog hosts to inject credentials (the guest never holds them, which is the design), so the Claude CLI, which is Node, rejects the minted leaf without `NODE_EXTRA_CA_CERTS`. Its init blocks, the init timeout fires, `prewarm` sets `fatal_error`, `ready()` refuses while that is set, and `WaitReady` times out.

Observed in prod, guest serial and sidecar log respectively:

```
ember-claude-shim: egress-copy: up closed total=342 err=none
ember-claude-shim: egress-copy: down closed total=2334 err=none
ember-claude-shim: initialization timed out after 30 seconds, sending SIGINT
ember-claude-shim: CLI prewarm failed: timed out waiting for Claude initialization
```
```
INFO  egress allowed (inject, tls)  dest=api.anthropic.com:443
WARN  egress swap: guest TLS handshake failed  host=api.anthropic.com  err=EOF
```

## Why nobody wired it

`apply_egress_ca_trust` already does the whole job, and session guests call it every turn. Its docstring said why the build path did not:

> At base-build time the egress lane is not open either, so the fetch there fails with ECONNRESET and the trust is simply absent for every session that base ever serves.

That was true when written. #5358 (issue #5357) opened the egress lane during base builds and invalidated it, but nothing added the trust step, because the comment still said it was impossible. The lane has been open and untrusted since.

## Change

Call `apply_egress_ca_trust()` once at the top of `prewarm`, before the first spawn. The exported variables land in `os.environ` and every CLI child inherits them via `child_env = os.environ.copy()`. The resulting snapshot then carries both the installed bundle and the trust variables into every session restored from it.

The per-turn call is untouched. That is what lets a CA rotation take effect without a fleet base rebuild, and a restored guest still needs it because the snapshot carries build-time values.

`init_timeout` 30 to 60: a build now pays a CA fetch plus a cold authenticated init on a node that may be baking other guests. Two tests pinned 30 and now assert 60.

Both stale comments are rewritten to name #5358 as what changed the premise. That old text is what sent this investigation toward "ADR 023's unsolved problem" mid-outage, when the delivery mechanism had in fact been solved for the GitHub lane and prod already runs `egress.ca.enabled: true`. Tracked separately in #5366.

## Security

The guest gains only the CA's PUBLIC certificate. The private key stays sidecar-only and never reaches a guest, so this does not let a guest mint leaves, impersonate a host, or reach anything new. Trusting the CA only permits the sidecar to MITM the guest, which is the design, and prod session guests already apply this exact trust every turn. ADR 023 decision 3's boundary is unchanged.

Two facts worth recording: the base build now completes an authenticated `api.anthropic.com` call over the same credential-injection path every turn already uses, and a CLI parked at build time holds the build-time CA until respawned, because Node reads `NODE_EXTRA_CA_CERTS` at process start. That affects https telemetry from a parked CLI after a rotation, not turns, which ride the cleartext lane.

## Verification

Falsified rather than assumed. With the tests unchanged and ONLY the `prewarm` trust call removed:

```
FAILED test_process_manager_prewarm_child_inherits_fetched_egress_ca_trust
FAILED test_process_manager_prewarm_marks_ready_after_init
Executed 1 out of 1 test: 1 fails remotely.
```

The inheritance test spawns a real child process and asserts all five trust variables arrive, rather than only checking that the function was called.

Restored: `1 test passes`. The no-sidecar path (dev, which runs no egress-proxy) still completes prewarm with the trust variables absent, and is tested.

Note for reviewers: a bare `ci test` on a fresh worktree reports "affected-targets: no changes, nothing to test" and exits 0 having tested nothing. Run the target explicitly: `ci test -- //projects/embervm/runtimes/claude:shim_test`.

## Expected effect

The image rebuild moves only the claude-runtime digest, so the prod roll rebuilds one base rather than the fleet. Success is `claude-runtime` reaching `BaseBuilt` with a ref off `7c08eb4dbc91`, then 13 of 13 workloads built.

### DIFF
--- a/projects/embervm/runtimes/claude/shim.py
+++ b/projects/embervm/runtimes/claude/shim.py
@@ -690,17 +690,14 @@ def fetch_egress_ca(timeout=EGRESS_VSOCK_CONNECT_TIMEOUT_SECONDS):
 def apply_egress_ca_trust():
     """Fetch + install the egress CA and export the trust variables.
 
-    Called PER TURN, not at startup. A session guest is restored from a shared
-    snapshot, so its process memory is cloned with main() long since finished:
-    anything done there runs once, at base-build time, on a different machine,
-    and can never run again for a real session. At base-build time the egress
-    lane is not open either, so the fetch there fails with ECONNRESET and the
-    trust is simply absent for every session that base ever serves.
+    Since #5358 the egress lane is open during base builds. Prewarm applies the
+    trust once before spawning any CLI, so the resulting snapshot carries both
+    the installed bundle and these environment variables into every session.
 
-    A turn is the first moment that is guaranteed to be post-restore with the
-    lane live, which is why ensure_workspace_volume() is called from the same
-    place. Re-running is cheap (one vsock round trip) and is what keeps a CA
-    rotation from needing a fleet base rebuild.
+    A session guest is restored from that shared snapshot, so the per-turn call
+    remains necessary for CA rotation without a fleet base rebuild. Re-running
+    is cheap (one vsock round trip) and updates the environment before a turn
+    can lazily spawn a new CLI.
     """
     bundle = install_egress_ca()
     if not bundle:
@@ -3467,6 +3464,7 @@ class ProcessManager:
     def prewarm(self):
         """Start configured CLIs and leave them ready for the first turn."""
         try:
+            apply_egress_ca_trust()
             _ensure_cli_dir(self.claude.workspace)
             for cli in self._prewarm_clis:
                 if cli == "claude":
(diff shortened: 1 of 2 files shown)


jomcgi:
## Helm Manifest Diff

